PR #45, opened 2026-04-07
feat: add ab-keys CLI command

johnsonlee:
Adds `graphite-query ab-keys <graph-dir>` to find all AB test keys in the codebase, including those loaded from config files (YAML/properties).

Sources:
1. Inline string literals
2. Enum constants
3. Config file lookups (parsed from YAML/properties via `graph.resources`)

100% line coverage for graphite-query module.


johnsonlee:
Wrong approach. Should be a generic graph feature (ConfigNode), not a special-purpose command. Replacing with proper implementation.

PR #76, opened 2026-04-24
Add unified C4 architecture API

johnsonlee:
## Summary

This PR adds one agent-discoverable C4 architecture endpoint for saved Graphite graphs. It derives Context, Container, and Component views automatically from graph evidence, returns a Structurizr workspace JSON model by default, and renders the same workspace as Structurizr DSL, Mermaid, or PlantUML when requested.

## Delivery Status

| Area | Detail |
| --- | --- |
| API and agent discovery | ✅ `GET /api/architecture/c4` supports `level={context,container,component,all}`.<br>✅ `Accept` negotiation and `format=` support `json`, `dsl`, `mermaid`, and `plantuml`.<br>✅ `/openapi.json`, `/swagger.json`, and graphite-mcp expose the C4 API for agents.<br>✅ C4 OpenAPI no longer advertises a model `limit`; structured JSON/DSL are not cropped for diagram readability. |
| C4 semantics | ✅ Context centers the subject software system and directly connected people/software systems.<br>✅ Container uses runtime/deployable boundaries when graph evidence exists; archives/packages/classes are evidence, not container definitions by themselves.<br>✅ Component groups implementation evidence into capability-level elements inside an inferred container.<br>⚠️ Static graph evidence cannot precisely infer business personas, user journeys, owned data stores, serverless functions, deployment topology, or business systems that leave no code/resource trace. Those require runtime/business evidence outside the saved graph. |
| Dependency and archive evidence | ✅ Third-party dependencies use artifact provenance instead of package-name-only guessing.<br>✅ Single-root package namespaces such as `okhttp3.*` stay rooted at the public namespace instead of overfitting to internals.<br>✅ Java archive layout strings are centralized in `JavaArchiveLayout` and composed from shared fragments.<br>✅ Library-to-runtime and library-to-library relationships are represented when present in graph evidence. |
| Rendering and model integrity | ✅ Structurizr JSON and DSL are rendered from the full inferred workspace model.<br>✅ Mermaid and PlantUML render readable slices from that same model; node/edge caps are renderer-only and do not remove workspace elements/relationships.<br>✅ C4 model metadata/evidence are typed DTOs; Gson handles wire conversion at the codec boundary instead of hand-written map schemas.<br>✅ C4 domain fields use enums for level, element type/kind, relationship type/kind, architecture type, external dependency kind, and rendering layer kind.<br>✅ Text renderers share a typed rendering plan and syntax-specific renderers only own Mermaid/PlantUML syntax. |
| Tests and documentation | ✅ Direct C4 tests cover `C4ModelInferer`, `SystemBoundaryDetector`, `ExternalSystemClassifier`, `ContainerClusterer`, and `ComponentSelector`.<br>✅ Regression tests cover full workspace model vs text renderer cropping, structured JSON/DSL behavior, OpenAPI discovery, and text output formats.<br>✅ English and Chinese C4 design docs describe C4 semantic mapping, output contracts, renderer budgets, and decision rationale. |

## Validation

| Check | Result |
| --- | --- |
| Head commit | `e5a3739` |
| GitHub Actions `Run unit tests` | Passed, run #408 |
| GitHub Actions coverage threshold | Passed, every module >= 98% |
| `git diff --check` | Passed |
| `:explore:check` | Passed |
| `:explore:koverLog` | Passed, `98.6606%` line coverage |
| `:core:check :cypher:check :query:check :sootup:check :explore:check` | Passed |


graphite-explore/src/main/kotlin/io/johnsonlee/graphite/cli/c4/C4ArchitectureService.kt
```diff
@@ -0,0 +1,63 @@
...
+        designOverlay: Map<String, Any?>? = loadC4DesignOverlay()
+    ): Map<String, Any?> =
+        mapper.toWorkspace(inferer.buildViewModel(graph, level, limit, designOverlay))
+
+    internal fun defaultLimit(level: String): Int =
+        when (level) {
+            "context" -> 12
+            "container" -> 12
+            "component" -> 16
+            else -> 12
```
johnsonlee: where does these magic numbers come from?

graphite-explore/src/main/kotlin/io/johnsonlee/graphite/cli/c4/C4ArchitectureService.kt
```diff
@@ -0,0 +1,63 @@
...
+        renderer.renderMermaid(workspace)
+
+    internal fun renderPlantUml(workspace: Map<String, Any?>): String =
+        renderer.renderPlantUml(workspace)
+
+private fun loadC4DesignOverlay(): Map<String, Any?>? {
+    val graphDir = graphDir ?: return null
+    val candidates = listOf(
+        graphDir.resolve("c4-design.json"),
+        graphDir.resolve("architecture").resolve("c4.json")
```
johnsonlee: By default there isn't any `c4-design.json` and `c4.json`

graphite-explore/src/main/kotlin/io/johnsonlee/graphite/cli/c4/C4ModelInferer.kt
```diff
@@ -0,0 +1,1980 @@
...
+        if (thirdPartyDependencies.isNotEmpty()) {
+            thirdPartyDependencies.forEach { dependency ->
+                add(
+                    mapOf<String, Any?>(
+                        "id" to dependency["id"],
+                        "type" to "softwareSystem",
+                        "name" to dependency["name"],
+                        "description" to "Reusable third-party library capabilities referenced from the application",
+                        "kind" to dependency["kind"],
+	                            "architectureType" to externalArchitectureType(dependency["kind"]?.toString().orEmpty()),
```
johnsonlee: **Inconsistent indentation**: Lines 391-398 (and again at 455-462) use tab indentation while the rest of the file uses 4-space indent. Likely a copy-paste artifact. Normalize to match the rest of the file.

graphite-explore/src/main/kotlin/io/johnsonlee/graphite/cli/c4/C4ModelInferer.kt
```diff
@@ -0,0 +1,1980 @@
...
+        addEdge(pair.second, pair.first, weight)
+    }
+    val strongestNeighbor = units.associateWith { unit ->
+        adjacency[unit]
+            ?.entries
+            ?.maxWithOrNull(compareBy<Map.Entry<String, Int>> { it.value }.thenByDescending { it.key })
+            ?.key
+    }
+    val parents = units.associateWith { it }.toMutableMap()
+    fun find(unit: String): String {
```
johnsonlee: **Union-find without path compression**: `find()` walks parent pointers linearly without path compression. For small inputs this is fine, but adding one line (`parents[current] = parents[parents[current]!!]!!`) inside the while loop gives O(alpha(n)) amortized cost. A matter of engineering discipline on graph algorithms.

graphite-explore/src/main/kotlin/io/johnsonlee/graphite/cli/c4/C4ModelInferer.kt
```diff
@@ -0,0 +1,1980 @@
...
+        targetKind == "integration" -> "uses"
+        else -> "collaborates-with"
+    }
+}
+
+private fun describeArchitecturalDependency(
+    relationshipKind: String,
+    sourceName: String,
+    targetName: String,
+    scope: String
```
johnsonlee: **Unused parameter**: `scope` is accepted but never referenced in the function body. Either remove it or use it to vary the description text (which seems to be the original intent given the parameter name).

graphite-explore/src/main/kotlin/io/johnsonlee/graphite/cli/c4/C4ModelInferer.kt
```diff
@@ -0,0 +1,1980 @@
...
+
+private fun isJavaRuntimeArchitectureClass(className: String): Boolean =
+    className.startsWith("java.") ||
+        className.startsWith("javax.") ||
+        className.startsWith("jakarta.") ||
+        className.startsWith("jdk.")
+
+
+
+private fun externalArchitectureType(kind: String): String =
```
johnsonlee: **Unnecessary private wrappers**: Lines 1965-1978 wrap top-level `C4Support` functions as private instance methods via fully-qualified calls. Since `C4ModelInferer` is in the same package as `C4Support`, these functions are already directly callable. These wrappers add indirection with no benefit -- just call the package-level functions directly at call sites.

graphite-explore/src/main/kotlin/io/johnsonlee/graphite/cli/c4/C4ModelInferer.kt
```diff
@@ -0,0 +1,1980 @@
...
+    }
+    return boundary
+}
+
+private fun dominantNamespace(packageName: String): String {
+    val segments = packageName.split('.').filter { it.isNotBlank() }
+    return segments.take(minOf(2, segments.size)).joinToString(".")
+}
+
+private fun containerName(className: String, systemBoundary: String): String {
```
johnsonlee: **Dead code**: `containerName()` and `externalContainerName()` (lines 1904-1920) are never called anywhere in this file or the codebase. Remove them.

graphite-explore/src/main/kotlin/io/johnsonlee/graphite/cli/c4/C4ModelInferer.kt
```diff
@@ -0,0 +1,1980 @@
...
+        } + buildContainerDependencyRelationships(graph, callSites, layout)
+    return mapOf(
+        "type" to "container",
+        "elements" to layout.containers.map { container ->
+            mapOf(
+                "id" to container.id,
+                "type" to "container",
+                "name" to container.name,
+                "description" to "Internal runtime boundary derived from code graph structure",
+                "kind" to inferContainerKind(
```
johnsonlee: **Redundant re-computation**: `inferContainerKind()` is called 3 times per container element: once for `kind` (line 975), again for `architectureType` (line 982), and again for every relationship pair (lines 939-951). Compute once per container and store the result. This is not a performance issue (the function is cheap), but it's a readability/maintenance issue -- if the kind inference logic changes, you need to verify all three call sites produce consistent results.

graphite-explore/src/main/kotlin/io/johnsonlee/graphite/cli/c4/C4ModelInferer.kt
```diff
@@ -0,0 +1,1980 @@
...
+                "name" to externalSystemName(graph, key, entries.map { it.key }),
+                "weight" to weight,
+                "source" to externalSystemSource(key),
+                "kind" to kind,
+                "confidence" to externalSystemConfidence(key),
+                "responsibility" to externalSystemResponsibility(kind, key)
+            )
+        }
+        .sortedByDescending { (it["weight"] as Int) }
+        .take(6)
```
johnsonlee: **Magic numbers throughout the inference engine**: `.take(6)` here, `.take(12)` at line 592, `.take(20)` at line 275, `.take(3)` at 1347/1430, `.take(5)` at 1411, `0.7` threshold at 1584, `0.85` dominance at 1890, `3x` runner-up separation at 1889. These are tuning knobs that control diagram quality. Name them as constants with documentation explaining the design intent, e.g. `MAX_CONTEXT_EXTERNAL_SYSTEMS = 6`. When someone needs to tune the output for a different codebase scale, they will need to find and understand each of these.

graphite-explore/src/main/kotlin/io/johnsonlee/graphite/cli/c4/C4DiagramRenderer.kt
```diff
@@ -0,0 +1,580 @@
...
+        @Suppress("UNCHECKED_CAST")
+        val containers = (app["containers"] as? List<Map<String, Any?>>).orEmpty()
+        val externals = softwareSystems.filter { it["id"] != primarySystem["id"] }
+        val actorIds = people.mapNotNull { it["id"]?.toString() }.toSet()
+        val allowed = (containers + externals).mapNotNull { it["id"]?.toString() }.toSet() + actorIds
+        val lines = mutableListOf("graph TD")
+        val actorsAndSystems = people + containers + externals
+        val elementsById = actorsAndSystems.associateBy { it["id"]?.toString().orEmpty() }
+
+        fun rawEdges(source: Map<String, Any?>): List<Map<String, Any?>> {
```
johnsonlee: **Shadowed function**: `renderContainers()` defines a local `rawEdges(source)` function (line 178) that shadows the outer `rawEdges(source, allowed)` function (line 111). Both have the same purpose but subtly different signatures and behavior. This is confusing and error-prone -- rename one or extract the shared logic.

graphite-explore/src/main/kotlin/io/johnsonlee/graphite/cli/c4/C4DiagramRenderer.kt
```diff
@@ -0,0 +1,580 @@
...
+            "%% Container",
+            renderContainers(),
+            "",
+            "%% Component",
+            renderComponents()
+        ).joinToString("\n")
+    }
+}
+
+internal fun renderPlantUml(workspace: Map<String, Any?>): String {
```
johnsonlee: **Structural duplication**: `renderPlantUml()` (lines 305-578) is a near-complete copy of `renderMermaid()` (lines 5-303). The layout decisions (which elements to render, edge selection strategy, layer grouping, connected-component filtering, truncation logic) are identical -- only the output syntax differs.

Consider extracting a `RenderingPlan` that captures layout decisions:
```kotlin
data class RenderingPlan(
    val layers: List<Layer>,
    val edges: List<EdgePlan>,
    val truncated: Int
)
```
Then `renderMermaid()` and `renderPlantUml()` become thin syntax emitters over the shared plan. This eliminates ~250 lines of duplication and ensures layout consistency across formats.

graphite-explore/src/main/kotlin/io/johnsonlee/graphite/cli/c4/C4ArchitectureService.kt
```diff
@@ -0,0 +1,63 @@
...
+    )
+    val type = object : TypeToken<Map<String, Any?>>() {}.type
+    return candidates.firstNotNullOfOrNull { candidate ->
```
johnsonlee: **Unhandled parse errors**: If `c4-design.json` exists but contains malformed JSON, `gson.fromJson()` will throw (e.g., `JsonSyntaxException`). This propagates up through `buildModel()` and crashes the HTTP handler. Wrap in `runCatching` and log a warning -- a broken overlay file should degrade gracefully to pure inference, not 500 the endpoint.

graphite-explore/src/main/kotlin/io/johnsonlee/graphite/cli/c4/C4ArchitectureService.kt
```diff
@@ -0,0 +1,63 @@
...
+            else -> 12
+        }
+
+    internal fun renderMermaid(workspace: Map<String, Any?>): String =
+        renderer.renderMermaid(workspace)
+
+    internal fun renderPlantUml(workspace: Map<String, Any?>): String =
+        renderer.renderPlantUml(workspace)
+
+private fun loadC4DesignOverlay(): Map<String, Any?>? {
```
johnsonlee: **Indentation mismatch**: `loadC4DesignOverlay()` (lines 40-56) is indented at the class body level but breaks the visual rhythm -- the closing brace at line 56 is not vertically aligned with `private fun` at line 40. Likely a formatting artifact.

johnsonlee (commented):
## Staff Engineer Review: C4 Architecture API

This is an ambitious feature -- automated C4 inference from bytecode analysis is genuinely novel. The inference heuristics (union-find clustering, TF-IDF container naming, artifact-level provenance) are well thought out and the Structurizr/Mermaid/PlantUML multi-format output is a nice touch. That said, a PR of this scale (+6425/-671) deserves scrutiny on structural quality, not just correctness.

### Architecture Concerns

**1. `C4ModelInferer` is a 1980-line God Class**

This single class owns all inference logic for all three C4 levels, overlay merging, system boundary detection, container clustering, component scoring, relationship canonicalization, and ~30 helper functions. At this size, it becomes difficult to test individual heuristics in isolation, reason about interactions between inference stages, or onboard another contributor.

A natural decomposition:
- `SystemBoundaryDetector` -- `deriveSystemBoundary()` and its helpers
- `ContainerClusterer` -- `inferContainerLayout()`, `clusterInternalPackageUnits()`, `inferContainerName()`
- `ComponentSelector` -- `buildComponentView()`, `componentScore()`, `selectReadableComponentRelationships()`
- `DesignOverlayMerger` -- `mergeC4DesignOverlay()`, `mergeContextOverlay()`, `mergeStructuralOverlay()`
- `ExternalSystemClassifier` -- all `externalSystem*()` functions + `summarizeExternalSystems()`

**2. Pervasive `Map<String, Any?>` as domain model**

The entire inference-to-mapping-to-rendering pipeline passes untyped maps. `C4Types.kt` defines good data classes for ~10% of the model surface (`ContainerDescriptor`, `SubjectDescriptor`, etc.), but elements, relationships, views, and workspace data remain as `Map<String, Any?>`. Every consumer guesses at keys, and the 50+ `@Suppress("UNCHECKED_CAST")` annotations across the codebase are symptoms.

This is the single highest-leverage refactor available: define `C4Element`, `C4Relationship`, `C4View` data classes. The Structurizr mapper and diagram renderers become straightforward transformations instead of defensive map navigation.

**3. Massive duplication in `C4DiagramRenderer`**

`renderMermaid()` (303 lines) and `renderPlantUml()` (275 lines) are structurally identical -- same layout logic, same edge selection, same layer grouping -- with different output syntax. Nearly every nested function (`sanitizeId`, `architectureTypeOf`, `layerOf`, `rawEdges`, `cappedEdges`, three `render*` functions) is copy-pasted between the two. Extract a shared rendering plan (which elements, which edges, grouping decisions), then have thin format-specific renderers that only handle syntax.

### Testing

The C4 unit tests are thin relative to the 1980-line inference engine. The primary test fixture is a tiny graph (~10 nodes, 1 call site) that does not exercise most inference paths. Notably untested:
- Multi-container clustering with realistic cross-package traffic
- System boundary detection with competing namespace prefixes
- Context-level library family grouping/collapse
- Component relationship canonicalization across architectural layers
- Transitive reduction on non-trivial topologies

Tests call `ExploreCommand.buildC4Model()` which goes through the full Structurizr mapper, making it impossible to verify inference logic in isolation.

### Verdict

The feature itself is valuable and the heuristics are solid. The concerns above are about maintainability at scale, not correctness. I would recommend addressing them in follow-up PRs rather than blocking the merge -- but they should be tracked.

See inline comments below for specific issues.

graphite-explore/src/main/kotlin/io/johnsonlee/graphite/cli/ExploreRoutes.kt
```diff
@@ -0,0 +1,318 @@
...
+                .sortedByDescending { it.value }
+                .take(limit)
+                .map { it.key }
+                .toSet()
+
+            val nodes = topClasses.mapIndexed { idx, cls ->
+                val shortName = cls.substringAfterLast('.')
+                mapOf(
+                    "id" to cls.hashCode(),
+                    "type" to "Class",
```
johnsonlee, line 208: **Hash collision risk**: Using `String.hashCode()` as a node identifier (lines 209, 216-218) can produce collisions -- two different class names with the same hash will silently merge or filter incorrectly. Consider using the full class name as the ID, or a deterministic counter.

graphite-explore/src/main/kotlin/io/johnsonlee/graphite/cli/ExploreRoutes.kt
```diff
@@ -0,0 +1,318 @@
...
+                    "pattern" to pattern,
+                    "limit" to limit,
+                    "count" to resources.size,
+                    "resources" to resources
+                )
+            )
+        }
+
+        app.get("/api/resources/{path}*") { ctx ->
+            val path = ctx.path().removePrefix("/api/resources/").trimStart('/')
```
johnsonlee: **Fragile path extraction**: `ctx.path().removePrefix("/api/resources/")` couples this handler to the exact route string. If the route ever changes (e.g., versioned API), this silently breaks. Consider using Javalin's path parameter support instead (`/api/resources/{path}*` already captures it).

graphite-explore/src/test/kotlin/io/johnsonlee/graphite/cli/ExploreCommandTest.kt
```diff
@@ -816,6 +1032,923 @@ class ExploreCommandTest {
         assertEquals(0, nodes.size, "Negative depth should return nothing (remaining < 0 check)")
     }
 
+    @Test
+    fun `buildC4Model derives context container and component views`() {
```
johnsonlee: **Integration tests masquerading as unit tests**: Tests like `buildC4Model derives context container and component views` call through `ExploreCommand.buildC4Model()` which chains `C4ModelInferer -> C4StructurizrMapper`. This means a test failure could be in inference, mapping, or both -- no isolation. For a 1980-line inference engine, direct tests against `C4ModelInferer.buildViewModel()` would catch regressions faster and localize failures.

Also, the test fixture is a minimal graph that exercises the happy path but not the interesting inference decisions (competing namespaces, multi-container clustering, library family collapse, etc.).

graphite-explore/src/main/kotlin/io/johnsonlee/graphite/cli/c4/C4Support.kt
```diff
@@ -0,0 +1,163 @@
...
+    edges: List<Map<String, Any?>>,
+    preserveStrongDirectEvidence: Boolean
+): List<Map<String, Any?>> {
+    val reducible = edges.filter { edge ->
+        val kind = edge["kind"]?.toString().orEmpty()
+        kind != "runs-on" || !preserveStrongDirectEvidence
+    }
+    if (reducible.size < 3) return edges
+
+    fun maxAlternatePathCapacity(source: String, destination: String, omitted: Map<String, Any?>): Int {
```
johnsonlee: **O(E^2) worst case**: `maxAlternatePathCapacity()` is called once per edge in the reducible set (line 87), and each call traverses edges via BFS. For E edges, this is O(E^2) in the worst case. Fine for typical diagram sizes (< 50 edges), but worth documenting the expected bound or adding an early-exit if edge count exceeds a threshold.

graphite-explore/src/main/kotlin/io/johnsonlee/graphite/cli/c4/C4ModelInferer.kt
```diff
@@ -0,0 +1,1980 @@
...
+    val classes = (methods.map { it.declaringClass.className } +
+        callSites.flatMap { listOf(it.caller.declaringClass.className, it.callee.declaringClass.className) })
+        .filterNot { isSyntheticArchitectureClass(it) }
+        .distinct()
+        .sorted()
+    val apiEndpoints = apiSpecExtractor.extract(graph)
+    val systemBoundary = deriveSystemBoundary(methods, callSites)
+    val inferred = if (level == "all") {
+        val containerView = buildContainerView(graph, methods, callSites, apiEndpoints, systemBoundary, limit)
+        val componentView = buildComponentView(methods, callSites, apiEndpoints, systemBoundary, limit)
```
johnsonlee: **Double container-layout computation when level="all"**: `buildContainerView()` (line 28) calls `inferContainerLayout(graph, ...)` and then `buildComponentView()` (line 29) calls `inferContainerLayout(graph=null, ...)`. The union-find clustering, traffic analysis, and TF-IDF naming run twice. Consider caching the layout or passing it from the container view into the component view.

graphite-explore/src/main/kotlin/io/johnsonlee/graphite/cli/c4/C4ModelInferer.kt
```diff
@@ -0,0 +1,1980 @@
...
+            "level" to "all",
+            "availableLevels" to C4ArchitectureService.LEVELS,
+            "context" to contextView,
+            "container" to containerView,
+            "component" to componentView
+        )
+    } else {
+        val view = when (level) {
+            "context" -> {
+                val containerView = buildContainerView(graph, methods, callSites, apiEndpoints, systemBoundary, limit)
```
johnsonlee: **Context view always pays for full container inference**: Even when `level="context"`, the code first builds the full container view (line 41), because `buildContextView()` requires the container list. This means a context-level request is never cheaper than a container-level request. If the context view only needs container count (line 426), consider a lighter-weight path.

graphite-explore/src/main/kotlin/io/johnsonlee/graphite/cli/c4/C4ModelInferer.kt
```diff
@@ -0,0 +1,1980 @@
...
+    )
+}
+
+private fun selectClasses(
+    element: Map<String, Any?>,
+    methods: List<MethodDescriptor>,
+    callSites: List<CallSiteNode>,
+    systemBoundary: String
+): Set<String> {
+    val allClasses = (methods.map { it.declaringClass.className } +
```
johnsonlee: **O(M+C) recomputed per overlay element**: `selectClasses()` rebuilds the full set of internal classes from `methods` and `callSites` on every invocation (lines 301-304). In `mergeStructuralOverlay()` this is called once per overlay element (line 217). Hoist the `allClasses` computation outside the loop and pass it in.

graphite-explore/src/main/kotlin/io/johnsonlee/graphite/cli/c4/C4StructurizrMapper.kt
```diff
@@ -0,0 +1,349 @@
...
+        relationships: List<Map<String, Any?>>,
+        sourceView: String
+    ) {
+        relationships.forEach { relationship ->
+            val fromId = relationship["from"]?.toString() ?: return@forEach
+            val toId = relationship["to"]?.toString() ?: return@forEach
+            val description = relationship["description"]?.toString()
+                ?: relationship["type"]?.toString()
+                ?: "uses"
+            val key = Triple(fromId, toId, description)
```
johnsonlee: **Relationship dedup key includes description**: The dedup key `Triple(fromId, toId, description)` means the same source-target pair with different description text (e.g., from context vs container views) creates duplicate Structurizr relationships. Consider using just `Pair(fromId, toId)` or `Triple(fromId, toId, kind)` as the identity.

graphite-explore/src/main/kotlin/io/johnsonlee/graphite/cli/c4/C4StructurizrMapper.kt
```diff
@@ -0,0 +1,349 @@
...
+                relationships = relationshipIdsForView(view),
+                properties = structProperties(
+                    "graphite.level" to level,
+                    "graphite.systemBoundary" to view["systemBoundary"],
+                    "graphite.designOverlay" to graphiteModel["designOverlay"],
+                    "graphite.drift" to view["drift"]
+                )
+            )
+        )
+    } ?: emptyList()
```
johnsonlee: **Component view grouping uses display name, not container ID**: Components are grouped by `it["container"]?.toString()` (display name), and the `scopeId` at line 233 is `"container:$containerName"`. If two containers happen to have the same display name (e.g., both named "Core"), their components merge into one view. Group by `containerId` instead.

graphite-explore/src/main/kotlin/io/johnsonlee/graphite/cli/c4/C4StructurizrMapper.kt
```diff
@@ -0,0 +1,349 @@
...
+
+    val componentViews = componentView?.let { view ->
+        val componentElements = (view["elements"] as? List<Map<String, Any?>>).orEmpty()
+        componentElements.groupBy { it["container"]?.toString().orEmpty() }
```
johnsonlee: **O(R * C) relationship-to-component lookup**: For each `relId`, this scans all `components.values` to find the source component (lines 225-228). Pre-build a `relId -> componentId` index outside the loop.

graphite-explore/src/main/kotlin/io/johnsonlee/graphite/cli/c4/C4DiagramRenderer.kt
```diff
@@ -0,0 +1,580 @@
...
+        val dedupedEdges = reduceSharedLibraryFanIn(
+            reduceSharedInternalFanIn(
+                reduceDiagramTransitiveEdges(
+                    selectedEdges
+                        .distinctBy { "${it["from"]}:${it["to"]}:${it["kind"]}" }
+                        .sortedByDescending { (it["weight"] as? Int) ?: 0 }
+                )
+            )
+        )
+        val truncated = 0
```
johnsonlee: **Dead truncation check**: `val truncated = 0` is hardcoded, then checked on line 247 (`if (truncated > 0)`). This branch is unreachable. Either implement actual truncation tracking for the container view, or remove the dead code.

graphite-explore/src/main/kotlin/io/johnsonlee/graphite/cli/ExploreRoutes.kt
```diff
@@ -0,0 +1,318 @@
...
+
+        app.post("/api/cypher") { ctx ->
+            val body = ctx.body()
+            val query = try {
+                JsonParser.parseString(body).asJsonObject.get("query").asString
+            } catch (e: Exception) {
+                ctx.queryParam("query") ?: run { ctx.status(400).result("Missing 'query' parameter"); return@post }
+            }
+            try {
+                val executor = CypherExecutor(graph)
```
johnsonlee: **CypherExecutor created per request**: A new `CypherExecutor(graph)` is instantiated on every Cypher request (lines 246, 257). If the executor does any setup work (parsing grammar, building indexes), this is wasted on every call. Consider creating it once in the constructor or as a lazy val.

graphite-webgraph/src/main/kotlin/io/johnsonlee/graphite/webgraph/NodeSerializer.kt
```diff
@@ -593,6 +629,28 @@ internal object NodeSerializer {
             enumValues[key] = (0 until count).map { readAnyValue(dis, strings, formatVersion) }
         }
```
johnsonlee, line 628: **Version guard uses FORMAT_VERSION constant, not literal 3**: `if (formatVersion >= FORMAT_VERSION)` works now because `FORMAT_VERSION == 3`. But when v4 is introduced and `FORMAT_VERSION` is bumped to 4, this guard becomes `formatVersion >= 4`, which would skip reading classOrigins/artifactDependencies from v3 files -- a silent data loss on deserialization.

Consider `formatVersion >= 3` (literal) so the guard remains stable across future version bumps. The same pattern applies at line 638.

graphite-sootup/src/main/kotlin/io/johnsonlee/graphite/sootup/SootUpAdapter.kt
```diff
@@ -173,12 +185,22 @@ class SootUpAdapter(
     private var activeMethod: MethodDescriptor? = null
     private var activeMethodLocals = mutableSetOf<LocalKey>()
     private var activeMethodParameters = mutableSetOf<ParameterBinding>()
+    private val persistedClassOrigins: Map<String, String> by lazy {
```
johnsonlee: **Lazy val captures mutable state**: `persistedClassOrigins` is `by lazy` and reads `classOriginSourceCounts` and `classOriginsByName` at first-access time. Currently safe because it is only accessed at line 270, after all indexing is complete. But if a future caller triggers the lazy init earlier (e.g., during `processMethod`), it would freeze an incomplete snapshot. Consider making this an explicit function call at the point where indexing is known to be complete, rather than relying on access ordering.

johnsonlee (commented):
Second pass -- deeper review of mapper, serialization, and inference interactions. No duplicates from the first round.

graphite-explore/src/main/kotlin/io/johnsonlee/graphite/cli/c4/C4ModelInferer.kt
```diff
@@ -0,0 +1,420 @@
...
+        )
+        val thirdPartyDependencies = collapsedContextDependencies.dependencies.filter { it["kind"] == "library" }
+        val externalServiceDependencies = collapsedContextDependencies.dependencies.filter { it["kind"] == "external-system" }
+        val libraryRelationships = inferContextLibraryDependencyRelationships(
+            graph = graph,
+            artifactToContextId = collapsedContextDependencies.artifactToContextId,
+            contextDependencies = thirdPartyDependencies
+    )
+    val runtimeBoundaryLibraries = selectRuntimeBoundaryLibraries(thirdPartyDependencies, libraryRelationships)
+    val elements = buildList<C4Element> {
```
johnsonlee: **Indentation drift in buildContextView**: Lines 103-104 are indented at 4 spaces (function body level), while the `buildList` body starting at line 106 is indented at 12 spaces. The `properties = mapOf(` block at lines 164-167 also has inconsistent indent -- lines 165-167 are at 20 spaces while line 168 jumps to 24. This appears to be a merge/reformat artifact. Run the formatter on this function.

johnsonlee (commented):
Third pass after the major refactoring. Resolved 26/27 previous comments -- the refactoring addressed nearly all prior feedback. One remaining open comment (integration test coverage) and one new issue below.

PR #82, opened 2026-07-25
Optimize explorer long-running memory usage

johnsonlee:
## Summary
- remove the explicit lazy graph load path and keep `MAPPED` as the long-running explorer path instead of moving pressure into first-query/file-read behavior
- move mapped node indexes off JVM heap with mmap-backed `graph.nodeoffsets` and `graph.typeindex`, while preserving `graph.nodeindex` compatibility for older graph directories
- keep forward graph serving query-ready after load: forward BVGraph, labels, cumulative outdegree, and mmap comparison lookup are initialized during mapped load
- persist `graph.labelprefix` from the existing save-time forward offsets so mapped load no longer rebuilds label offsets by scanning every node's outdegree
- fix the incoming-edge residency issue: the first `incoming()` loads existing `backward.*` or builds the transpose once, stores it as compressed BVGraph files when possible, reloads the compressed graph, and lets the temporary uncompressed arrays be collected
- keep `GraphStore.save()` forward-only for adjacency; save writes only existing forward-derived auxiliary files, not backward compression
- use a transient compressed backward graph when the graph directory is not writable, so the current process still avoids retaining the uncompressed transpose
- add incoming explorer waterline coverage and tighten the long-running guardrail to fail on post-warmup heap growth over `16 MiB` or RSS growth over `64 MiB`

## Environment
- Host: Darwin arm64
- JVM: OpenJDK 17.0.18 Homebrew
- Explorer JMH heap: `-Xmx4g`
- WebGraph load/query JMH heap: `-Xmx8g`
- End-to-end JMH heap: `-Xmx4g`
- Android fixture: Robolectric `android-all-14-robolectric-10818077.jar`

## Explorer Memory Benchmarks
Main baseline command:
```
./gradlew :explore:jmh -Pjmh.filter='ExplorerMemoryBenchmark.android_.*' --no-daemon
```

Final waterline command:
```
./gradlew :explore:jmh -Pjmh.filter='ExplorerMemoryBenchmark.android_(incomingExplorerWaterline|longRunningExplorerWaterline)' --no-daemon
```

| Benchmark / metric | Main or prior guardrail | Current | Conclusion |
|--------------------|-------------------------|---------|------------|
| `android_initialExplorerSession` time | main baseline `14451.838 ms/op` | `788.340 ms/op` | much faster |
| `android_initialExplorerSession` retained heap | main baseline `713390624 B` | `1515104 B` | retained heap removed from initial session |
| `android_browserForwardExploration` time | main baseline `2190.773 ms/op` | `836.186 ms/op` | faster |
| `android_browserForwardExploration` retained heap | main baseline `215089672 B` | `578720 B` | retained heap removed from forward browsing |
| `android_longRunningExplorerWaterline` max used heap | first mapped waterline `368439008 B` | `167434568 B` | lower peak heap under `-Xmx4g` |
| `android_longRunningExplorerWaterline` steady used heap before/after | Attempt 021 `97210072 B -> 97123592 B` | `97179960 B -> 97334336 B` | `154376 B` growth after warmup |
| `android_longRunningExplorerWaterline` post-warmup heap growth | guardrail `16777216 B` | `154376 B` | passes |
| `android_longRunningExplorerWaterline` post-warmup RSS growth | guardrail `67108864 B` | `9699328 B` | passes |
| `android_incomingExplorerWaterline` max used heap | Attempt 021 `192419600 B` | `193384752 B` | effectively unchanged |
| `android_incomingExplorerWaterline` steady used heap before/after | Attempt 021 `113776400 B -> 113791136 B` | `113692976 B -> 113706200 B` | `13224 B` growth after warmup |
| `android_incomingExplorerWaterline` post-warmup heap growth | guardrail `16777216 B` | `13224 B` | passes |
| `android_incomingExplorerWaterline` post-warmup RSS growth | guardrail `67108864 B` | `5767168 B` | passes |

Explorer conclusion: lazy mode alone was only relocation. The final incoming path still initializes on demand for forward-only load performance, but after the first incoming traversal the resident graph is compressed `backward.*` or a transient compressed BVGraph, not the uncompressed transpose arrays. Both forward and incoming workloads settle after warmup.

## WebGraph Guardrail Benchmarks
Command:
```
./gradlew :webgraph:jmh -Pjmh.filter='(AndroidLoadBenchmark.mapped_load|AndroidQueryBenchmark.mapped_singleHopRelationship|GraphEndToEndBenchmark.android_build_save_load_query)' --no-daemon
```

| Benchmark | Main / prior | Current | Conclusion |
|-----------|--------------|---------|------------|
| `AndroidLoadBenchmark.mapped_load` | original baseline `2292.892 ms/op`; query-ready Attempt 021 `256.929 ms/op` | `138.768 ms/op` | `~16.5x` faster than original; no deferred first-query load |
| `AndroidQueryBenchmark.mapped_singleHopRelationship` | main `0.701 ms/op`; Attempt 021 `0.667 ms/op` | `0.633 ms/op` | no query regression |
| `GraphEndToEndBenchmark.android_build_save_load_query` | main `115782.725 ms/op`; Attempt 021 `106287.182 ms/op` | `105041.169 ms/op` | no build/save/E2E regression |

Method/load conclusion: `mapped_load` now clears the strict order-of-magnitude target even after mapped graphs are made forward-query-ready during load. The accepted label-prefix change is not the rejected mmap label-offset approach: the hot query path still uses the same heap `IntArray`; only the load-time reconstruction pass is removed.

## Rejected Experiments
| Approach | Result | Why rejected |
|----------|--------|--------------|
| mmap `graph.labeloffsets` | `mapped_singleHopRelationship` regressed to `1.551 ms/op`; Android E2E regressed to `278193.662 ms/op` | moved offset residency but made hot label access slower |
| file-channel node reads | long-running RSS growth was not better than mmap | moved page-cache behavior to syscalls without improving the waterline |
| `BVGraph.loadMapped(forward)` | `mapped_load` regressed to `268.716 ms/op`; `mapped_singleHopRelationship` regressed to `0.800 ms/op` | moved forward graph bytes out of heap but made both load and hot query slower |

## Validation
- `git diff --check`
- `./gradlew :webgraph:test --tests io.johnsonlee.graphite.webgraph.GraphStoreTest :explore:test --tests io.johnsonlee.graphite.cli.ExploreCommandTest :webgraph:detekt :explore:detekt --no-daemon`
- `./gradlew :webgraph:test --tests io.johnsonlee.graphite.webgraph.GraphStoreTest :webgraph:detekt --no-daemon`
- `./gradlew :webgraph:jmh -Pjmh.filter='(AndroidLoadBenchmark.mapped_load|AndroidQueryBenchmark.mapped_singleHopRelationship|GraphEndToEndBenchmark.android_build_save_load_query)' --no-daemon`
- `./gradlew :explore:jmh -Pjmh.filter='ExplorerMemoryBenchmark.android_(incomingExplorerWaterline|longRunningExplorerWaterline)' --no-daemon`
- `./gradlew check -S --no-daemon`
- `./gradlew koverLog --no-daemon`

Coverage from `koverLog`: `core` `98.0592%`, `cypher` `98.0652%`, `explore` `98.0691%`, `query` `100%`, `sootup` `98.2783%`, `webgraph` `98.3513%`.

## Notes
- Persisted graph layout now includes `graph.labelprefix`, `graph.nodeoffsets`, `graph.typeindex`, and optional `backward.*`.
- Older graph directories without `graph.labelprefix` still load; the prefix is rebuilt from `forward.*` as a fallback.
- `GraphStore.save()` still writes only `forward.*` for adjacency; first incoming query pays the one-time backward compression cost.
- Read-only graph directories still get a compressed backward graph in the current process via a transient temp directory, but cannot reuse `backward.*` across later processes.
- Detailed attempt log is in `docs/webgraph-storage.md` Attempts 010-022.


johnsonlee:
Post-merge CI exposed an environment-sensitive test flake in the traversal-cap regression introduced here.

Evidence on exact main `0f3944452cc0d3559c9cffb4b66fbe4475149176`:

- [build attempt 1](https://github.com/johnsonlee/graphite/actions/runs/33315254531) failed in `ExploreCommandTest > buildSubgraph caps excessive traversals` with `StackOverflowError` at `Int2ObjectOpenHashMap.java:327`.
- Rerunning the same workflow on the same SHA passed (attempt 2).
- An independent targeted run on the same SHA also passed.

The test constructs a 2,100-node linear chain and calls the internal helper with `depth = 2_100`, so asserting the 2,000-node cap requires roughly 2,000 recursive JVM frames. Whether that overflows depends on runner/JIT stack-frame details. The production HTTP path clamps subgraph depth to `MAX_SUBGRAPH_DEPTH = 4`, so this is CI/test fragility rather than a #106 product regression.

Please make the cap test deterministic by exceeding 2,000 nodes with a shallow fan-out graph and depth within the production contract (for example, one center with 2,099 outgoing neighbors at depth 1), or make traversal iterative if deep internal calls are intended to be supported. The current linear-chain case should not remain as a gate because reruns can flip its result without any code change.

PR #91, opened 2026-08-26
Optimize cross-graph Cypher search and traversal

johnsonlee:
## Summary

- optimize the exact six-property broad-discovery Cypher query observed on Graphite 2.2.2
- stream filtered `DISTINCT ... LIMIT` rows and route guarded string disjunctions through mapped raw-field/index lookups
- deduplicate unordered property streams by filter ownership, with memory independent of match count
- preserve cross-graph provenance, old mapped graph behavior, generic fallback, and the JVM contract of existing `Graph` implementations
- make mapped-index admission predicate-and-limit-specific, bounded, and reset on LRU eviction
- document `/api/cypher` as the aggregate endpoint so clients do not enumerate graphs and fan out themselves

## Reported workload

```cypher
MATCH (n)
WHERE (exists(n.class) AND n.class CONTAINS 'ThankYou')
   OR (exists(n.name) AND n.name CONTAINS 'ThankYou')
   OR (exists(n.caller_class) AND n.caller_class CONTAINS 'ThankYou')
   OR (exists(n.caller_name) AND n.caller_name CONTAINS 'ThankYou')
   OR (exists(n.callee_class) AND n.callee_class CONTAINS 'ThankYou')
   OR (exists(n.callee_name) AND n.callee_name CONTAINS 'ThankYou')
RETURN DISTINCT n.class AS class, n.name AS name,
    n.caller_class AS caller, n.caller_name AS callerMethod,
    n.callee_class AS callee, n.callee_name AS calleeMethod
LIMIT 120
```

`RETURN DISTINCT` excluded the old filtered-node fast path, and the guarded six-way `OR` could not compile as a direct string filter. The generic path deserialized every unlabeled node, evaluated the full predicate per node, projected all matches, deduplicated them, and applied `LIMIT` last.

The client also enumerated `/api/graphs` and called graph-scoped Cypher routes. That is client-side fan-out. `/api/cypher` already executes one server-side cross-graph query. `/api/graphs` reads cached descriptors, so its timeout during this workload points to server saturation rather than expensive graph-list generation.

## Implementation

- filtered single-node `DISTINCT ... LIMIT` execution retains at most the requested distinct rows while still scanning later graphs to merge complete provenance
- guarded `STARTS WITH`, `ENDS WITH`, and `CONTAINS` disjunctions narrow unlabeled scans to relevant node types and use mapped raw string fields or indexes
- each mapped candidate stream belongs to one filter; later streams suppress nodes matching an earlier filter, so unordered streams are deduplicated without retaining node IDs
- dynamic annotation properties and unsupported/complex expressions remain on the generic evaluator
- storage lookup is an optional `StringPropertyLookup` capability plus a safe `Graph` extension; `Graph.class` does not gain a JVM method, preserving compatibility with implementations compiled against 2.2.2
- finite index admission is keyed by node type, property, match mode, expected string, and lookup limit; a costly large-page scan cannot make a later `LIMIT 1` request build an index
- admission retains at most 32 predicates / 64 KiB and is cleared when a property index leaves the four-entry LRU

## Benchmark environment

Apple M3 Max, 64 GiB RAM, macOS 14.3 arm64, OpenJDK 17.0.18, JMH 1.37, one benchmark thread and one fork. Baseline and branch runs used the same machine and dependency caches. `main` was commit `e4d1c6a`.

## Android-scale benchmark

The primary pressure benchmark uses the Android fixture JAR: 5.9 million nodes and about 6.5 million edges. `AndroidBroadDiscoveryBenchmark` runs the six-property query shape with the common term `android` and validates all 120 rows. Its name describes the workload, not its original caller. The identical benchmark source was run on `main` and this branch.

```shell
./gradlew :webgraph:jmh \
  -Pjmh.filter='AndroidBroadDiscoveryBenchmark.*' \
  --no-daemon
```

| Exact benchmark | `main` | branch | speedup |
|---|---:|---:|---:|
| `AndroidBroadDiscoveryBenchmark.coldBroadDiscovery` | 7,426.015 ms/op | 207.392 ms/op | 35.81x |
| `AndroidBroadDiscoveryBenchmark.repeatedBroadDiscovery` | 7,433.167 ms/op | 204.300 ms/op | 36.38x |

The final memory fix preserves the Android-scale speedup. The deterministic 16-graph `BroadDiscoveryMappedQueryBenchmark` is supporting evidence: repeated `2.861 ms/op`, cold hit `11.661 ms/op`, and cold miss `7.949 ms/op`. The earlier cold profiler measured allocation falling from 137.930 to 25.203 MB/op.

## Review fixes

| Admission control | `main` | branch |
|---|---:|---:|
| different predicate after an admitted full miss | 0.555 ms/op | 0.390 ms/op |
| same predicate after a large-limit scan | 0.518 ms/op | 0.463 ms/op |
| lookup after four-entry LRU eviction | 0.189 ms/op | 0.263 ms/op |

The earlier branch spikes were 25.074 ms, 21.339 ms, and 9.810 ms respectively. Admission now isolates both predicate and lookup budget; all three controls are at `main` latency. The same-predicate JMH uses `CONTAINS 'path-0'` on both sides, changes only `LIMIT 50000` to `LIMIT 1`, and asserts the index remains absent after setup and measurement.

The eager candidate union originally retained and sorted every match. A later primitive-ID-set version fixed lazy single-graph limits but could retain every match during qualified cross-graph provenance scans. The final filter-owned merge retains no node IDs. Its regressions prove:

- a single-graph `LIMIT 1` consumes one candidate rather than all 1,000
- explicitly unordered streams produce IDs `[1, 2, 0]` exactly once each
- a real `save -> loadMapped` unordered type index returns `[4, 90]`, not `[4, 90, 4]`, for `RETURN DISTINCT n.id, rand()`
- two qualified graphs with 5,000 nodes each drain all 20,000 indexed candidates, return one row, merge both graph IDs, and use no match-sized deduplication state

An ABI regression asserts that `Graph.class` has no `nodesByStringProperty` method and that graphs without the optional capability use the generic fallback rather than throwing `AbstractMethodError`.

## Method-level regression

```shell
./gradlew :cypher:jmh \
  -Pjmh.filter='io.johnsonlee.graphite.cypher.CypherBenchmark.*' \
  --no-daemon
```

| Exact `CypherBenchmark` method | `main` | branch |
|---|---:|---:|
| `aggregationCountGroupBy` | 34.443 us/op | 34.449 us/op |
| `countStar` | 2.263 us/op | 2.284 us/op |
| `functionCalls` | 24.868 us/op | 24.683 us/op |
| `nodeMatchWithWhere` | 58.748 us/op | 57.674 us/op |
| `regexFilter` | 23.571 us/op | 24.045 us/op |
| `returnDistinct` | 103.294 us/op | 102.030 us/op |
| `simpleNodeMatch` | 20.418 us/op | 20.417 us/op |
| `singleHopRelationship` | 30.085 us/op | 29.920 us/op |
| `variableLengthPath` | 26.093 us/op | 26.071 us/op |
| `withPipeline` | 76.079 us/op | 76.587 us/op |

The confidence intervals overlap, so there is no measured method-level regression.

## End-to-end and ordinary-query regression

```shell
./gradlew :webgraph:jmh \
  -Pjmh.filter='GraphEndToEndBenchmark.android_build_save_load_query' \
  --no-daemon
```

| Exact benchmark | `main` | branch |
|---|---:|---:|
| `GraphEndToEndBenchmark.android_build_save_load_query` | 30,182.415 ms/op | 23,877.213 ms/op |

This single shot covers Android JAR -> build -> save -> mapped load -> Cypher query and shows no end-to-end regression.

Ordinary `AndroidQueryBenchmark` mapped queries remain in the same sub-millisecond range; all branch and `main` confidence intervals overlap.

| Android mapped query | `main` | branch |
|---|---:|---:|
| count | 0.003 ms/op | 0.002 ms/op |
| int filter | 0.172 ms/op | 0.182 ms/op |
| distinct | 0.171 ms/op | 0.209 ms/op |
| limited node match | 0.070 ms/op | 0.087 ms/op |
| single hop | 0.651 ms/op | 0.653 ms/op |

## Limits

This is query planning and allocation control, not a hard CPU budget. Arbitrary Cypher can still request a large scan. Request deadlines, cooperative cancellation, query cost limits, and a Cypher concurrency bulkhead remain separate availability work.

## Verification

```shell
./gradlew check -S --no-daemon
./gradlew koverLog --rerun-tasks --no-daemon
./gradlew :cypher:test :cypher:detekt :webgraph:test :webgraph:detekt :webgraph:jmhJar --no-daemon
```

All module tests, baseline-aware detekt tasks, JMH compilation, and Kover verification pass. Application line coverage is Core 98.3471%, Cypher 98.0897%, and WebGraph 98.0072%.

Regression tests cover the exact broad query, cross-graph provenance and full-source consumption, unordered mapped candidate deduplication, dynamic-property fallback, ABI fallback, predicate/limit admission bounds, clear/reset behavior, and LRU eviction.

### DIFF
--- a/graphite-core/src/main/kotlin/io/johnsonlee/graphite/graph/Graph.kt
+++ b/graphite-core/src/main/kotlin/io/johnsonlee/graphite/graph/Graph.kt
@@ -20,6 +20,12 @@ data class ClassOverview(
     val callSiteCount: Int
 )
 
+enum class StringMatchMode {
+    STARTS_WITH,
+    ENDS_WITH,
+    CONTAINS
+}
+
 /**
  * The unified program graph that combines all analysis graphs.
  * This is the central abstraction of Graphite.
@@ -47,6 +53,19 @@ interface Graph {
      * to indicate callers should fall back to [nodes].
      */
     fun nodeCount(type: Class<out Node>): Long? = null
+
+    /**
+     * Return nodes whose string [property] matches [expected] when the graph has
+     * a storage-aware access path. Implementations may return null when the type
+     * or property is unsupported, or when the access path is not ready, so
+     * callers must fall back to [nodes].
+     */
+    fun <T : Node> nodesByStringProperty(
+        type: Class<T>,
+        property: String,
+        mode: StringMatchMode,
+        expected: String
+    ): Sequence<T>? = null
 
     /**
      * Return a precomputed edge count when the graph can answer without
--- a/graphite-core/src/main/kotlin/io/johnsonlee/graphite/graph/Graph.kt
+++ b/graphite-core/src/main/kotlin/io/johnsonlee/graphite/graph/Graph.kt
@@ -26,7 +26,10 @@ enum class StringMatchMode {
     CONTAINS
 }
 
-/** Optional storage capability for graphs that can avoid materializing a full node scan. */
+/**
+ * Optional storage capability for graphs that can avoid materializing a full node scan.
+ * Matching nodes must be returned in ascending [Node.id] order.
+ */
 interface StringPropertyLookup {
     fun <T : Node> nodesByStringProperty(
         type: Class<T>,
--- a/graphite-core/src/main/kotlin/io/johnsonlee/graphite/graph/Graph.kt
+++ b/graphite-core/src/main/kotlin/io/johnsonlee/graphite/graph/Graph.kt
@@ -26,10 +26,7 @@ enum class StringMatchMode {
     CONTAINS
 }
 
-/**
- * Optional storage capability for graphs that can avoid materializing a full node scan.
- * Matching nodes must be returned in ascending [Node.id] order.
- */
+/** Optional storage capability for graphs that can avoid materializing a full node scan. */
 interface StringPropertyLookup {
     fun <T : Node> nodesByStringProperty(
         type: Class<T>,
--- a/graphite-core/src/main/kotlin/io/johnsonlee/graphite/graph/Graph.kt
+++ b/graphite-core/src/main/kotlin/io/johnsonlee/graphite/graph/Graph.kt
@@ -58,13 +58,15 @@ interface Graph {
      * Return nodes whose string [property] matches [expected] when the graph has
      * a storage-aware access path. Implementations may return null when the type
      * or property is unsupported, or when the access path is not ready, so
-     * callers must fall back to [nodes].
+     * callers must fall back to [nodes]. At most [limit] matching nodes are
+     * returned when a finite limit is supplied.
      */
     fun <T : Node> nodesByStringProperty(
         type: Class<T>,
         property: String,
         mode: StringMatchMode,
-        expected: String
+        expected: String,
+        limit: Int = Int.MAX_VALUE
     ): Sequence<T>? = null
 
     /**
--- a/graphite-core/src/main/kotlin/io/johnsonlee/graphite/graph/Graph.kt
+++ b/graphite-core/src/main/kotlin/io/johnsonlee/graphite/graph/Graph.kt
@@ -26,6 +26,17 @@ enum class StringMatchMode {
     CONTAINS
 }
 
+/** Optional storage capability for graphs that can avoid materializing a full node scan. */
+interface StringPropertyLookup {
+    fun <T : Node> nodesByStringProperty(
+        type: Class<T>,
+        property: String,
+        mode: StringMatchMode,
+        expected: String,
+        limit: Int
+    ): Sequence<T>?
+}
+
 /**
  * The unified program graph that combines all analysis graphs.
  * This is the central abstraction of Graphite.
@@ -53,21 +64,6 @@ interface Graph {
      * to indicate callers should fall back to [nodes].
      */
     fun nodeCount(type: Class<out Node>): Long? = null
-
-    /**
-     * Return nodes whose string [property] matches [expected] when the graph has
-     * a storage-aware access path. Implementations may return null when the type
-     * or property is unsupported, or when the access path is not ready, so
-     * callers must fall back to [nodes]. At most [limit] matching nodes are
-     * returned when a finite limit is supplied.
-     */
-    fun <T : Node> nodesByStringProperty(
-        type: Class<T>,
-        property: String,
-        mode: StringMatchMode,
-        expected: String,
-        limit: Int = Int.MAX_VALUE
-    ): Sequence<T>? = null
 
     /**
      * Return a precomputed edge count when the graph can answer without
@@ -200,6 +196,20 @@ interface Graph {
     fun classOverview(limit: Int): ClassOverview? = null
 
 }
+
+/**
+ * Use a storage-aware string lookup when [Graph] also implements
+ * [StringPropertyLookup]. Existing graph implementations remain binary
+ * compatible and return null through this extension.
+ */
+fun <T : Node> Graph.nodesByStringProperty(
+    type: Class<T>,
+    property: String,
+    mode: StringMatchMode,
+    expected: String,
+    limit: Int = Int.MAX_VALUE
+): Sequence<T>? = (this as? StringPropertyLookup)
+    ?.nodesByStringProperty(type, property, mode, expected, limit)
 
 /**
  * Pattern for matching methods.
--- a/graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
+++ b/graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
@@ -399,7 +399,8 @@ class QueryPipeline private constructor(
                 nodeClass,
                 filter.property,
                 filter.mode,
-                filter.expected
+                filter.expected,
+                limit - rows.size
             )
             val candidates = indexedNodes ?: source.graph.nodes(nodeClass).filter(filter::matches)
             for (node in candidates) {
--- a/graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
+++ b/graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
@@ -529,13 +529,31 @@ class QueryPipeline private constructor(
             val candidates = if (accelerated.any { it == null }) {
                 graph.nodes(candidateType).filter(disjunction::matches)
             } else {
-                accelerated.asSequence().filterNotNull().flatten()
+                mergeSortedNodes(accelerated.filterNotNull())
             }
-            val unique = LinkedHashMap<Int, Node>()
-            candidates.forEach { node -> unique.putIfAbsent(node.id.value, node) }
-            unique.values.sortedBy { it.id.value }.forEach { yield(it) }
+            for (node in candidates) yield(node)
         }
     }
+
+    private fun <T : Node> mergeSortedNodes(sequences: List<Sequence<T>>): Sequence<Node> = sequence {
+        val iterators = sequences.map { it.iterator() }
+        val heads = iterators.map { iterator -> iterator.nextOrNull() }.toMutableList()
+        while (true) {
+            var next: Node? = null
+            for (head in heads) {
+                if (head != null && (next == null || head.id.value < next.id.value)) next = head
+            }
+            next ?: break
+            yield(next)
+            heads.indices.forEach { index ->
+                while (heads[index]?.id == next.id) {
+                    heads[index] = iterators[index].nextOrNull()
+                }
+            }
+        }
+    }
+
+    private fun <T> Iterator<T>.nextOrNull(): T? = if (hasNext()) next() else null
 
     private data class DirectStringFilter(
         val property: String,
--- a/graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
+++ b/graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
@@ -15,6 +15,7 @@ import io.johnsonlee.graphite.core.ResourceEdge
 import io.johnsonlee.graphite.core.TypeEdge
 import io.johnsonlee.graphite.graph.Graph
 import io.johnsonlee.graphite.graph.StringMatchMode
+import io.johnsonlee.graphite.graph.nodesByStringProperty
 
 private const val COUNT_QUERY_CLAUSES = 2
 private const val DISTINCT_LIMIT_QUERY_CLAUSES = 3
@@ -90,11 +91,14 @@ class QueryPipeline private constructor(
         val earlyLimit = computeEarlyLimit(clauses)
 
         var consumedWhereIndex = -1
-        for ((clauseIndex, clause) in clauses.withIndex()) {
+        for (clauseIndex in clauses.indices) {
+            val clause = clauses[clauseIndex]
             when (clause) {
                 is CypherClause.Match -> {
                     val pushedWhere = clauses.getOrNull(clauseIndex + 1) as? CypherClause.Where
-                    val soughtRows = pushedWhere?.let { tryElementIdSeek(clause, it, rows) }
+                    val soughtRows = pushedWhere
+                        ?.takeIf { it.condition is CypherExpr.Comparison }
+                        ?.let { tryElementIdSeek(clause, it, rows) }
                     rows = if (soughtRows != null) {
                         consumedWhereIndex = clauseIndex + 1
                         soughtRows
@@ -375,7 +379,8 @@ class QueryPipeline private constructor(
         val nodeClass = nodePattern.labels.firstOrNull()
             ?.let { NodePropertyAccessor.resolveNodeLabel(it) }
             ?: Node::class.java
-        val directStringFilter = DirectStringFilter.compile(where.condition, variable)
+        val directStringFilter = (where.condition as? CypherExpr.StringOp)
+            ?.let { DirectStringFilter.compile(it, variable) }
         if (!ret.distinct && directStringFilter != null && nodePattern.labels.size <= 1 && nodePattern.properties.isEmpty()) {
             return executeDirectStringFilter(
                 nodeClass,
@@ -386,18 +391,18 @@ class QueryPipeline private constructor(
                 limitCount
             )
         }
-        val directStringDisjunction = DirectStringDisjunction.compile(where.condition, variable)
-        if (ret.distinct && directStringDisjunction != null &&
-            nodePattern.labels.size <= 1 && nodePattern.properties.isEmpty()
-        ) {
-            return executeDirectStringDisjunction(
-                nodeClass,
-                variable,
-                directStringDisjunction,
-                ret.items,
-                columns,
-                limitCount
-            )
+        if (ret.distinct && nodePattern.labels.size <= 1 && nodePattern.properties.isEmpty()) {
+            val directStringDisjunction = DirectStringDisjunction.compile(where.condition, variable)
+            if (directStringDisjunction != null) {
+                return executeDirectStringDisjunction(
+                    nodeClass,
+                    variable,
+                    directStringDisjunction,
+                    ret.items,
+                    columns,
+                    limitCount
+                )
+            }
         }
 
         val rows = mutableListOf<Map<String, Any?>>()
--- a/graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
+++ b/graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
@@ -1,9 +1,14 @@
 package io.johnsonlee.graphite.cypher
 
+import io.johnsonlee.graphite.core.AnnotationNode
 import io.johnsonlee.graphite.core.CallEdge
+import io.johnsonlee.graphite.core.CallSiteNode
 import io.johnsonlee.graphite.core.ControlFlowEdge
 import io.johnsonlee.graphite.core.DataFlowEdge
 import io.johnsonlee.graphite.core.Edge
+import io.johnsonlee.graphite.core.EnumConstant
+import io.johnsonlee.graphite.core.FieldNode
+import io.johnsonlee.graphite.core.LocalVariable
 import io.johnsonlee.graphite.core.Node
 import io.johnsonlee.graphite.core.NodeId
 import io.johnsonlee.graphite.core.ResourceEdge
@@ -18,6 +23,20 @@ private const val SINGLE_HOP_LIMIT_QUERY_CLAUSES = 3
 private const val SINGLE_HOP_PATTERN_ELEMENTS = 3
 private const val SINGLE_GRAPH_ID = "single"
 private val QUALIFIED_NODE_PROPERTIES = setOf(GRAPH_ID_PROPERTY, ELEMENT_ID_PROPERTY, QUALIFIED_ID_PROPERTY)
+private val DIRECT_STRING_NODE_PROPERTIES = listOf(
+    EnumConstant::class.java to setOf("name"),
+    LocalVariable::class.java to setOf("name"),
+    FieldNode::class.java to setOf("class", "name"),
+    CallSiteNode::class.java to setOf("caller_class", "caller_name", "callee_class", "callee_name"),
+    AnnotationNode::class.java to setOf(
+        "class",
+        "name",
+        "caller_class",
+        "caller_name",
+        "callee_class",
+        "callee_name"
+    )
+)
 
 /**
  * Executes a sequence of [CypherClause] elements against a [Graph],
@@ -339,7 +358,7 @@ class QueryPipeline private constructor(
         val where = clauses[1] as? CypherClause.Where ?: return null
         val ret = clauses[2] as? CypherClause.Return ?: return null
         val limit = clauses[3] as? CypherClause.Limit ?: return null
-        if (match.optional || ret.distinct || match.patterns.size != 1 || ret.items.any { containsAggregation(it.expression) }) {
+        if (match.optional || match.patterns.size != 1 || ret.items.any { containsAggregation(it.expression) }) {
             return null
         }
         if (ret.items.any { (it.expression as? CypherExpr.Variable)?.name == "*" }) return null
@@ -357,7 +376,7 @@ class QueryPipeline private constructor(
             ?.let { NodePropertyAccessor.resolveNodeLabel(it) }
             ?: Node::class.java
         val directStringFilter = DirectStringFilter.compile(where.condition, variable)
-        if (directStringFilter != null && nodePattern.labels.size <= 1 && nodePattern.properties.isEmpty()) {
+        if (!ret.distinct && directStringFilter != null && nodePattern.labels.size <= 1 && nodePattern.properties.isEmpty()) {
             return executeDirectStringFilter(
                 nodeClass,
                 variable,
@@ -367,8 +386,26 @@ class QueryPipeline private constructor(
                 limitCount
             )
         }
+        val directStringDisjunction = DirectStringDisjunction.compile(where.condition, variable)
+        if (ret.distinct && directStringDisjunction != null &&
+            nodePattern.labels.size <= 1 && nodePattern.properties.isEmpty()
+        ) {
+            return executeDirectStringDisjunction(
+                nodeClass,
+                variable,
+                directStringDisjunction,
+                ret.items,
+                columns,
+                limitCount
+            )
+        }
 
         val rows = mutableListOf<Map<String, Any?>>()
+        val distinctRows = if (ret.distinct) {
+            LinkedHashMap<Map<String, Any?>, MutableMap<String, Any?>>()
+        } else {
+            null
+        }
         val predicateBindings = mutableMapOf<String, Any?>(variable to null)
         for (candidate in nodeCandidates(nodeClass)) {
             if (!matchesNodeConstraints(candidate, nodePattern, emptyMap())) continue
@@ -378,11 +415,36 @@ class QueryPipeline private constructor(
 
             val bindings = mutableMapOf<String, Any?>(variable to candidate)
             addProvenance(bindings, candidate)
-            rows.add(projectRow(ret.items, columns, bindings))
-            if (rows.size >= limitCount) return CypherResult(columns, rows)
+            val projected = projectRow(ret.items, columns, bindings)
+            if (distinctRows == null) {
+                rows.add(projected)
+                if (rows.size >= limitCount) return CypherResult(columns, rows)
+            } else {
+                addDistinctRow(distinctRows, projected, limitCount)
+                if (!qualified && distinctRows.size >= limitCount) {
+                    return CypherResult(columns, distinctRows.values.toList())
+                }
+            }
         }
 
-        return CypherResult(columns, rows)
+        return CypherResult(columns, distinctRows?.values?.toList() ?: rows)
+    }
+
+    @Suppress("UNCHECKED_CAST")
+    private fun addDistinctRow(
+        rows: LinkedHashMap<Map<String, Any?>, MutableMap<String, Any?>>,
+        row: Map<String, Any?>,
+        limit: Int
+    ) {
+        val visible = row.filterKeys { it != INTERNAL_PROVENANCE_KEY }
+        val existing = rows[visible]
+        if (existing != null) {
+            val graphIds = (existing[INTERNAL_PROVENANCE_KEY] as? Set<String>).orEmpty() +
+                (row[INTERNAL_PROVENANCE_KEY] as? Set<String>).orEmpty()
+            if (graphIds.isNotEmpty()) existing[INTERNAL_PROVENANCE_KEY] = graphIds
+        } else if (rows.size < limit) {
+            rows[visible] = row.toMutableMap()
+        }
     }
 
     private fun executeDirectStringFilter(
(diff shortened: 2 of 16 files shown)


graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
```diff
@@ -277,21 +355,98 @@ class QueryPipeline private constructor(
         val nodeClass = nodePattern.labels.firstOrNull()
             ?.let { NodePropertyAccessor.resolveNodeLabel(it) }
             ?: Node::class.java
+        val directStringFilter = DirectStringFilter.compile(where.condition, variable)
+        if (directStringFilter != null && nodePattern.labels.size <= 1 && nodePattern.properties.isEmpty()) {
```
johnsonlee: **Blocking — this fast path changes cross-graph query semantics.** `DirectStringFilter.compile` accepts virtual qualified properties such as `graphId`, `elementId`, and `qualifiedId`, but `executeDirectStringFilter` evaluates its fallback against the raw `Node`, where those properties do not exist. I reproduced `MATCH (n:StringConstant) WHERE n.graphId STARTS WITH 'ord' RETURN elementId(n) AS id LIMIT 1`: `main` returns `orders:1`, while this PR returns no rows. Please exclude qualified virtual properties from this path or evaluate/revalidate against the `QualifiedNode`, and add regression tests for all three properties.

graphite-webgraph/src/main/kotlin/io/johnsonlee/graphite/webgraph/MappedWebGraphBackedGraph.kt
```diff
@@ -97,6 +118,31 @@ internal class MappedWebGraphBackedGraph(
...
+        val key = StringPropertyKey(type, property)
+        val index = synchronized(stringPropertyIndexLock) {
+            stringPropertyIndexes[key] ?: run {
+                val accesses = (stringPropertyAccesses[key] ?: 0) + 1
+                stringPropertyAccesses[key] = accesses
+                trimStringPropertyAccesses()
+                if (accesses < MIN_STRING_PROPERTY_INDEX_ACCESSES) {
+                    null
+                } else {
+                    buildStringPropertyIndex(type, property).also { stringPropertyIndexes[key] = it }
```
johnsonlee: **Blocking performance regression.** The second access unconditionally builds the complete property/trigram index, even when the existing scan reaches `LIMIT 1` at the first node. In a same-machine JMH with 100,000 mapped nodes and two early-hit `LIMIT 1` queries, `main` measured `0.029 ms/op` and `79 KB/op`; this PR measured `21.922 ms/op` and `39.9 MB/op` (about 756x slower and 504x more allocation). The PR benchmark only covers late hits/misses. Please use an admission policy that accounts for scan cost/selectivity, and add this cold early-hit control before claiming no regression.

graphite-webgraph/src/main/kotlin/io/johnsonlee/graphite/webgraph/MappedWebGraphBackedGraph.kt
```diff
@@ -192,15 +238,246 @@ internal class MappedWebGraphBackedGraph(
...
+                StringMatchMode.ENDS_WITH -> actual.endsWith(expected)
+                StringMatchMode.CONTAINS -> actual.contains(expected)
+            }
+            if (matches) matched[size++] = stringId
+        }
+        return matched.copyOf(size)
+    }
+
+    private fun buildTrigramIndex(): Int2ObjectOpenHashMap<IntArray>? {
+        if (uniqueStringIds.size > MAX_TRIGRAM_INDEX_STRINGS) return null
```
johnsonlee, line 626: The `500,000` distinct-string guard does not make this index memory-bounded: posting count scales with the sum of distinct trigrams across all string contents, so legal long values can allocate arbitrarily large builders plus final arrays. The four-index LRU is a count bound, not a byte bound, and it applies per graph. Please add a posting/byte budget with an abort-to-scan fallback (and retained-heap/RSS evidence) before describing this as bounded.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
```diff
@@ -119,6 +128,75 @@ class QueryPipeline private constructor(
...
+
+    @Suppress("ReturnCount")
+    private fun seekNode(elementId: String): Pair<CypherGraph, Node>? {
+        if (!qualified) {
+            val nodeId = elementId.toIntOrNull() ?: return null
+            return sources.single().let { source -> source.graph.node(NodeId(nodeId))?.let { source to it } }
+        }
+
+        val separator = elementId.lastIndexOf(':')
+        if (separator <= 0 || separator == elementId.lastIndex) return null
```
johnsonlee: This rejects `separator == 0`, but the public `CypherGraph` API currently permits an empty graph namespace. Such a node has a valid existing `elementId` like `:1`; `main` matches `WHERE elementId(n) = ':1'`, whereas this seek returns no rows and consumes the `WHERE`. Either reject blank graph IDs at construction as an intentional API change, or preserve the existing empty-namespace behavior here. Please add the corresponding regression test.

johnsonlee (commented):
Blocking review: the target-workload speedups are reproducible, but this implementation currently changes existing query semantics and introduces a severe mapped-query regression for early-hit workloads. I reproduced both functional failures against main and measured the early-hit case with JMH. Please address the blocking inline findings and add regression coverage before merging. GitHub does not permit the PR author to submit a formal Request changes review on their own PR.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
johnsonlee: Fixed in a9f9e29. `DirectStringFilter.compile` now rejects `graphId`, `elementId`, and `qualifiedId`, so those predicates stay on the generic evaluator with a `QualifiedNode` binding. `QueryPipelineTest.string filters preserve virtual qualified properties` asserts concrete results for STARTS WITH / ENDS WITH / CONTAINS across all three virtual properties. `./gradlew check -S --no-daemon` passes.

graphite-webgraph/src/main/kotlin/io/johnsonlee/graphite/webgraph/MappedWebGraphBackedGraph.kt
johnsonlee: Fixed in ca59dac. Cypher passes the remaining LIMIT to mapped lookup. Before an index exists, finite queries lazily scan raw mmap string IDs and materialize only matches; satisfying LIMIT within 256 nodes records no admission, while a longer scan makes the property eligible for indexing on the next access. The new cold-two-early control is 0.038 ms/op and 96.860 KB/op versus 0.038 ms/op and 96.723 KB/op on the identical main fixture. Cold late-hit improved to 4.853 ms/op and the two-query build case to 18.438 ms/op. Tests assert two early limited lookups leave index count at zero and a long scan builds on the next access.

graphite-webgraph/src/main/kotlin/io/johnsonlee/graphite/webgraph/MappedWebGraphBackedGraph.kt
johnsonlee: Fixed in ca59dac. Each property now has independent conservative limits: 8 MiB base ID arrays, 16 MiB / 1,000,000 postings for trigram construction, and 2 MiB / 32 entries for predicate results (including the retained query string). Trigram overflow discards the partial builder and scans the unique-string dictionary; base-array rejection keeps finite queries on raw-field scan. A forced zero-posting-budget test verifies identical hits/misses through fallback. After Full GC with one index in each of 16 mapped graphs, live heap was 16,949 KiB versus 11,918 KiB on main (+5,031 KiB); RSS was 233,056 versus 178,592 KiB, with the larger delta tracking G1 committed heap (131,072 versus 69,632 KiB). The estimated hard bound for these caches is 26 MiB/property and 104 MiB/mapped graph under the existing four-entry LRU.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
johnsonlee: Fixed in a9f9e29 by accepting a separator at offset zero while still rejecting a missing separator or missing local ID. `QueryPipelineTest.element id seek preserves an empty graph namespace` asserts that `WHERE elementId(n) = :1` returns the concrete `:1` ID. The public empty-namespace behavior is preserved.

graphite-webgraph/src/main/kotlin/io/johnsonlee/graphite/webgraph/MappedWebGraphBackedGraph.kt
```diff
@@ -81,6 +91,17 @@ internal class MappedWebGraphBackedGraph(
...
     }
 
+    private val stringPropertyIndexLock = Any()
+    private val stringPropertyAccesses = LinkedHashMap<StringPropertyKey, Int>()
+    private val stringPropertyIndexes = object : LinkedHashMap<StringPropertyKey, MappedStringPropertyIndex>(
+        MAX_STRING_PROPERTY_INDEXES + 1,
+        STRING_PROPERTY_INDEX_LOAD_FACTOR,
+        true
+    ) {
+        override fun removeEldestEntry(eldest: MutableMap.MutableEntry<StringPropertyKey, MappedStringPropertyIndex>?): Boolean =
```
johnsonlee: **Blocking — eviction leaves this admission state hot.** `removeEldestEntry` removes only `stringPropertyIndexes`; `stringPropertyAccesses[eldest.key]` remains `1`. The next finite lookup therefore skips the raw scan and rebuilds the entire index, even for a first-node `LIMIT 1`. In an independent JMH with 100,000 `ResourceFile` and 100,000 `Field` nodes, admitting/building 5 keys against this 4-entry LRU and then querying the evicted key's first node measured **9.810 ms/op** on this PR versus **0.153 ms/op** on `main` (about **64x** slower). Please reset/remove admission state on eviction (or otherwise avoid rebuilding before observed scan cost justifies it) and add an eviction-to-early-hit regression control.

graphite-core/src/main/kotlin/io/johnsonlee/graphite/graph/Graph.kt
```diff
@@ -48,6 +54,21 @@ interface Graph {
...
     fun nodeCount(type: Class<out Node>): Long? = null
 
+    /**
+     * Return nodes whose string [property] matches [expected] when the graph has
+     * a storage-aware access path. Implementations may return null when the type
+     * or property is unsupported, or when the access path is not ready, so
+     * callers must fall back to [nodes]. At most [limit] matching nodes are
+     * returned when a finite limit is supplied.
+     */
+    fun <T : Node> nodesByStringProperty(
```
johnsonlee: **Blocking — this breaks binary compatibility for existing `Graph` implementations.** With the current Kotlin compiler settings this source default is emitted as an abstract interface method plus `Graph$DefaultImpls`. I packaged `DefaultGraph*.class` compiled from `main` with this branch's jars; `new CypherExecutor(oldGraph).execute(...)` fails deterministically with `AbstractMethodError` at `QueryPipeline.executeDirectStringFilter` because the old class has no `nodesByStringProperty(..., int)` method. Third-party implementations compiled against the released interface will fail the same way when this fast path is reached. Please preserve binary compatibility (for example, emit a real JVM default or avoid adding a new abstract interface method) and add a compatibility regression check.

johnsonlee (commented):
Changes still requested after follow-up review of `2ce5e226`. The virtual-property, empty-namespace, and retained-budget fixes are verified, and a clean `./gradlew check -S --no-daemon` passes. Two merge blockers remain inline: index eviction/admission still causes a large early-hit regression, and the public `Graph` API change breaks existing compiled implementations. The original performance thread remains unresolved.

graphite-webgraph/src/main/kotlin/io/johnsonlee/graphite/webgraph/MappedWebGraphBackedGraph.kt
johnsonlee: Follow-up on `2ce5e226`: the new cold early→early control fixes the exact original reproduction, but the admission policy still has a blocking early-hit regression. With `clear → one full miss (admits) → first-node LIMIT 1` on the same 100k fixture, this PR measured **25.074 ms/op / 44.99 MB** versus `main` at **13.008 ms/op / 30.47 MB** (1.93x slower, +47.7% allocation). Once the long scan writes `stringPropertyAccesses[key] = 1`, the next request builds the full index before considering that it hits immediately. A separate inline comment documents the even larger LRU-eviction case. Keeping this thread open until admission no longer turns subsequent selective queries into mandatory full builds.

johnsonlee (commented):
Changes still requested on `279bdde`. This commit does not modify `Graph.kt` or `MappedWebGraphBackedGraph.kt`, so the open admission/LRU and binary-compatibility blockers remain reproducible. A clean `./gradlew check -S --no-daemon` passes and the new broad-discovery JMH results are reproducible, but the new fast path also introduces a blocking non-streaming materialization issue inline.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
```diff
@@ -277,21 +375,242 @@ class QueryPipeline private constructor(
...
+                    completeScanLimit
+                )
+            }
+            val candidates = if (accelerated.any { it == null }) {
+                graph.nodes(candidateType).filter(disjunction::matches)
+            } else {
+                accelerated.asSequence().filterNotNull().flatten()
+            }
+            val unique = LinkedHashMap<Int, Node>()
+            candidates.forEach { node -> unique.putIfAbsent(node.id.value, node) }
```
johnsonlee: **Blocking — this path is not actually streaming.** `candidates.forEach` exhausts every property lookup and retains every matching `Node` in `unique`; the following `sortedBy` also materializes and sorts the complete set before the first `yield`. `executeDirectStringDisjunction` cannot enforce `LIMIT` until after all of that work. In a counting-graph regression with 1,000 matching `CallSiteNode`s, `RETURN DISTINCT ... LIMIT 1` consumed all **1,000** nodes before returning one row. On a large mapped graph, a common or empty keyword therefore causes O(all matches) deserialization/heap retention plus O(M log M) sorting and can recreate the saturation/OOM behavior this commit is intended to avoid. Please lazily union/emit candidates and add a consumption-based test; the current test only asserts the final row count.

graphite-webgraph/src/main/kotlin/io/johnsonlee/graphite/webgraph/MappedWebGraphBackedGraph.kt
johnsonlee: Still reproducible at `279bdde`: the same 5-key/4-entry eviction benchmark measures **3.356 ± 0.489 ms/op** versus `main` at **0.058 ± 0.017 ms/op** (about **57.9x**). This new broad-discovery path makes the bug more important on mixed graphs: Enum.name + Local.name + Field.class/name + four CallSite properties can touch 8 index keys while only 4 are retained, so stale admission state can force a rebuild/eviction cycle on every repeated query. No mapped-index code changed in this head.

graphite-webgraph/src/main/kotlin/io/johnsonlee/graphite/webgraph/MappedWebGraphBackedGraph.kt
johnsonlee: Rechecked at head `279bdde`: this commit does not touch mapped admission. The mixed sequence remains **25.489 ± 2.414 ms / 44,991,780 B/op** versus `main` at **13.008 ± 1.994 ms / 30,467,932 B/op** (1.96x slower, +47.7% allocation). The thread remains open.

graphite-core/src/main/kotlin/io/johnsonlee/graphite/graph/Graph.kt
johnsonlee: Still blocking at `279bdde`. I recompiled a delegating `LegacyGraph` implementation against the `main` core JAR, then ran it with this head's core/cypher JARs. The optimized query deterministically throws `AbstractMethodError` at `QueryPipeline.kt:460`; the new broad-disjunction path also invokes the same abstract slot at line 516. `javap` still shows the method as abstract with its body only in `Graph$DefaultImpls`, and there is still no legacy-artifact compatibility test.

johnsonlee: Verified fixed at `0344e03`. I compiled a delegating `LegacyGraph` against the current `main` core JAR, then ran both the direct string-filter path and broad-disjunction path with this head's core/cypher JARs; both completed with `BINARY_COMPATIBILITY_PASS` and no `AbstractMethodError`. `javap` also confirms `Graph` no longer owns an abstract `nodesByStringProperty` slot; the optional capability does. Resolving this thread.

graphite-webgraph/src/main/kotlin/io/johnsonlee/graphite/webgraph/MappedWebGraphBackedGraph.kt
johnsonlee: Verified fixed at `0344e03`. Eviction now clears every admission for the evicted property under the same lock, and an independent access-order test confirms the next early lookup raw-scans instead of rebuilding. The new JMH control `earlyHitAfterLruEviction` measures `0.231 ± 0.181 ms/op`. Resolving this thread.

graphite-webgraph/src/main/kotlin/io/johnsonlee/graphite/webgraph/MappedWebGraphBackedGraph.kt
johnsonlee: `0344e03` fixes the different-predicate reproduction (`earlyHitAfterAdmittedMiss`: `0.468 ± 0.296 ms/op`), but the admission policy is still limit-insensitive for the same predicate. On a 100k mapped fixture, after `CONTAINS 'path-' LIMIT 100000`, the identical predicate with `LIMIT 1` measures **21.339 ± 0.703 ms/op** here versus **0.080 ± 0.069 ms/op** on `main` (about **267x slower**; identical JMH settings, setup excluded from timing). A 300-node regression also proves the second call changes index count from 0 to 1. `StringPropertyAdmissionKey` contains property/mode/expected but not the scan budget, and the added test changes both mode and expected before its early hit, so it misses this case. A large-limit page followed by a latency-sensitive small-limit page can still pay the original full-build spike. Keeping this blocking thread open.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
johnsonlee: Rechecked at `0344e03`: this path is unchanged and remains blocking. The 1,000-match consumption audit still consumes all 1,000 nodes for `DISTINCT ... LIMIT 1`; a second lazy-sequence sentinel test fails as soon as the implementation reads candidate 2 after candidate 1 already satisfies the limit. `candidates.forEach` plus `unique.values.sortedBy` still completes before the first `yield`, while the caller applies the limit afterward. Please add a consumption-bound regression and make the union incrementally emit candidates.

johnsonlee (commented):
Changes are still requested on `0344e03`. The Graph ABI and LRU-admission blockers are independently verified fixed and their threads are resolved. Clean `./gradlew check koverXmlReport koverPrintCoverage --no-daemon` and GitHub CI pass, and the original different-predicate admission benchmark is back below 0.5 ms/op. Two blocking inline findings remain: same-predicate high-LIMIT → `LIMIT 1` still forces a complete mapped index build, and the broad string-disjunction path still consumes/retains/sorts every match before returning row 1. Formal `REQUEST_CHANGES` is unavailable because the authenticated account is the PR author, so this review records the blocking state as a comment.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
```diff
@@ -277,21 +379,261 @@ class QueryPipeline private constructor(
...
+                    filter.property,
+                    filter.mode,
+                    filter.expected,
+                    completeScanLimit
+                )
+            }
+            val candidates = if (accelerated.any { it == null }) {
+                graph.nodes(candidateType).filter(disjunction::matches)
+            } else {
+                mergeSortedNodes(accelerated.filterNotNull())
```
johnsonlee: **Blocking — the real mapped lookup is not sorted, so this merge can emit the same node more than once.** `buildStringPropertyIndex` copies `nodeTypeIndex.ids(type)` verbatim, and persisted type IDs preserve `DefaultGraph`/fastutil hash iteration order. A real save/loadMapped contract test returned IDs `[1,30,70,2,90,4,5,50,40,3]`, not ascending order; the new unit test only supplies hand-sorted sequences. With interleaved property streams, the merge can therefore emit `[4,6,4,...]`. I reproduced a mapped `RETURN DISTINCT n.id, rand() LIMIT 3` result with IDs `[4,6,4]`, evaluating one MATCH node twice and displacing another row; the previous eager union deduplicated by ID regardless of input order. Please make the provider/merge safe for unsorted IDs and add a real mapped integration test. Sorting only newly written type indexes is insufficient because existing persisted graph directories already contain hash-order IDs.

johnsonlee (commented):
Changes are still requested on `2d54097`. The lazy merge fixes full materialization for sorted providers, but it introduces a new correctness blocker because the built-in mapped provider (including existing persisted graphs) does not satisfy the new ordering precondition. The same-predicate high-LIMIT → `LIMIT 1` admission regression is also untouched and remains about 285x slower than `main`. Targeted tests pass and GitHub CI is still running; blocking details are inline.

graphite-webgraph/src/main/kotlin/io/johnsonlee/graphite/webgraph/MappedWebGraphBackedGraph.kt
johnsonlee: Still blocking at `2d54097`; this commit does not change mapped admission. The identical 100k-node high-LIMIT → same-predicate `LIMIT 1` JMH now measures **22.773 ± 0.881 ms/op**, versus **0.080 ± 0.069 ms/op** on `main` (about **285x slower**). The 300-node index-count regression also still fails with `expected 0 but was 1`.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
johnsonlee: The original eager materialization is fixed in `2d54097`: with two sorted property streams over 1,000 matching nodes, `LIMIT 1` now consumes exactly the two initial heads, and the cold first-type control improves from `1.776 ms / 3.403 MB` to `0.356 ms / 0.857 MB`. Resolving this original thread. The new mapped-provider ordering/correctness blocker introduced by the merge is tracked separately at https://github.com/johnsonlee/graphite/pull/91#discussion_r3865330892.

graphite-webgraph/src/main/kotlin/io/johnsonlee/graphite/webgraph/MappedWebGraphBackedGraph.kt
johnsonlee: Fixed in 8ad1b54. Finite admission is now keyed by node type, property, match mode, expected string, and lookup limit, so a large-page scan cannot admit a later LIMIT 1 request. The new 300-node regression proves a same-predicate limit=300 scan leaves the limit=1 path on raw scan (index count remains 0), while repeating limit=300 builds exactly one index. The new JMH control earlyHitAfterLargeLimitScan uses an identical 50k ResourceFile + 50k Field fixture on main and this branch: main 0.484 +/- 0.172 ms/op, branch 0.465 +/- 0.226 ms/op; confidence intervals overlap. The earlier 21.339 ms build spike is gone. Admission remains bounded to 32 entries / 64 KiB.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
johnsonlee: Fixed in 2d54097. Property lookup sequences now declare ascending node-ID order and are combined with a lazy k-way merge that retains one head per stream, deduplicates equal IDs, and yields before advancing. The 1,000-node counting regression now consumes exactly 2 candidates for LIMIT 1, one head from each of two property streams, rather than all 1,000. A second regression exhausts interleaved streams [0,2] and [0,1,2] and asserts ordered deduplicated output [0,1,2]. On the Android 5.9M-node fixture at the latest head, cold/repeated broad discovery are 217.354/231.419 ms/op versus main at 7,426.015/7,433.167 ms/op (34.17x/32.12x).

graphite-webgraph/src/jmh/kotlin/io/johnsonlee/graphite/webgraph/MappedStringAdmissionBenchmark.kt
```diff
@@ -0,0 +1,147 @@
...
+        check(indexCount(ResourceFileNode::class.java, "path") == 0)
+    }
+}
+
+@State(Scope.Thread)
+open class LargeLimitBenchmarkState : MappedStringAdmissionBenchmarkState() {
+    @Setup(Level.Invocation)
+    fun admitLargeLimit() {
+        clearIndexes()
+        check(execute(LARGE_LIMIT_PATH_QUERY).rows.size == ADMISSION_RESOURCE_NODES)
```
johnsonlee: This benchmark does not exercise the regression it claims to cover. Setup runs `CONTAINS 'path-' LIMIT 50000`, but `earlyHit()` measures `CONTAINS 'path-0' LIMIT 1`; because the admission key already includes `expected`, this control would pass even if `limit` were removed from the key. Please use the identical predicate on both sides with only the limit changed, and assert setup/measurement leave the index unbuilt as appropriate. An independent true same-predicate JMH confirms the implementation is currently fixed (`0.282 ± 0.131 ms/op`), but the checked-in benchmark and PR evidence are presently a false-negative.

johnsonlee (commented):
Changes are still requested on `8ad1b54`. The limit-aware admission implementation is independently verified fixed (`0.282 ± 0.131 ms/op` versus `22.773 ± 0.881` on the prior head), so that blocker can close. The mapped-ordering correctness blocker is untouched and remains reproducible on both raw and indexed paths. The checked-in JMH control also does not actually exercise the same-predicate/different-limit case described in the docs; details are inline.

graphite-webgraph/src/main/kotlin/io/johnsonlee/graphite/webgraph/MappedWebGraphBackedGraph.kt
johnsonlee: Verified fixed at `8ad1b54`. A true same-predicate 100k-node JMH (`CONTAINS 'path-' LIMIT 100000` followed by the identical predicate at `LIMIT 1`) now measures **0.282 ± 0.131 ms/op**, down from **22.773 ± 0.881 ms/op** on `2d54097` and back to the same sub-millisecond class as `main` (`0.080 ± 0.069`). Independent index-count assertions remain 0 after both the large-limit scan and the low-limit lookup. Resolving this implementation blocker; a separate inline notes that the newly checked-in JMH accidentally changes the expected string and should be corrected.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
johnsonlee: Fixed in d2cd0ee. I removed the ascending-order contract and k-way merge. Candidate sequences are now consumed lazily in storage order and deduplicated with an IntOpenHashSet, so old hash-ordered mapped type indexes are supported without collecting Node objects or sorting the complete match set. LIMIT 1 now consumes one candidate. The new real save/loadMapped regression first verifies that ID 90 precedes ID 4 in the persisted CallSite type order, then runs RETURN DISTINCT n.id, rand() over interleaved caller/callee streams and asserts IDs [4, 90] exactly once each. An unordered unit stream test covers the same contract. Latest Android cold/repeated results are 202.359/199.474 ms/op versus main 7,426.015/7,433.167 ms/op (36.70x/37.26x).

graphite-webgraph/src/jmh/kotlin/io/johnsonlee/graphite/webgraph/MappedStringAdmissionBenchmark.kt
johnsonlee: Fixed in d2cd0ee. Setup and measurement now both use CONTAINS path-0; only the limit changes from 50000 to 1. Setup asserts the large-limit query returns the single path-0 row and leaves index count at 0. The measured method also asserts index count remains 0 after the LIMIT 1 query. Using the identical corrected fixture/harness, main measures 0.518 +/- 0.392 ms/op and this branch 0.463 +/- 0.356 ms/op. The existing 300-node unit regression independently verifies that the large-limit admission does not transfer to LIMIT 1 and that repeating the original large limit builds exactly one index.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
```diff
@@ -277,21 +380,250 @@ class QueryPipeline private constructor(
...
+                graph.nodes(candidateType).filter(disjunction::matches)
+            } else {
+                uniqueNodes(accelerated.filterNotNull())
+            }
+            for (node in candidates) yield(node)
+        }
+    }
+
+    private fun <T : Node> uniqueNodes(sequences: List<Sequence<T>>): Sequence<Node> = sequence {
+        val seen = IntOpenHashSet()
```
johnsonlee: **Blocking performance/memory — this set is bounded by all matches, not by `LIMIT`, for cross-graph queries.** `executeDirectStringDisjunction` only stops early when `!qualified`; qualified execution drains every source to merge complete provenance. Thus a cross-graph `DISTINCT ... LIMIT 1` with a broad direct OR retains every unique matching ID for the current source/type here. At 5.9M IDs, the fastutil table alone needs roughly a 32 MiB backing array before overhead, and larger corpora continue growing, reintroducing the saturation/OOM shape this PR is meant to remove. The lazy consumption test uses a single graph and cannot catch this. Please deduplicate unordered streams with memory bounded by the filter count (for example, let the first matching filter own a node and suppress it in later streams by checking earlier predicates), and add a qualified/cross-graph large-match consumption or retained-memory regression.

johnsonlee (commented):
Changes are still requested on `d2cd0ee`. The unordered mapped correctness fix and corrected same-predicate JMH are independently verified, including raw/indexed/old-fixture tests, full checks, and performance controls. However, the replacement ID-set deduplication introduces a new cross-graph memory blocker inline: qualified execution does not stop at LIMIT, so the set can retain every matching node ID. GitHub CI is still running.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
johnsonlee: Fixed in eefda6d. The primitive ID set is gone. Candidate stream i now emits a node only when none of filters 0..<i matches it, so deduplication is independent of match count and does not assume storage order. Added a qualified two-graph regression with 5,000 nodes per graph: it drains all 20,000 indexed candidates, returns one DISTINCT row, and merges both graph IDs. The existing single-graph LIMIT 1 guard still consumes one candidate, and the unordered-stream fixture now obeys the lookup contract. Final Android-JAR JMH is 207.392 ms/op cold and 204.300 ms/op repeated versus main at 7,426.015/7,433.167 ms/op; the 16-graph repeated result is 2.861 ms/op.

johnsonlee (commented):
Follow-up review complete on `eefda6d`: all blocking inline comments are independently verified fixed, all review threads are resolved, and no new blocker was found. Verification includes raw and indexed unordered mapped streams, old persisted graph compatibility, cross-graph provenance and non-deterministic projection deduplication, ABI compatibility, limit-aware admission, LRU eviction, memory budgets, and the corrected JMH controls. A clean no-cache `./gradlew check koverXmlReport koverPrintCoverage --rerun-tasks --no-daemon` ran 2,015 tests with zero failures; every module remains above 98% line coverage, and GitHub CI is green. Performance controls show no regression: same-predicate high-LIMIT → LIMIT 1 is at main parity; single-graph streaming consumes one candidate; the final filter-owned cross-graph implementation removes match-sized retained state and improves both all-overlap and four-filter disjoint controls versus the prior head. No changes remain requested. Formal approval is unavailable because the authenticated account is the PR author.

PR #92, opened 2026-08-26
Add real-world large corpus performance gates

johnsonlee:
## Summary

- replace the sub-million-node Elasticsearch fixture with Apache Tika, Apache Hive, and the Kotlin compiler
- apply the Android protocols to all three corpora: 8 GiB method/load/query JMH and a 4 GiB `JAR -> build -> save -> mapped load -> Cypher` gate
- make the gate execute on every invocation, outside Kover, in isolated workers with a four-minute test timeout
- verify exact source and mapped node/method/edge/call-site shape plus deterministic property and relationship query results
- separate Android and large-corpus fixture configurations, honor explicit JMH path overrides, and isolate eager and mapped benchmark states

## Main comparison

The branch is rebased onto `origin/main@1d910a8`. This PR changes fixture wiring, JMH/test sources, and documentation, but no production source. The new named benchmarks do not exist on main, so these runs measure the production implementation from that exact main commit through the new harness and establish its initial baseline.

- Method-level conclusion: no regression; there is no production-code delta from main.
- End-to-end conclusion: no regression; there is no production-code delta from main.

## Environment

- Apple M3 Max, 64 GiB RAM, macOS 14.3 (`arm64`)
- OpenJDK 17.0.18 (Homebrew)
- JMH 1.37

## Automated 4 GiB gate

Commands:

```bash
./gradlew :webgraph:largeCorpusTest -Dlarge.corpus.record=true --no-daemon
./gradlew clean :webgraph:largeCorpusTest --no-daemon
```

| Corpus | Nodes | Source edges | Persisted edges | Methods | Call sites | Pipeline | Peak heap | Time / heap ceiling |
| --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| Tika | 3,897,012 | 4,497,723 | 4,342,382 | 312,788 | 1,002,088 | 18.104 s | 3,399,385,088 B | 120 s / 4 GiB hard cap |
| Hive | 5,986,673 | 6,378,063 | 6,161,463 | 404,016 | 1,437,647 | 30.506 s | 3,993,927,680 B | 180 s / 4 GiB hard cap |
| Kotlin compiler | 3,268,537 | 3,674,711 | 3,559,500 | 249,669 | 900,366 | 16.421 s | 3,088,251,984 B | 120 s / 4 GiB hard cap |

`GraphStore` intentionally retains the last edge for duplicate `(from, to)` pairs, so the gate records both logical source edges and the exact persistable edge count. Pipeline time is the sum of separately timed build, save, mapped-load, and query phases; save follows build immediately and validation scans run afterward, outside those timers. The clean run executed `largeCorpusTest`; it was neither `UP-TO-DATE` nor restored `FROM-CACHE`. Sampled Java heap is diagnostic rather than a portable assertion because it varies with GC and runner behavior; the isolated worker's `-Xmx4g`/OOM behavior is the hard memory gate. Explicit persisted-graph overrides are rejected in a Gradle preflight unless their node count matches the named corpus, so invalid overrides fail before any JMH fork.

## Method-level JMH

```bash
./gradlew :sootup:jmh -Pjmh.filter='GraphBuildBenchmark.build(Tika|Hive|KotlinCompiler)GraphEndToEndConfig$' --no-daemon
```

| Exact benchmark | Main-equivalent baseline |
| --- | ---: |
| `GraphBuildBenchmark.buildTikaGraphEndToEndConfig` | 11,878.662 ms/op |
| `GraphBuildBenchmark.buildHiveGraphEndToEndConfig` | 21,389.999 ms/op |
| `GraphBuildBenchmark.buildKotlinCompilerGraphEndToEndConfig` | 11,150.302 ms/op |

## End-to-end JMH

```bash
./gradlew :webgraph:jmh -Pjmh.filter='GraphEndToEndBenchmark.(tika|hive|kotlinCompiler)_build_save_load_query$' --no-daemon
```

| Exact benchmark | Main-equivalent baseline |
| --- | ---: |
| `GraphEndToEndBenchmark.tika_build_save_load_query` | 15,445.314 ms/op |
| `GraphEndToEndBenchmark.hive_build_save_load_query` | 27,400.797 ms/op |
| `GraphEndToEndBenchmark.kotlinCompiler_build_save_load_query` | 17,673.737 ms/op |

## Persisted load JMH

```bash
./gradlew :webgraph:jmh -Pjmh.filter='LargeCorpusLoadBenchmark' --no-daemon
```

| Exact benchmark | Corpus | Main-equivalent baseline |
| --- | --- | ---: |
| `LargeCorpusLoadBenchmark.eager_load` | Tika | 3,394.974 ms/op |
| `LargeCorpusLoadBenchmark.eager_load` | Hive | 5,505.411 ms/op |
| `LargeCorpusLoadBenchmark.eager_load` | Kotlin compiler | 3,258.401 ms/op |
| `LargeCorpusLoadBenchmark.mapped_load` | Tika | 86.478 ms/op |
| `LargeCorpusLoadBenchmark.mapped_load` | Hive | 129.515 ms/op |
| `LargeCorpusLoadBenchmark.mapped_load` | Kotlin compiler | 72.894 ms/op |

## Mapped query JMH

```bash
./gradlew :webgraph:jmh -Pjmh.filter='LargeCorpusQueryBenchmark.mapped_.*' --no-daemon
```

| Exact benchmark | Tika (ms/op) | Hive (ms/op) | Kotlin compiler (ms/op) |
| --- | ---: | ---: | ---: |
| `LargeCorpusQueryBenchmark.mapped_countStar` | 0.003 | 0.002 | 0.002 |
| `LargeCorpusQueryBenchmark.mapped_intConstantFilter` | 0.047 | 0.049 | 0.056 |
| `LargeCorpusQueryBenchmark.mapped_returnDistinct` | 0.185 | 0.110 | 0.114 |
| `LargeCorpusQueryBenchmark.mapped_simpleNodeMatch` | 0.073 | 0.088 | 0.064 |
| `LargeCorpusQueryBenchmark.mapped_singleHopRelationship` | 0.041 | 0.696 | 0.298 |

## Verification

```bash
git diff --check
./gradlew check koverLog --no-daemon
```

Both passed locally. The repository-wide Gradle run completed successfully with 83 actionable tasks, including Detekt, Kover verification, regular tests, and a fresh execution of all three large-corpus gates. Full fixture fingerprints, phase timings, persisted sizes, limitations, and reproduction commands are in `docs/large-corpus-performance-baseline.md`.


graphite-webgraph/build.gradle.kts
```diff
@@ -70,14 +75,48 @@ jmh {
...
             when {
-                jar.name.contains("elasticsearch") -> systemProperty("elasticsearch.jar.path", jar.absolutePath)
-                jar.name.contains("android-all") -> systemProperty("android.jar.path", jar.absolutePath)
+                jar.name.startsWith("android-all-") -> systemProperty("android.jar.path", jar.absolutePath)
             }
         }
     }
 }
+
+val largeCorpusTest by tasks.registering(Test::class) {
```
johnsonlee, line 126: Blocking: this performance gate is cacheable/up-to-date like an ordinary `Test` task. After one real run, the same command returned `UP-TO-DATE` in 4s; after `clean`, it returned `FROM-CACHE` in 5s, with no test JVM or measurements. The fixture JARs are also only resolved in `doFirst` and are not task inputs, so a cached result can bypass the in-test SHA checks. A timing/heap/current-fixture gate must execute on every invocation. Please disable both up-to-date and build-cache reuse for `largeCorpusTest` (and declare fixture files as inputs if any caching remains).

graphite-webgraph/src/test/kotlin/io/johnsonlee/graphite/webgraph/LargeCorpusPerformanceGateTest.kt
```diff
@@ -0,0 +1,252 @@
...
+            val saveMillis = elapsedMillis(saveStart)
+            closeQuietly(sourceGraph)
+            sourceGraph = null
+
+            val loadStart = System.nanoTime()
+            loadedGraph = GraphStore.loadMapped(output)
+            val mappedLoadMillis = elapsedMillis(loadStart)
+
+            val queryStart = System.nanoTime()
+            val result = loadedGraph.query("MATCH (n:CallSiteNode) RETURN count(*) AS count")
```
johnsonlee: Blocking: this query proves only the `CallSiteNode` count. It takes the node-count fast path and the mapped backend reads only the type-index count, so it never deserializes node properties or touches edges, labels, or method metadata. A persistence regression that drops/corrupts all edges or method data can therefore pass—and may look faster. Please compare source vs mapped method/edge counts and run at least one deterministic property query plus one relationship query before claiming an exact-shape round trip.

graphite-webgraph/src/jmh/kotlin/io/johnsonlee/graphite/webgraph/GraphBenchmark.kt
```diff
@@ -171,94 +130,105 @@ open class EsQueryBenchmark {
...
+    lateinit var corpus: String
 
     private lateinit var eagerGraph: Graph
     private lateinit var mappedGraph: Graph
 
     @Setup
     fun setup() {
-        val graphPath = BenchmarkCorpus.persistedGraph(BenchmarkCorpusKind.ANDROID)
+        val graphPath = BenchmarkCorpus.persistedGraph(BenchmarkCorpusKind.valueOf(corpus))
         eagerGraph = GraphStore.load(graphPath)
```
johnsonlee: `GraphStore.load(graphPath)` defaults to `LoadMode.AUTO`; all three corpora exceed the 1M-node cutoff, so this resolves to `loadMapped`. Consequently every `eager_*` benchmark actually measures mapped mode, while each `mapped_*` run retains a second mapped graph in the same fork. Please use `GraphStore.load(graphPath, GraphStore.LoadMode.EAGER)` and split eager/mapped states so filtered mapped-only runs do not retain an unrelated graph.

graphite-webgraph/build.gradle.kts
```diff
@@ -70,14 +75,48 @@ jmh {
 
 tasks.test {
     maxHeapSize = "4g"
-    // Large Android/ES fixtures must prove the 4G gate without inheriting heap from prior tests.
+    // The Android fixture must prove the 4 GiB gate without inheriting heap from prior tests.
     forkEvery = 1
+    exclude("**/*CorpusPerformanceGateTest.class")
     doFirst {
         integrationFixtures.resolve().forEach { jar ->
```
johnsonlee: The normal `:webgraph:test` task excludes the new corpus gates but still resolves the single `integrationFixtures` configuration, which now contains Tika, Hive, and the Kotlin compiler as well as Android. A first regular test run therefore downloads/resolves all three large new JARs even though it uses only Android. Please split Android and large-corpus fixture configurations so the default unit-test path remains isolated.

docs/webgraph-storage.md
```diff
@@ -142,9 +142,11 @@ Use both micro and end-to-end benchmarks. A change is not accepted based on synt
...
 | `GraphBuildPersistBenchmark` | Synthetic 10M save/load guardrail | `./gradlew :webgraph:jmh -Pjmh.filter=GraphBuildPersistBenchmark` |
 | `GraphEndToEndBenchmark` | Real JAR `build -> save -> load -> query` | `./gradlew :webgraph:jmh -Pjmh.filter=GraphEndToEndBenchmark` |
-| `GraphBenchmark` | Persisted-graph load/query comparisons | `./gradlew :webgraph:jmh -Pjmh.filter='(Es|Android).*(Load|Query)Benchmark'` |
+| `GraphBenchmark` | Persisted-graph load/query comparisons | `./gradlew :webgraph:jmh -Pjmh.filter='(Android|LargeCorpus).*(Load|Query)Benchmark'` |
 
-`GraphEndToEndBenchmark` and `GraphBenchmark` auto-discover fixture JARs from Gradle cache, or accept explicit overrides via `-Delasticsearch.jar.path`, `-Dandroid.jar.path`, `-Delasticsearch.graph.path`, and `-Dandroid.graph.path`.
+`GraphEndToEndBenchmark` and the load/query benchmarks auto-discover fixture JARs from the Gradle cache. Explicit overrides use `-Dandroid.jar.path`, `-Dtika.jar.path`, `-Dhive.jar.path`, and `-Dkotlin.compiler.jar.path`; persisted graph overrides use the corresponding `.graph.path` properties.
```
johnsonlee: These documented overrides do not work with the shown Gradle JMH commands. The build unconditionally appends Gradle-cache JAR paths to the fork, replacing an explicit `-D*.jar.path`, and it does not forward any `-D*.graph.path` property at all. I reproduced this with `-Dtika.jar.path=/tmp/definitely-wrong`: the fork still received the cached Tika path. Please either forward caller-provided properties (with explicit values taking precedence) or document the direct `java -D... -jar ...` invocation that actually supports them.

johnsonlee (commented):
Changes requested. GitHub does not allow the PR author to submit a formal REQUEST_CHANGES review on their own PR, so this review is submitted as COMMENT; the two comments marked Blocking must be resolved before merge. I independently forced all three corpus gates (pass), built/listed the JMH jars, ran the regular WebGraph tests and detekt, and reproduced both UP-TO-DATE and FROM-CACHE skips.

docs/large-corpus-performance-baseline.md
```diff
@@ -0,0 +1,121 @@
...
+
+## Reproducibility
+
+Recorded on 2026-08-27 with:
+
+- Apple M3 Max, 64 GiB RAM, macOS 14.3 (`arm64`)
+- OpenJDK 17.0.18 (Homebrew)
+- JMH 1.37
+- production sources identical to `origin/main` at `e4d1c6a`; this change adds only fixtures,
+  benchmarks, tests, and documentation, so the measurements establish the main implementation's
```
johnsonlee, line 16: Blocking: this baseline is already stale relative to the branch it will enter. Current `main` is `1d910a8` (#91), which directly changes `QueryPipeline`, `Graph`, and `MappedWebGraphBackedGraph`; the green PR run and GitHub merge ref still merge this head into `e4d1c6a`. I manually verified that the three count/time gates still pass on a current-main merge, but the committed JMH load/query/end-to-end numbers were not measured on that production code. Since this document is intended to be the initial comparison point, please update/rebase and rerun those benchmarks on current main/merge before committing the baseline.

johnsonlee (commented):
Second-pass changes requested. I also constructed the prospective merge against current `main@1d910a8` and forced all three corpus gates with `--rerun-tasks --no-build-cache`; they pass in 1m42s. The comments below cover evidence the existing CI/gate still does not establish. As before, GitHub prevents the PR author from submitting a formal REQUEST_CHANGES state on their own PR.

graphite-webgraph/build.gradle.kts
```diff
@@ -70,14 +75,48 @@ jmh {
...
     }
 }
+
+val largeCorpusTest by tasks.registering(Test::class) {
+    description = "Runs Tika, Hive, and Kotlin compiler gates in isolated 4 GiB JVMs"
+    group = "verification"
+    testClassesDirs = sourceSets.test.get().output.classesDirs
+    classpath = sourceSets.test.get().runtimeClasspath
+    include("**/*CorpusPerformanceGateTest.class")
+    maxHeapSize = "4g"
```
johnsonlee, line 132: Blocking: this 4 GiB/time gate is running coverage-instrumented code, not the production pipeline it claims to measure. A forced `--info` run shows the test worker command includes `-javaagent:.../kover-jvm-agent-0.9.1.jar ... -Xmx4g`. That overhead can create false OOM/time failures; Hive already sampled about 3.9 GB in repeated runs. Please exclude this performance task from Kover instrumentation or launch the measured pipeline in a clean JVM, and record the baseline in that same configuration.

graphite-webgraph/src/test/kotlin/io/johnsonlee/graphite/webgraph/LargeCorpusPerformanceGateTest.kt
```diff
@@ -0,0 +1,252 @@
...
+
+        val measurement = PeakHeapSampler().use { sampler -> runPipeline(jar, sampler) }
+        println(measurement.baselineLine(baseline))
+
+        if (!java.lang.Boolean.getBoolean(RECORD_PROPERTY)) {
+            assertEquals(baseline.nodeCount, measurement.nodes, "Node baseline changed for ${baseline.id}")
+            assertEquals(baseline.methodCount, measurement.methods, "Method baseline changed for ${baseline.id}")
+            assertEquals(baseline.callSiteCount, measurement.callSites, "Call-site baseline changed for ${baseline.id}")
+            assertTrue(
+                measurement.pipelineMillis <= baseline.maxPipelineMillis,
```
johnsonlee, line 128: This is an elapsed-time assertion, not a timeout: if build/save/load/query hangs, control never reaches it. Neither the test/task nor the workflow has a timeout, so a 120/180-second ceiling can consume the GitHub job default (up to hours) instead of failing near the stated limit. Please add an execution timeout at the test/process or CI-job boundary; keep this assertion as the completed-run performance check.

graphite-webgraph/src/test/kotlin/io/johnsonlee/graphite/webgraph/LargeCorpusPerformanceGateTest.kt
```diff
@@ -0,0 +1,252 @@
...
+                "${baseline.id} pipeline took ${measurement.pipelineMillis} ms; ceiling=${baseline.maxPipelineMillis} ms"
+            )
+            assertTrue(
+                measurement.peakHeapBytes <= FOUR_GIB_BYTES,
```
johnsonlee: This assertion cannot independently fail. Line 100 already requires `maxMemory <= 4 GiB`, while the sampler records `totalMemory - freeMemory`; by JVM invariants, `used <= total <= maxMemory <= 4 GiB`. A run exceeding the limit OOMs before reaching this check. If sampled heap is meant to detect loss of headroom, give each corpus a meaningful ceiling below `-Xmx4g` (and define the metric's limitations); otherwise remove the assertion and describe `-Xmx4g`/OOM as the actual memory gate.

graphite-webgraph/build.gradle.kts
johnsonlee: Additional wiring consequence: `./gradlew :webgraph:koverLog --dry-run` includes `:webgraph:largeCorpusTest`, and the task currently produces a `largeCorpusTest.ic` coverage report. The workflow invokes `check` and then `koverLog` in a separate Gradle run; once the cache/up-to-date blocker is fixed by making this gate always run, the coverage step will execute all three corpora a second time. Please detach this task from both Kover instrumentation and Kover report dependencies, rather than only adjusting the recorded thresholds.

graphite-webgraph/src/jmh/kotlin/io/johnsonlee/graphite/webgraph/GraphBenchmark.kt
johnsonlee: Resolved in `0566941`. Eager states now call `GraphStore.load(path, GraphStore.LoadMode.EAGER)` explicitly. Android and large-corpus eager/mapped benchmarks each have separate JMH state classes and teardown, so a mapped-only filter initializes only the mapped graph. The updated load and all 15 mapped-query benchmarks ran successfully.

graphite-webgraph/src/test/kotlin/io/johnsonlee/graphite/webgraph/LargeCorpusPerformanceGateTest.kt
johnsonlee: Resolved in `0566941`. The gate now scans source logical edges, computes the exact unique `(from, to)` set that `GraphStore` can persist under its documented last-edge-wins simple-graph semantics, and compares mapped node, persisted-edge, method, and call-site counts. It also compares ordered property-query rows and ordered `DISTINCT` relationship-query rows between source and mapped graphs, with non-empty assertions. The strict three-corpus run passed.

graphite-webgraph/build.gradle.kts
johnsonlee: Resolved in `0566941`. The build now has separate non-transitive `androidIntegrationFixture` and `largeCorpusFixtures` configurations. Regular `test` resolves only Android; `largeCorpusTest` resolves only Tika/Hive/Kotlin. JMH combines both because it exposes both benchmark families.

graphite-webgraph/build.gradle.kts
johnsonlee: Resolved in `0566941`. `largeCorpusTest` now uses `doNotTrackState(...)`, which disables both up-to-date checks and build-cache reuse. A clean verification (`./gradlew clean :webgraph:largeCorpusTest --no-daemon`) executed the gate itself while only eligible compilation tasks came `FROM-CACHE`; all three corpora ran and passed. Because the task is entirely untracked, no cacheable fixture input remains to declare.

docs/webgraph-storage.md
johnsonlee: Resolved in `0566941`. Both JMH builds now prefer an explicit `System.getProperty(...)` over the resolved fixture, and WebGraph forwards all four `.graph.path` properties. Negative verification showed the fork VM receiving `-Dtika.jar.path=/tmp/graphite-review-missing-tika.jar` and `-Dtika.graph.path=/tmp/graphite-review-explicit-graph`; setup then failed on that explicit missing JAR instead of falling back to the cache. The documentation now states precedence and forwarding.

docs/large-corpus-performance-baseline.md
johnsonlee: Resolved in `0566941`. The branch is rebased onto current `main@1d910a8`, and the gate plus method-level, end-to-end, eager/mapped load, and all 15 mapped-query measurements were rerun on that production source. The committed baseline and PR body now identify `1d910a8` and contain the refreshed results.

graphite-webgraph/src/test/kotlin/io/johnsonlee/graphite/webgraph/LargeCorpusPerformanceGateTest.kt
johnsonlee: Resolved in `0566941`. The gate uses JUnit 4 `@Test(timeout = 240_000)`, providing a hard per-corpus execution timeout around fixture verification and the full build/save/load/query pipeline. The existing 120/180-second assertions remain completed-run performance ceilings.

graphite-webgraph/build.gradle.kts
johnsonlee: Resolved in `0566941`. Kover now declares `currentProject.instrumentation.disabledForTestTasks.add("largeCorpusTest")`. An `--info` run confirmed the 4 GiB gate worker has no Kover JVM agent. Direct `./gradlew :webgraph:koverLog --dry-run` includes normal `:webgraph:test` but not `largeCorpusTest`, so the coverage-only invocation does not repeat the corpora. The repository-wide `check koverLog` run passed.

graphite-webgraph/build.gradle.kts
johnsonlee: Resolved in .  now uses , which disables both up-to-date checks and build-cache reuse. A clean verification (To honour the JVM settings for this build a single-use Daemon process will be forked. For more on this, please refer to https://docs.gradle.org/8.13/userguide/gradle_daemon.html#sec:disabling_the_daemon in the Gradle documentation.
Daemon will be stopped at the end of the build 

> Configure project :
[root project 'graphite'] Configuring nexus skipped: `NEXUS_URL` or `NEXUS_USERNAME` or `NEXUS_PASSWORD` not found

> Configure project :core
[project ':core'] Configuring nexus skipped: `NEXUS_URL` or `NEXUS_USERNAME` or `NEXUS_PASSWORD` not found

> Configure project :cypher
[project ':cypher'] Configuring nexus skipped: `NEXUS_URL` or `NEXUS_USERNAME` or `NEXUS_PASSWORD` not found

> Configure project :sootup
[project ':sootup'] Configuring nexus skipped: `NEXUS_URL` or `NEXUS_USERNAME` or `NEXUS_PASSWORD` not found

> Configure project :webgraph
[project ':webgraph'] Configuring nexus skipped: `NEXUS_URL` or `NEXUS_USERNAME` or `NEXUS_PASSWORD` not found

> Task :clean UP-TO-DATE
> Task :core:clean
> Task :cypher:clean
> Task :explore:clean
> Task :query:clean
> Task :sootup:clean
> Task :webgraph:clean
> Task :core:checkKotlinGradlePluginConfigurationErrors SKIPPED
> Task :core:compileKotlin FROM-CACHE
> Task :core:compileJava NO-SOURCE
> Task :core:processResources NO-SOURCE
> Task :core:classes UP-TO-DATE
> Task :core:jar
> Task :webgraph:checkKotlinGradlePluginConfigurationErrors SKIPPED
> Task :webgraph:compileKotlin FROM-CACHE
> Task :webgraph:compileJava NO-SOURCE
> Task :webgraph:processResources NO-SOURCE
> Task :webgraph:classes UP-TO-DATE
> Task :webgraph:jar
> Task :cypher:checkKotlinGradlePluginConfigurationErrors SKIPPED
> Task :cypher:generateGrammarSource FROM-CACHE
> Task :cypher:compileKotlin FROM-CACHE
> Task :cypher:compileJava FROM-CACHE
> Task :cypher:processResources NO-SOURCE
> Task :cypher:classes UP-TO-DATE
> Task :cypher:jar
> Task :sootup:checkKotlinGradlePluginConfigurationErrors SKIPPED
> Task :sootup:compileKotlin FROM-CACHE
> Task :sootup:compileJava NO-SOURCE
> Task :sootup:processResources NO-SOURCE
> Task :sootup:classes UP-TO-DATE
> Task :sootup:jar
> Task :webgraph:compileTestKotlin FROM-CACHE
> Task :webgraph:compileTestJava NO-SOURCE
> Task :webgraph:koverFindJar FROM-CACHE
> Task :webgraph:processTestResources NO-SOURCE
> Task :webgraph:testClasses UP-TO-DATE
> Task :webgraph:largeCorpusTest

HiveCorpusPerformanceGateTest > build save mapped load and query stay within the large corpus gate STANDARD_OUT
    LARGE_CORPUS_BASELINE	hive	nodes=5986673	sourceEdges=6378063	persistedEdges=6161463	methods=404016	callSites=1437647	persistedBytes=506335478	buildMs=23682	saveMs=5430	mappedLoadMs=204	queryMs=3184	pipelineMs=34705	peakHeapBytes=3870140600

HiveCorpusPerformanceGateTest > build save mapped load and query stay within the large corpus gate PASSED

KotlinCompilerCorpusPerformanceGateTest > build save mapped load and query stay within the large corpus gate STANDARD_OUT
    LARGE_CORPUS_BASELINE	kotlin-compiler	nodes=3268537	sourceEdges=3674711	persistedEdges=3559500	methods=249669	callSites=900366	persistedBytes=289113024	buildMs=13108	saveMs=7058	mappedLoadMs=151	queryMs=2170	pipelineMs=24289	peakHeapBytes=3030286336

KotlinCompilerCorpusPerformanceGateTest > build save mapped load and query stay within the large corpus gate PASSED

TikaCorpusPerformanceGateTest > build save mapped load and query stay within the large corpus gate STANDARD_OUT
    LARGE_CORPUS_BASELINE	tika	nodes=3897012	sourceEdges=4497723	persistedEdges=4342382	methods=312788	callSites=1002088	persistedBytes=328441107	buildMs=16454	saveMs=4700	mappedLoadMs=181	queryMs=2846	pipelineMs=26488	peakHeapBytes=3426648064

TikaCorpusPerformanceGateTest > build save mapped load and query stay within the large corpus gate PASSED

BUILD SUCCESSFUL in 1m 34s
20 actionable tasks: 11 executed, 8 from cache, 1 up-to-date) executed the gate itself while only eligible compilation tasks came ; all three corpora ran and passed. Because the task is entirely untracked, no cacheable fixture input remains to declare.

graphite-webgraph/build.gradle.kts
johnsonlee: The Kover/task isolation is now technically correct, but it exposes a new blocking CI regression: on `0566941`, `./gradlew :webgraph:koverLog --offline --no-daemon` reports `application line coverage: 96.8198%`, while `.github/workflows/build.yml` exits nonzero for every module below 98%. I reproduced that the report now depends only on the ordinary `:webgraph:test`, as intended, so the pending CI run will fail at the coverage threshold. Please restore >=98% with ordinary focused tests (or make a separately justified threshold change); do not reattach this 4 GiB performance gate to Kover just to raise coverage.

graphite-webgraph/src/test/kotlin/io/johnsonlee/graphite/webgraph/LargeCorpusPerformanceGateTest.kt
johnsonlee: The new ceiling is non-tautological, but it fails immediately on this heads own GitHub run [33028916167](https://github.com/johnsonlee/graphite/actions/runs/33028916167): Hive reported `peakHeapBytes=4192129080`, above the 3.875 GiB ceiling (`4160749568`), so `:webgraph:largeCorpusTest` failed and coverage never ran. The sampled region also includes `sourceEdgeCounts()`, which iterates ~6M nodes and allocates a fresh `mutableSetOf<Int>()` for every node; that assertion work is inside both `pipelineStart` and the heap sampler. Please remove/allocation-optimize and separately time validation work, then establish a CI-stable headroom threshold from repeated clean runs. The gate should measure the production build/save/load/query pipeline, not millions of short-lived verifier allocations.

graphite-webgraph/src/jmh/kotlin/io/johnsonlee/graphite/webgraph/BenchmarkCorpus.kt
```diff
@@ -18,17 +18,29 @@ internal enum class BenchmarkCorpusKind(
...
     ANDROID(
         id = "android",
         jarPathProperty = "android.jar.path",
         graphPathProperty = "android.graph.path",
         matcher = { it.startsWith("android-all-") && it.endsWith(".jar") }
+    ),
+    TIKA(
+        id = "tika",
+        jarPathProperty = "tika.jar.path",
+        graphPathProperty = "tika.graph.path",
```
johnsonlee, line 33: The forwarding fix exposes an identity gap: the override path is accepted after checking only three filenames, without proving that it belongs to `kind`. I supplied the same small graph to the Tika, Hive, and Kotlin `*.graph.path` properties; all three `LargeCorpusQueryBenchmark.mapped_countStar` parameters completed and emitted separately labeled results for the identical graph. That can silently create a mislabeled committed baseline. Please persist/validate a corpus marker derived from the fixture SHA and loader config, or at minimum validate the expected corpus node count before running an override.

graphite-webgraph/src/test/kotlin/io/johnsonlee/graphite/webgraph/LargeCorpusPerformanceGateTest.kt
johnsonlee: Resolved in `f586f00` using the review's portable alternative: the tautological sampled-heap assertion is removed. The 10 ms Java-used-heap sample remains diagnostic, and the baseline documents that it varies with GC/runner behavior, can miss brief peaks, and excludes native/mmap memory. The isolated worker's `-Xmx4g` cap and OOM failure are now explicitly the hard memory gate. This was refined after GitHub's runner sampled Hive near the cap despite a healthy completed pipeline; all three strict gates pass with the hard cap.

johnsonlee: `f586f00` correctly removes the flaky sampled-heap pass/fail threshold and documents `-Xmx4g`/OOM as the portable hard gate; the forced three-corpus run now passes. I am keeping this thread open because the other half of the follow-up remains: `sourceEdgeCounts()` and its per-node `mutableSetOf` allocations are still inside `pipelineStart`/sampling, so the published pipeline time is contaminated by verifier work. Please move that validation outside the production timer (or separately report/document validation-inclusive time).

graphite-webgraph/build.gradle.kts
johnsonlee: Confirmed by the new GitHub run [33029585207](https://github.com/johnsonlee/graphite/actions/runs/33029585207): Build and coverage generation succeeded, then `Enforce coverage threshold` failed exactly with `Module webgraph coverage is 96.8198%, below 98% threshold`. The four new JMH `*QueryState` helpers account for 24 uncovered lines because the current Kover filter only matches `*Benchmark*`; excluding/renaming those benchmark-only state classes raises the recomputed value to about 98.007%. This remains blocking.

graphite-webgraph/src/jmh/kotlin/io/johnsonlee/graphite/webgraph/BenchmarkCorpus.kt
johnsonlee: Resolved in `1b510e8`. Every corpus now carries its exact expected node count (including Android 5,938,826), and an explicit `.graph.path` is opened mapped and validated before the benchmark state accepts it. Negative verification passed: supplying the Tika graph as `hive.graph.path` is rejected with `hive.graph.path has 3897012 nodes; expected 5986673 for hive`. The override behavior is documented.

graphite-webgraph/build.gradle.kts
johnsonlee: Resolved in `1b510e8` (introduced in `f48a096`). The four JMH-only helpers are now named `*BenchmarkState`, so the existing `*Benchmark*` Kover filter excludes them without reattaching `largeCorpusTest`. Local `:webgraph:koverLog` reports 98.0072%, and GitHub run 33030266943 passed Build, coverage generation, and the 98% enforcement step. Direct Kover still depends only on ordinary `:webgraph:test`.

graphite-webgraph/src/test/kotlin/io/johnsonlee/graphite/webgraph/LargeCorpusPerformanceGateTest.kt
johnsonlee: Resolved in `1b510e8`. `pipelineMillis` is now the sum of the separately timed production build, save, mapped-load, and query phases, so fixture and source/mapped validation scans cannot contaminate it. The edge verifier also avoids allocating a set for zero/one-outgoing-edge nodes and allocates only when a source has multiple edges. The refreshed strict results are Tika 17.515 s, Hive 28.679 s, and Kotlin compiler 16.211 s; the document explicitly says sampled heap covers validation but is diagnostic only.

johnsonlee: `1b510e8` fixes the arithmetic, but the measurement sequence still runs the full verifier between build and save: lines 154–161 scan every source node/edge and execute two queries before `saveStart` at line 163. That makes the reported save phase a verifier-warmed save rather than the production `build -> save` path; splitting the clocks does not remove cache/JIT/state contamination. Please start and run `GraphStore.save` immediately after `.load(jar)`, then perform the source validation while the graph is still available. Reopening until both the reported value and phase ordering represent the production pipeline.

graphite-webgraph/src/jmh/kotlin/io/johnsonlee/graphite/webgraph/BenchmarkCorpus.kt
johnsonlee: Correction after inspecting the benchmark protocol: `LargeCorpusLoadBenchmark` has an explicit one-second JMH warmup, so measured iterations already follow repeated loads; this setup check does not turn a claimed cold/first-load measurement into a warm one. The node-count check also satisfies the minimum acceptance criterion in the root comment and rejects accidental cross-corpus reuse, as verified with a 2-node graph. No remaining blocker in this thread; resolving it.

johnsonlee: A second negative test exposes a fail-open path at the task level. Using a valid Tika persisted graph as `-Dhive.graph.path` with the documented Gradle JMH command logs `IllegalArgumentException: hive.graph.path has 3897012 nodes; expected 5986673 for hive` in both Hive load setups, but JMH continues with the other corpus parameters and Gradle finishes `BUILD SUCCESSFUL` with exit code 0; Hive is merely absent from the result table. This happens despite `failOnError.set(true)`, so automation that trusts the documented command's status can silently accept an incomplete baseline. Please make override validation a preflight task failure before JMH parameter execution (or otherwise make any invalid corpus produce a non-zero task exit), and add a regression check for the documented invocation.

johnsonlee: Resolved in `9f1e49a`. The `jmh` task now depends on `validateJmhGraphOverrides`, a Gradle `JavaExec` preflight that validates every configured graph before any benchmark fork. The documented command with a valid Tika graph supplied as `-Dhive.graph.path` now fails in 4 seconds with exit code 1 at `:webgraph:validateJmhGraphOverrides`; `:webgraph:jmh` never starts. A no-override control skips preflight and completes JMH successfully. Exact per-corpus node validation remains in the fork as defense in depth.

graphite-webgraph/src/test/kotlin/io/johnsonlee/graphite/webgraph/LargeCorpusPerformanceGateTest.kt
johnsonlee: Resolved in `9f1e49a`. `GraphStore.save(sourceGraph, output)` now starts immediately after `JavaProjectLoader.load(jar)` completes, before any node/edge/method scan or source query. `pipelineMillis` remains the sum of the four separately timed production phases. The reordered strict run passed all corpora with Tika 18.104 s, Hive 30.506 s, and Kotlin compiler 16.421 s; documentation and PR data were refreshed.

johnsonlee: `9f1e49a` addresses the phase-order issue: `GraphStore.save` now starts immediately after build, and the full three-corpus gate passes locally with `pipelineMillis` still equal to the four timed phases. Keeping this thread open only until the current head's CI and final independent audit complete.

graphite-webgraph/src/jmh/kotlin/io/johnsonlee/graphite/webgraph/BenchmarkCorpus.kt
johnsonlee: `9f1e49a` addresses the fail-open path: the documented Gradle JMH invocation now runs `validateJmhGraphOverrides` before any benchmark fork, and the same invalid override exits 1 with `BUILD FAILED` in local and independent reproduction. Keeping this thread open only until the current head's CI and final independent audit complete.

graphite-webgraph/src/test/kotlin/io/johnsonlee/graphite/webgraph/LargeCorpusPerformanceGateTest.kt
johnsonlee: Current-head CI is complete and green: run 33032111643 passed Build, all three corpus gates, coverage generation, and threshold enforcement. Final audit also confirms `HEAD=9f1e49a`, merge-base equals current `origin/main@1d910a8`, a clean worktree, and `CLEAN`/`MERGEABLE`. Resolving as requested.

graphite-webgraph/src/jmh/kotlin/io/johnsonlee/graphite/webgraph/BenchmarkCorpus.kt
johnsonlee: Current-head CI is complete and green: run 33032111643 passed every step. Final audit reconfirmed the invalid documented invocation exits 1 before JMH, the no-override control completes JMH, `HEAD=9f1e49a` is based on current `origin/main@1d910a8`, and the PR is `CLEAN`/`MERGEABLE`. Resolving as requested.

PR #97, opened 2026-08-28
Restore case-insensitive multi-graph Cypher search performance

johnsonlee:
## Summary

- restore wrapped case-insensitive Cypher discovery latency by routing supported `toLower(coalesce(...))` string predicates through transformed mapped-graph lookups instead of scanning every node
- preserve the pre-optimization traversal order, `DISTINCT`/`LIMIT` behavior, null semantics, and graph provenance
- add a required fixed pre-PR-95 latency gate covering synthetic 1/17-graph scaling and all four repository fixture JARs, with eight query shapes and target distributions
- fail closed on result-digest mismatches, missing benchmark variants, invalid results, and cleanup/resource errors

## Root cause

PR #95 exposed the production query shape in a regression test, but its budget checks were not the source of the 9–10x slowdown. The wrapped predicates were not eligible for the mapped string-property lookup, so every graph performed a full node scan serially. The new lookup path recognizes only the semantically safe wrapped forms and uses persisted node byte offsets as an allocation-free canonical order when merging property streams.

## Correctness

The four-corpus harness executes every timed query on the fixed pre-PR-95 executor, current `main`, and this PR. It hashes complete columns, ordered rows, values, sets/maps, and graph provenance. All eight digests are identical across revisions.

The fixture distribution preflight pins zero-hit, dense four-graph, first-only, middle-only, last-only, broadly distributed, first/last bimodal, and highly skewed targets. Query forms vary caller/callee, class/method, `CONTAINS`/`STARTS WITH`/`ENDS WITH`, mixed OR predicates, and `LIMIT 1/50/250`.

## Benchmark environment

- Apple M3 Max, 64 GiB RAM
- macOS 14.3 arm64
- OpenJDK 17.0.18, Gradle 8.13
- JMH 1.37, one fork, one unmeasured warmup, three SingleShot measurements, `-Xmx8g`, GC profiler
- four persisted mapped graphs: Android, Tika, Hive, Kotlin Compiler; 19,091,048 nodes total
- fixed baseline: `44b57562f2b3d0c88882a9002bdc488e05e5d7a7`; PR base: `e2a2b1a2bf56c7311b639667dfea27a67350fddb`

Command:

```bash
java -jar graphite-webgraph/build/libs/webgraph-1.0.0-SNAPSHOT-jmh.jar \
  'io.johnsonlee.graphite.webgraph.(WrappedDiscoveryLatencyBenchmark|AllFixtureWrappedDiscoveryLatencyBenchmark).*CaseInsensitiveDiscovery' \
  -jvmArgsAppend '-Dandroid.graph.path=... -Dtika.graph.path=... -Dhive.graph.path=... -Dkotlin.compiler.graph.path=...' \
  -foe true -prof gc -rf json -rff benchmark-results/latency.json
```

## Latency results

| Exact benchmark | Pre-PR-95 | `main` | PR |
|---|---:|---:|---:|
| `AllFixtureWrappedDiscoveryLatencyBenchmark.zeroHitBroadContainsCaseInsensitiveDiscovery` | 15.209 s | 14.042 s | 0.520 s |
| `AllFixtureWrappedDiscoveryLatencyBenchmark.denseDistributedMethodContainsCaseInsensitiveDiscovery` | 14.019 s | 12.355 s | 1.836 s |
| `AllFixtureWrappedDiscoveryLatencyBenchmark.earlyGraphClassPrefixCaseInsensitiveDiscovery` | 11.506 s | 13.024 s | 1.481 s |
| `AllFixtureWrappedDiscoveryLatencyBenchmark.middleGraphsClassPrefixCaseInsensitiveDiscovery` | 14.314 s | 16.148 s | 1.349 s |
| `AllFixtureWrappedDiscoveryLatencyBenchmark.lateGraphClassPrefixCaseInsensitiveDiscovery` | 12.695 s | 11.374 s | 1.438 s |
| `AllFixtureWrappedDiscoveryLatencyBenchmark.broadlyDistributedClassPrefixCaseInsensitiveDiscovery` | 13.927 s | 13.479 s | 1.715 s |
| `AllFixtureWrappedDiscoveryLatencyBenchmark.firstLastGraphBimodalClassPrefixCaseInsensitiveDiscovery` | 16.289 s | 18.046 s | 2.842 s |
| `AllFixtureWrappedDiscoveryLatencyBenchmark.skewedMixedClassMethodOperatorCaseInsensitiveDiscovery` | 15.977 s | 16.817 s | 1.811 s |
| **Four-real-graph arithmetic mean** | **14.242 s** | **14.411 s** | **1.624 s** |
| `WrappedDiscoveryLatencyBenchmark.coldWrappedCaseInsensitiveDiscovery[graphCount=1]` | 3.994 ms | 4.207 ms | 0.566 ms |
| `WrappedDiscoveryLatencyBenchmark.coldWrappedCaseInsensitiveDiscovery[graphCount=17]` | 68.841 ms | 83.809 ms | 7.917 ms |
| `WrappedDiscoveryLatencyBenchmark.wrappedCaseInsensitiveDiscovery[graphCount=1]` | 5.660 ms | 5.108 ms | 0.126 ms |
| `WrappedDiscoveryLatencyBenchmark.wrappedCaseInsensitiveDiscovery[graphCount=17]` | 97.534 ms | 71.127 ms | 0.216 ms |

The four-real-graph average improves by 8.77x versus pre-PR-95 and 8.87x versus `main`. Every one of the 12 required rows passes the fixed-baseline minimum 50% improvement and current-base 15% regression gate.

## Verification

- `./gradlew check --no-daemon` — passed, including Hive/Tika/Kotlin Compiler large-corpus build/save/mapped-load/query gates and Kover verification
- `./gradlew :core:test :cypher:test :cypher:detekt :webgraph:test :webgraph:detekt :webgraph:compileJmhKotlin --rerun-tasks --no-build-cache --no-daemon` — passed
- `node --test .github/scripts/benchmark-gate.test.mjs` — 24/24 passed
- `actionlint .github/workflows/benchmark.yml` — passed
- fixed/base/candidate all-fixture correctness digest comparison — passed, 8/8
- 12-row `compare-latency-baseline` — passed
- `git diff --check` — passed

Method-level conclusion: no `CypherBenchmark` regression exceeds the existing 15% gate.

End-to-end conclusion: no adjacent build/save/load/query regression was found. Android mapped load improved from 149.418 to 139.761 ms, mapped query changes ranged from -12.6% to +6.7%, and Android build/save/load/query changed from 25,294.562 to 24,900.711 ms (-1.6%). The required Tika, Hive, and Kotlin Compiler full-pipeline gates pass.

### DIFF
--- a/.github/scripts/benchmark-gate.mjs
+++ b/.github/scripts/benchmark-gate.mjs
@@ -22,6 +22,7 @@ export const LATENCY_EXPECTED_BENCHMARK_KEYS = [
     "io.johnsonlee.graphite.webgraph.AllFixtureWrappedDiscoveryLatencyBenchmark.skewedMixedClassMethodOperatorCaseInsensitiveDiscovery",
     "io.johnsonlee.graphite.webgraph.AllFixtureWrappedDiscoveryLatencyBenchmark.zeroHitBroadContainsCaseInsensitiveDiscovery"
 ];
+export const LATENCY_EXPECTED_SHARDS = ["synthetic", "real-a", "real-b", "real-c", "real-d"];
 const LARGE_CORPUS_METRICS = [
     { key: "buildMs", label: "build", threshold: 20, minimum: 500, unit: "ms" },
     { key: "saveMs", label: "save", threshold: 25, minimum: 250, unit: "ms" },
@@ -403,6 +404,44 @@ export function renderLatencyBaselineReport(comparison) {
     return `${lines.join("\n")}\n`;
 }
 
+export function combineLatencyShards(directory) {
+    const errors = [];
+    const rows = [];
+    const seenKeys = new Set();
+    for (const shard of LATENCY_EXPECTED_SHARDS) {
+        const statusFile = path.join(directory, `latency-status-${shard}.json`);
+        if (!fs.existsSync(statusFile)) {
+            errors.push(`${shard}: latency shard status is missing`);
+            continue;
+        }
+        const status = readJson(statusFile);
+        for (const error of status.errors ?? []) errors.push(`${shard}: ${error}`);
+        for (const row of status.rows ?? []) {
+            if (seenKeys.has(row.key)) {
+                errors.push(`${shard}: duplicate latency benchmark ${row.key}`);
+            } else {
+                seenKeys.add(row.key);
+                rows.push(row);
+            }
+        }
+        if (status.passed !== true) errors.push(`${shard}: latency shard failed`);
+    }
+
+    const expected = new Set(LATENCY_EXPECTED_BENCHMARK_KEYS);
+    for (const key of expected) {
+        if (!seenKeys.has(key)) errors.push(`combined latency results: missing expected benchmark ${key}`);
+    }
+    for (const key of seenKeys) {
+        if (!expected.has(key)) errors.push(`combined latency results: unexpected benchmark ${key}`);
+    }
+    rows.sort((left, right) => left.key.localeCompare(right.key));
+    return {
+        passed: errors.length === 0 && rows.every((row) => !row.blocked),
+        errors,
+        rows
+    };
+}
+
 export function parseLargeCorpusLog(contents) {
     const results = new Map();
     for (const line of contents.split(/\r?\n/)) {
@@ -612,7 +651,7 @@ function compareLatencyBaselineCommand(args) {
         readJson(requireArg(args, "candidate")),
         Number(args.threshold ?? 15),
         Number(args["minimum-speedup"] ?? 50),
-        LATENCY_EXPECTED_BENCHMARK_KEYS
+        args["allow-subset"] === true ? null : LATENCY_EXPECTED_BENCHMARK_KEYS
     );
     writeFile(requireArg(args, "report"), renderLatencyBaselineReport(comparison));
     writeJson(requireArg(args, "status"), comparison);
@@ -627,9 +666,16 @@ function confirmLatencyBaselineCommand(args) {
         readJson(requireArg(args, "candidate")),
         Number(args.threshold ?? 15),
         Number(args["minimum-speedup"] ?? 50),
-        LATENCY_EXPECTED_BENCHMARK_KEYS
+        args["allow-subset"] === true ? null : LATENCY_EXPECTED_BENCHMARK_KEYS
     );
     const comparison = confirmLatencyBaseline(initial, confirmation);
+    writeFile(requireArg(args, "report"), renderLatencyBaselineReport(comparison));
+    writeJson(requireArg(args, "status"), comparison);
+    if (!comparison.passed) process.exitCode = 1;
+}
+
+function combineLatencyShardsCommand(args) {
+    const comparison = combineLatencyShards(requireArg(args, "directory"));
     writeFile(requireArg(args, "report"), renderLatencyBaselineReport(comparison));
     writeJson(requireArg(args, "status"), comparison);
     if (!comparison.passed) process.exitCode = 1;
@@ -677,6 +723,7 @@ function main(argv) {
     if (command === "compare-jmh") compareJmhCommand(args);
     else if (command === "compare-latency-baseline") compareLatencyBaselineCommand(args);
     else if (command === "confirm-latency-baseline") confirmLatencyBaselineCommand(args);
+    else if (command === "combine-latency-shards") combineLatencyShardsCommand(args);
     else if (command === "confirm-jmh") confirmJmhCommand(args);
     else if (command === "compare-large-corpus") compareLargeCorpusCommand(args);
     else if (command === "confirm-large-corpus") confirmLargeCorpusCommand(args);
--- a/.github/scripts/benchmark-gate.mjs
+++ b/.github/scripts/benchmark-gate.mjs
@@ -231,9 +231,9 @@ export function confirmJmh(initial, confirmation) {
     };
 }
 
-export function renderJmhReport(comparison) {
+export function renderJmhReport(comparison, title = "Method-level JMH") {
     const lines = [
-        "### Method-level JMH",
+        `### ${title}`,
         "",
         "A row blocks only when it exceeds the 15% limit, the 99.9% confidence intervals do not overlap,",
         "and a reverse-order confirmation run fails the same benchmark.",
@@ -586,6 +586,11 @@ export function renderLargeCorpusReport(comparison) {
 export function aggregateReports(directory, metadata) {
     const components = [
         { name: "method-level", report: "method-report.md", status: "method-status.json" },
+        {
+            name: "budgeted-mapped-string",
+            report: "budgeted-string-report.md",
+            status: "budgeted-string-status.json"
+        },
         { name: "large-corpus", report: "large-corpus-report.md", status: "large-corpus-status.json" },
         { name: "wrapped-query-latency", report: "latency-report.md", status: "latency-status.json" }
     ];
@@ -629,7 +634,7 @@ function compareJmhCommand(args) {
         readJson(requireArg(args, "candidate")),
         Number(args.threshold ?? 15)
     );
-    writeFile(requireArg(args, "report"), renderJmhReport(comparison));
+    writeFile(requireArg(args, "report"), renderJmhReport(comparison, args.title));
     writeJson(requireArg(args, "status"), comparison);
     if (!comparison.passed) process.exitCode = 1;
 }
@@ -689,7 +694,7 @@ function confirmJmhCommand(args) {
         Number(args.threshold ?? 15)
     );
     const comparison = confirmJmh(initial, confirmation);
-    writeFile(requireArg(args, "report"), renderJmhReport(comparison));
+    writeFile(requireArg(args, "report"), renderJmhReport(comparison, args.title));
     writeJson(requireArg(args, "status"), comparison);
     if (!comparison.passed) process.exitCode = 1;
 }
--- a/.github/scripts/benchmark-gate.mjs
+++ b/.github/scripts/benchmark-gate.mjs
@@ -8,6 +8,20 @@ import { pathToFileURL } from "node:url";
 export const COMMENT_MARKER = "<!-- graphite-benchmark-regression-gate -->";
 
 const MIB = 1024 * 1024;
+export const LATENCY_EXPECTED_BENCHMARK_KEYS = [
+    "io.johnsonlee.graphite.webgraph.WrappedDiscoveryLatencyBenchmark.coldWrappedCaseInsensitiveDiscovery[graphCount=1]",
+    "io.johnsonlee.graphite.webgraph.WrappedDiscoveryLatencyBenchmark.coldWrappedCaseInsensitiveDiscovery[graphCount=17]",
+    "io.johnsonlee.graphite.webgraph.WrappedDiscoveryLatencyBenchmark.wrappedCaseInsensitiveDiscovery[graphCount=1]",
+    "io.johnsonlee.graphite.webgraph.WrappedDiscoveryLatencyBenchmark.wrappedCaseInsensitiveDiscovery[graphCount=17]",
+    "io.johnsonlee.graphite.webgraph.AllFixtureWrappedDiscoveryLatencyBenchmark.broadlyDistributedClassPrefixCaseInsensitiveDiscovery",
+    "io.johnsonlee.graphite.webgraph.AllFixtureWrappedDiscoveryLatencyBenchmark.denseDistributedMethodContainsCaseInsensitiveDiscovery",
+    "io.johnsonlee.graphite.webgraph.AllFixtureWrappedDiscoveryLatencyBenchmark.earlyGraphClassPrefixCaseInsensitiveDiscovery",
+    "io.johnsonlee.graphite.webgraph.AllFixtureWrappedDiscoveryLatencyBenchmark.firstLastGraphBimodalClassPrefixCaseInsensitiveDiscovery",
+    "io.johnsonlee.graphite.webgraph.AllFixtureWrappedDiscoveryLatencyBenchmark.lateGraphClassPrefixCaseInsensitiveDiscovery",
+    "io.johnsonlee.graphite.webgraph.AllFixtureWrappedDiscoveryLatencyBenchmark.middleGraphsClassPrefixCaseInsensitiveDiscovery",
+    "io.johnsonlee.graphite.webgraph.AllFixtureWrappedDiscoveryLatencyBenchmark.skewedMixedClassMethodOperatorCaseInsensitiveDiscovery",
+    "io.johnsonlee.graphite.webgraph.AllFixtureWrappedDiscoveryLatencyBenchmark.zeroHitBroadContainsCaseInsensitiveDiscovery"
+];
 const LARGE_CORPUS_METRICS = [
     { key: "buildMs", label: "build", threshold: 20, minimum: 500, unit: "ms" },
     { key: "saveMs", label: "save", threshold: 25, minimum: 250, unit: "ms" },
@@ -243,6 +257,110 @@ export function renderJmhReport(comparison) {
     return `${lines.join("\n")}\n`;
 }
 
+export function compareLatencyBaseline(
+    fixedResults,
+    baseResults,
+    candidateResults,
+    regressionThreshold = 15,
+    minimumSpeedup = 50,
+    expectedKeys = null
+) {
+    const regression = compareJmh(baseResults, candidateResults, regressionThreshold);
+    const fixed = new Map(fixedResults.map((result) => [benchmarkKey(result), result]));
+    const candidate = new Map(candidateResults.map((result) => [benchmarkKey(result), result]));
+    const errors = [...regression.errors];
+    const regressionRows = new Map(regression.rows.map((row) => [row.key, row]));
+    const rows = [];
+
+    const actualKeys = [...new Set([...fixed.keys(), ...candidate.keys()])].sort();
+    const keys = expectedKeys === null ? actualKeys : [...expectedKeys].sort();
+    if (expectedKeys !== null) {
+        const expected = new Set(expectedKeys);
+        for (const [revision, results] of [
+            ["fixed baseline", fixedResults],
+            ["PR base", baseResults],
+            ["candidate", candidateResults]
+        ]) {
+            const actual = new Set(results.map(benchmarkKey));
+            for (const key of expected) {
+                if (!actual.has(key)) errors.push(`${revision}: missing expected latency benchmark ${key}`);
+            }
+            for (const key of actual) {
+                if (!expected.has(key)) errors.push(`${revision}: unexpected latency benchmark ${key}`);
+            }
+        }
+    }
+
+    for (const key of keys) {
+        const baseline = fixed.get(key);
+        const current = candidate.get(key);
+        const baseRow = regressionRows.get(key);
+        if (baseline === undefined || current === undefined || baseRow === undefined) {
+            errors.push(`${key}: missing from fixed baseline, PR base, or candidate results`);
+            continue;
+        }
+        const fixedScore = finiteNumber(baseline.primaryMetric?.score);
+        const candidateScore = finiteNumber(current.primaryMetric?.score);
+        if (fixedScore === null || candidateScore === null || fixedScore <= 0 || candidateScore <= 0) {
+            errors.push(`${key}: invalid fixed-baseline score`);
+            continue;
+        }
+        if (!isLowerBetter(baseline.mode) || baseline.mode !== current.mode ||
+            baseline.primaryMetric?.scoreUnit !== current.primaryMetric?.scoreUnit
+        ) {
+            errors.push(`${key}: fixed baseline and candidate use incompatible latency metrics`);
+            continue;
+        }
+        const speedup = ((fixedScore / candidateScore) - 1) * 100;
+        const improvementSeparated = confidenceSeparates(
+            confidenceBounds(current.primaryMetric ?? {}),
+            confidenceBounds(baseline.primaryMetric ?? {}),
+            true
+        );
+        const improvementBlocked = speedup < minimumSpeedup || !improvementSeparated;
+        rows.push({
+            ...baseRow,
+            fixedScore,
+            speedup,
+            minimumSpeedup,
+            improvementSeparated,
+            improvementBlocked,
+            blocked: baseRow.blocked || improvementBlocked
+        });
+    }
+
+    if (rows.length === 0) errors.push("No comparable latency benchmarks were found");
+    return {
+        passed: errors.length === 0 && rows.every((row) => !row.blocked),
+        errors,
+        rows
+    };
+}
+
+export function renderLatencyBaselineReport(comparison) {
+    const lines = [
+        "### Wrapped case-insensitive query latency",
+        "",
+        "The PR must remain at least 50% faster than the fixed pre-PR-95 baseline and must not regress",
+        "more than 15% against the PR base with separated 99.9% confidence intervals.",
+        "",
+        "| Benchmark | Pre-PR-95 | PR base | PR | Speedup vs fixed | Regression vs base | Gate |",
+        "|---|---:|---:|---:|---:|---:|:---:|"
+    ];
+    for (const row of comparison.rows) {
+        lines.push(
+            `| \`${shortBenchmarkName(row.key)}\` | ${formatScore(row.fixedScore)} ${row.unit} | ` +
+            `${formatScore(row.baseScore)} ${row.unit} | ${formatScore(row.candidateScore)} ${row.unit} | ` +
+            `${formatDelta(row.speedup)} | ${formatDelta(row.delta)} | ` +
+            `**${row.blocked ? "FAIL" : "PASS"}** |`
+        );
+    }
+    if (comparison.errors.length > 0) {
+        lines.push("", "Errors:", ...comparison.errors.map((error) => `- ${error}`));
+    }
+    return `${lines.join("\n")}\n`;
+}
+
 export function parseLargeCorpusLog(contents) {
     const results = new Map();
     for (const line of contents.split(/\r?\n/)) {
@@ -387,7 +505,8 @@ export function renderLargeCorpusReport(comparison) {
 export function aggregateReports(directory, metadata) {
     const components = [
         { name: "method-level", report: "method-report.md", status: "method-status.json" },
-        { name: "large-corpus", report: "large-corpus-report.md", status: "large-corpus-status.json" }
+        { name: "large-corpus", report: "large-corpus-report.md", status: "large-corpus-status.json" },
+        { name: "wrapped-query-latency", report: "latency-report.md", status: "latency-status.json" }
     ];
     const errors = [];
     const reports = [];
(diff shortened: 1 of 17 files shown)


graphite-webgraph/src/main/kotlin/io/johnsonlee/graphite/webgraph/MappedWebGraphBackedGraph.kt
```diff
@@ -334,6 +398,7 @@ internal class MappedWebGraphBackedGraph(
 
     override fun close() {
         clearStringPropertyIndexes()
+        stringPropertyNodeOrders = null
```
johnsonlee: The cold benchmark reset leaves `stringPropertyNodeOrders` warm. Both cold harnesses only invoke `clearStringPropertyIndexes`, while the synthetic setup already runs a query and JMH warmup initializes this candidate-only `IntArray`; every measured cold invocation therefore skips allocating and populating one entry per node. Across the real fixtures that is 19,091,048 entries, and fixed/base do not have the corresponding cache, so the gate does not measure comparable first/cold-query latency. Please reset this cache as part of the cold path (or reopen graphs per cold invocation) so the claimed cold result includes the new ordering cost.

.github/scripts/benchmark-gate.mjs
```diff
@@ -243,6 +257,110 @@ export function renderJmhReport(comparison) {
...
+            continue;
+        }
+        if (!isLowerBetter(baseline.mode) || baseline.mode !== current.mode ||
+            baseline.primaryMetric?.scoreUnit !== current.primaryMetric?.scoreUnit
+        ) {
+            errors.push(`${key}: fixed baseline and candidate use incompatible latency metrics`);
+            continue;
+        }
+        const speedup = ((fixedScore / candidateScore) - 1) * 100;
+        const improvementSeparated = confidenceSeparates(
```
johnsonlee: `confidenceSeparates` returns `true` when either confidence interval is missing or non-finite, so the fixed-baseline improvement gate accepts incomplete JMH evidence. For example, fixed/base/candidate results with scores 100/20/20 and no `scoreConfidence` produce `passed: true` and `improvementSeparated: true`. That contradicts the generated report, which says the 99.9% intervals are separated, and lets a malformed result prove the required speedup. Please validate that both bounds exist here and fail closed when they do not, with a regression test.

.github/workflows/benchmark.yml
```diff
@@ -186,10 +187,182 @@ jobs:
...
+          java -jar "${BASE_JAR}" \
+            'io.johnsonlee.graphite.webgraph.(WrappedDiscoveryLatencyBenchmark|AllFixtureWrappedDiscoveryLatencyBenchmark).*CaseInsensitiveDiscovery' \
+            -jvmArgsAppend "${GRAPH_JVM_ARGS}" -foe true -prof gc -rf json \
+            -rff "${PWD}/benchmark-results/base-latency-confirmation.json"
+          java -jar "${FIXED_JAR}" \
+            'io.johnsonlee.graphite.webgraph.(WrappedDiscoveryLatencyBenchmark|AllFixtureWrappedDiscoveryLatencyBenchmark).*CaseInsensitiveDiscovery' \
+            -jvmArgsAppend "${GRAPH_JVM_ARGS}" -foe true -prof gc -rf json \
+            -rff "${PWD}/benchmark-results/fixed-latency-confirmation.json"
+          node candidate/.github/scripts/benchmark-gate.mjs compare-latency-baseline \
+            --fixed benchmark-results/fixed-latency-confirmation.json \
```
johnsonlee: This discards the initial latency comparison and evaluates only the confirmation run. If benchmark A triggers the rerun, then recovers, while unrelated benchmark B fails only in the reverse-order run, the required gate still fails even though no benchmark repeated its failure. The documentation says a suspected failure blocks only after the reverse-order confirmation; please combine initial and confirmation rows by key, analogous to `confirmJmh`, so only repeated failures remain blocking.

johnsonlee (commented):
PR review: three benchmark-gate correctness issues found; details are inline.

graphite-webgraph/src/main/kotlin/io/johnsonlee/graphite/webgraph/MappedWebGraphBackedGraph.kt
```diff
@@ -125,14 +134,28 @@ internal class MappedWebGraphBackedGraph(
...
+        val orders = stringPropertyNodeOrders ?: synchronized(this) {
+            stringPropertyNodeOrders ?: buildStringPropertyNodeOrders().also { stringPropertyNodeOrders = it }
+        }
+        return orders.getOrElse(node.id.value) { -1 }.toLong()
+    }
+
+    private fun buildStringPropertyNodeOrders(): IntArray {
+        val orders = IntArray(nodeOffsets.size) { -1 }
+        var order = 0
+        for (nodeId in nodeTypeIndex.ids(Node::class.java)) orders[nodeId] = order++
```
johnsonlee: `buildStringPropertyNodeOrders()` performs an O(total nodes) traversal and allocates `IntArray(nodeOffsets.size)` without charging the active `GraphWorkConsumer` or checking cancellation. This is reached from `mergeNodeSequences -> stringPropertyNodeOrder` even for work-budgeted transformed OR queries, before the lookup iterators consume their tracked work. I reproduced it with three 540-node mapped graphs and `CypherExecutionBudget(maxWorkUnits=10)`: after `CypherBudgetExceededException`, reflection shows the first graph order array was already fully initialized. On the 19,091,048-node fixture this bypasses the work/cancellation budget for a material CPU/heap operation. Please make order construction budget-aware (or persist/precompute it outside query execution) and add a mapped budget regression test.

johnsonlee (commented):
Required per-module coverage is currently below the 98% gate; the two largest untested new paths are inline.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
```diff
@@ -683,25 +691,63 @@ class QueryPipeline private constructor(
...
                 }
                 if (!ownedByEarlierFilter) yield(node)
             }
         }
     }
 
+    private fun mergeNodeSequences(
+        sequences: List<Sequence<Node>>,
+        nodeOrder: (Node) -> Long
+    ): Sequence<Node> = sequence {
```
johnsonlee, line 705: The required CI is failing `cypher` coverage at 96.86%. The generated Kover XML marks the entire new `mergeNodeSequences` body uncovered by cypher-module tests; the mapped integration assertion in `GraphStoreTest` belongs to another module and cannot cover this code for the per-module gate. Please add a `QueryPipelineTest` graph that implements `StringPropertyLookupOrder` and exercises overlapping/empty ordered streams, deduplication, and LIMIT ordering.

graphite-webgraph/src/main/kotlin/io/johnsonlee/graphite/webgraph/MappedWebGraphBackedGraph.kt
```diff
@@ -225,9 +271,27 @@ internal class MappedWebGraphBackedGraph(
...
                 }
                 admitted = true
+                val stringCount = stringTable.size()
+                if (stringCount <= MAX_RAW_STRING_MATCH_STATE_BYTES) {
+                    matchStates = ByteArray(stringCount)
+                }
             }
             val stringId = rawStringPropertyIndex(nodeId, type, property)
-            if (stringId != null && stringMatches(stringTable.get(stringId), mode, expected)) {
+            val states = matchStates
```
johnsonlee, line 267: The required CI is also failing `webgraph` coverage at 94.73%, and this new raw-scan match-state machine is not exercised past the 256-node admission threshold: Kover reports the null-cache and cached match/miss branches uncovered. Please add a mapped-graph test with more than 256 nodes and repeated matching and non-matching string IDs, then assert order, limit, and work accounting through all three states.

.github/workflows/benchmark.yml
```diff
@@ -186,10 +187,182 @@ jobs:
...
+    - name: Verify all real-fixture query results
+      run: |
+        mkdir -p benchmark-results
+        for REVISION in fixed base candidate; do
+          JMH_JAR=$(find "${REVISION}/graphite-webgraph/build/libs" -name '*-jmh.jar' -print -quit)
+          test -n "${JMH_JAR}"
+          java -Xmx8g -cp "${JMH_JAR}" \
+            io.johnsonlee.graphite.webgraph.AllFixtureQueryCorrectness \
+            "${PWD}/fixture-graphs" \
+            | grep '^ALL_FIXTURE_QUERY_RESULT' \
```
johnsonlee: This pipeline can hide a failing correctness process because the default `run` shell is invoked with `bash -e`, not `pipefail`. `AllFixtureQueryCorrectness` prints all eight markers inside `try` and only then closes graphs in `finally`; if cleanup/resource close throws, Java exits non-zero after producing eight valid lines, `grep` still exits 0, and the subsequent line-count and `cmp` checks can all pass. That contradicts the PR promise that cleanup/resource errors fail closed. Please enable `set -o pipefail` for this step (and preferably preserve the full Java log before extracting markers).

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
johnsonlee: Verified on `87c74c2`: the new cypher-module test exercises overlapping ordered streams and deduplication, local `:cypher:test`/detekt pass, Kover now marks the full merge body covered, and module line coverage is 98.1536%. Resolving this thread.

graphite-webgraph/src/main/kotlin/io/johnsonlee/graphite/webgraph/MappedWebGraphBackedGraph.kt
johnsonlee: `87c74c2` restores the aggregate webgraph gate (local Kover: 98.001%) and adds work-aware coverage, but the requested raw cache-state coverage is still incomplete: current XML still reports misses on the `states == null` body at line 282 and partial instructions on both cached MATCH/MISS returns at lines 287-288. Keeping this thread open until the >256-node repeated-string state transitions are directly asserted.

graphite-webgraph/src/main/kotlin/io/johnsonlee/graphite/webgraph/MappedWebGraphBackedGraph.kt
johnsonlee: Expanded performance audit: this cache also introduces an unmeasured retained-heap regression. The four fixtures contain 19,091,048 nodes, so these arrays alone retain at least `19,091,048 * 4 = 76,364,192` bytes (72.8 MiB, plus headers) after the first wrapped OR query. The existing large-corpus heap gate never executes this query shape, and `compareLatencyBaseline` ignores JMH secondary/heap metrics; warmup also hides the one-time allocation. Please either avoid the per-node retained cache or add explicit cold allocation/retained-heap evidence and a blocking memory threshold, so the latency win is not purchased with an unchecked memory regression.

graphite-webgraph/src/main/kotlin/io/johnsonlee/graphite/webgraph/MappedWebGraphBackedGraph.kt
johnsonlee: Quantified on the same four persisted fixtures with current head: the first `stringPropertyNodeOrder` build took Android 89 ms, Tika 78 ms, Hive 128 ms, and Kotlin Compiler 65 ms (about 360 ms total), while allocating the exact 72.8 MiB of rank arrays. The advertised candidate real-query mean is 1.624 s, so warmup is excluding a material ~22% first-query cost. This reinforces that the cold gate must reset/measure the order cache, not only property indexes.

graphite-core/src/main/kotlin/io/johnsonlee/graphite/graph/Graph.kt
```diff
@@ -54,6 +59,36 @@ interface WorkAwareStringPropertyLookup : StringPropertyLookup {
...
+        transform: StringValueTransform,
+        mode: StringMatchMode,
+        expected: String,
+        limit: Int,
+        workConsumer: GraphWorkConsumer
+    ): Sequence<T>?
+}
+
+/** Optional ordering capability used to preserve the graph's canonical node traversal order. */
+interface StringPropertyLookupOrder {
```
johnsonlee, line 102: `mergeNodeSequences` is a k-way merge and is correct only if every lookup sequence is already monotonic by `stringPropertyNodeOrder`, but neither `StringPropertyLookup`/`TransformedStringPropertyLookup` nor this new capability states that requirement. A conforming graph can return stream A `[node2]` and stream B `[node1, node0]` with canonical orders 0<1<2; the merge yields node1 first, so `RETURN DISTINCT ... LIMIT 1` returns the wrong canonical row. The new test only supplies pre-sorted streams and misses this case. Please either make sorted lookup streams an explicit/enforced contract (for both lookup interfaces) or make the merge robust to unsorted sequences, with a regression test.

graphite-webgraph/src/main/kotlin/io/johnsonlee/graphite/webgraph/MappedWebGraphBackedGraph.kt
```diff
@@ -225,9 +271,27 @@ internal class MappedWebGraphBackedGraph(
                     }
                 }
                 admitted = true
+                val stringCount = stringTable.size()
+                if (stringCount <= MAX_RAW_STRING_MATCH_STATE_BYTES) {
```
johnsonlee: `MAX_RAW_STRING_MATCH_STATE_BYTES` caps one predicate iterator, not the aggregate query allocation. The two nested k-way merges eagerly call `hasNext()` on every candidate stream before the first yield; a valid `MATCH (n)` disjunction over the eight mapped properties (Enum name, Local name, Field class/name, and four CallSite fields) can therefore suspend eight raw scans after admission, each retaining its own `ByteArray(stringTable.size)`. At the allowed per-array limit that is 8 * 16 MiB = 128 MiB for one graph/query, and concurrent queries multiply it; these allocations are also outside `GraphWorkConsumer`. Android currently uses ~0.51 MiB per stream (~2.03 MiB for the production four-field shape), but the public limit permits the much larger case. Please enforce a per-query/per-graph aggregate memory/work budget or share the string-match state instead of allocating one full-table array per predicate, with a stress regression test.

graphite-webgraph/src/main/kotlin/io/johnsonlee/graphite/webgraph/MappedWebGraphBackedGraph.kt
johnsonlee: Fixed in a19540d. Canonical order now comes directly from the persisted node byte offset, so the 19,091,048-entry order arrays and their approximately 72.8 MiB allocation no longer exist. The cold reset also clears the new bounded raw-match state. Full check and coverage gates pass locally.

.github/workflows/benchmark.yml
johnsonlee: Fixed in a19540d. Added confirmLatencyBaseline and changed the workflow to combine the initial status with the reverse-order comparison by benchmark key. A row blocks only when that same row is blocked in both rounds; the A-first/B-second regression case is covered by a unit test.

.github/scripts/benchmark-gate.mjs
johnsonlee: Fixed in a19540d. Fixed-baseline speedup now requires finite bounds for both fixed and candidate metrics, records an explicit error when either is missing, and sets improvementSeparated=false. Added a malformed-JMH regression test proving the gate fails closed.

graphite-webgraph/src/main/kotlin/io/johnsonlee/graphite/webgraph/MappedWebGraphBackedGraph.kt
johnsonlee: Fixed in a19540d. Added a 261-node mapped regression that forces the post-admission UNKNOWN to MATCH/MISS transitions, repeated cached MATCH/MISS reads, exact result order and limit, and exact 261-unit work accounting. It also verifies budgeted lookup retains no raw match state. Full webgraph coverage gate passes.

graphite-webgraph/src/main/kotlin/io/johnsonlee/graphite/webgraph/MappedWebGraphBackedGraph.kt
johnsonlee: Fixed in a19540d. The per-node order cache was removed entirely. stringPropertyNodeOrder is now an O(1), allocation-free read of the already persisted node offset, so it performs no traversal, retains no additional order heap, and cannot bypass the work budget. Budgeted raw lookup also avoids allocating match-state arrays.

graphite-webgraph/src/main/kotlin/io/johnsonlee/graphite/webgraph/MappedWebGraphBackedGraph.kt
johnsonlee: Fixed in a19540d. Raw match state is now shared per graph and predicate behind one aggregate 16 MiB/32-entry cap instead of one 16 MiB array per iterator. Concurrent identical predicates share the same array, distinct predicates cannot exceed the graph cap, and budgeted lookups allocate none. Added a small-bound stress test that proves overflow is refused and existing state is shared.

.github/workflows/benchmark.yml
johnsonlee: Fixed in a19540d. The correctness step now enables pipefail, tees the complete Java output to a per-revision log, and extracts markers only after Java succeeds. Cleanup or close failures therefore fail the step even if all eight markers were printed.

graphite-core/src/main/kotlin/io/johnsonlee/graphite/graph/Graph.kt
johnsonlee: Fixed in a19540d. The monotonic lookup requirement is now explicit on both StringPropertyLookup and TransformedStringPropertyLookup whenever StringPropertyLookupOrder is implemented, and the merge checks the invariant while advancing each cursor. Added a regression test with an unsorted conforming-capability stream that fails with the contract error. Mapped lookups use persisted byte offsets as the stable canonical key.

graphite-webgraph/src/main/kotlin/io/johnsonlee/graphite/webgraph/MappedWebGraphBackedGraph.kt
```diff
@@ -225,9 +258,29 @@ internal class MappedWebGraphBackedGraph(
                     }
                 }
                 admitted = true
+                if (workConsumer == null) {
```
johnsonlee: Disabling match state whenever `workConsumer` is present introduces a material regression in the budgeted path used by `graphite-explore`. On the same persisted Android graph, a zero-hit transformed `caller_class` scan over 1,706,834 `CallSiteNode`s takes 160–166 ms on `87c74c2` but 321–341 ms on `a19540d` (2.0–2.1x slower), with exactly the same 1,706,834 charged work units. The old state only memoized string matches while each node was still charged; it did not bypass `consume()`. Please keep a query-local, aggregate-bounded state for budgeted scans (or an equivalent design) and add a budgeted mapped-string latency regression, so the memory fix does not trade away production query performance.

graphite-webgraph/src/main/kotlin/io/johnsonlee/graphite/webgraph/MappedWebGraphBackedGraph.kt
```diff
@@ -471,23 +531,68 @@ private const val TRIGRAM_POSTING_ESTIMATED_BYTES = 12L
...
+    private val maxRetainedBytes: Long = MAX_RAW_STRING_MATCH_STATE_BYTES.toLong(),
+    private val maxEntries: Int = MAX_RAW_STRING_MATCH_STATES
+) {
+    private val states = LinkedHashMap<RawStringMatchKey, ByteArray>()
+    private var bytes = 0L
+
+    @Synchronized
+    fun stateFor(key: RawStringMatchKey, stringCount: Int): ByteArray? {
+        val existing = states[key]
+        val requiredBytes = PRIMITIVE_ARRAY_HEADER_ESTIMATED_BYTES + stringCount.toLong()
```
johnsonlee: `requiredBytes` counts only the `ByteArray` header/payload, but this map also strongly retains `RawStringMatchKey.expected` and the map entry. Query literal length is not bounded, and `stateFor` is still called after `StringPropertyAdmissions.admit` rejects an oversized (>64 KiB estimate) predicate, so 32 distinct very large literals can remain reachable until clear/close while `retainedBytes()` still reports at most 16 MiB. The existing admission/matching-cache estimators already account for `expected.length * Char.SIZE_BYTES`; please include key/entry/string bytes in this aggregate cap (and reject a state whose key cannot fit), with an oversized-key regression test.

.github/workflows/benchmark.yml
```diff
@@ -186,10 +187,265 @@ jobs:
...
+      with:
+        ref: ${{ github.event.pull_request.head.sha }}
+
+    - name: Download latency shard results
+      continue-on-error: true
+      uses: actions/download-artifact@v8
+      with:
+        pattern: benchmark-latency-shard-*-${{ github.event.pull_request.number }}-${{ github.run_attempt }}
+        path: benchmark-results
+        merge-multiple: true
```
johnsonlee, line 573: Every shard artifact contains the same raw filenames (`fixed-latency.json`, `base-latency.json`, `candidate-latency.json`, and possible confirmation files). `merge-multiple: true` flattens all five artifacts into this directory, so colliding files are overwritten and the uploaded “combined” artifact retains raw JMH evidence for only an arbitrary last shard. The later `benchmark-*` download also matches both the combined and all shard artifacts and repeats the collision. The uniquely named status files let the gate stay green, but the 12 reported results can no longer be audited against their raw scores/GC data or diagnosed after a failure. Please make every raw/result filename shard-qualified (or preserve per-shard directories and combine from them), and keep the aggregate download from re-merging shard artifacts.

graphite-webgraph/src/main/kotlin/io/johnsonlee/graphite/webgraph/MappedWebGraphBackedGraph.kt
johnsonlee: Fixed in da9f040. Budgeted scans now use the same graph-shared, aggregate-bounded match state while still calling workConsumer.consume() for every inspected node. The 261-node regression now asserts all 261 work units and retained memoization. Added MappedStringAdmissionBenchmark.budgetedTransformedZeroHit with repeated string IDs; the local 50,000-node 5-shot mean is 1.794 ms/op.

graphite-webgraph/src/main/kotlin/io/johnsonlee/graphite/webgraph/MappedWebGraphBackedGraph.kt
johnsonlee: Fixed in da9f040. The aggregate estimate now includes the ByteArray header/payload, map entry/key overhead, property string header/chars, and expected literal string header/chars. A state is rejected before allocation/retention when the complete estimate cannot fit. Added an oversized 128-character key regression with a 256-byte cap, asserting zero entries and zero retained bytes.

johnsonlee: Verified on `da9f040`: the estimator now charges the entry, array header/payload, property string, and expected string before admission; the oversized-key regression refuses the state and leaves both count and retained bytes at zero. Forced WebGraph tests/Detekt/Kover pass locally, `RawStringMatchStates` remains 16/16 lines covered, and aggregate WebGraph line coverage is 98.17%. Resolving this thread.

graphite-webgraph/src/main/kotlin/io/johnsonlee/graphite/webgraph/MappedWebGraphBackedGraph.kt
johnsonlee: The implementation regression is fixed on `da9f040`: repeating the same 1,706,834-node Android scan five times gave 164–172 ms versus 161–167 ms on `87c74c2` (median delta about +1.7%), with exactly 1,706,834 work units in every run. Tests/Detekt/Kover also pass. Keeping the thread open for the requested regression protection, though: `MappedStringAdmissionBenchmark.budgetedTransformedZeroHit` is not executed by any required job. `method-level` builds only `:cypher:jmhJar` and filters `CypherBenchmark.*`; the WebGraph latency job filters only the two wrapped-discovery benchmark classes. Please wire this benchmark into a blocking base-vs-candidate gate (with a compatible baseline harness) so this 2x production-path regression cannot silently return.

johnsonlee: Fixed in 2cf05cd. `budgeted-mapped-string-latency` is now a blocking required-job dependency of `benchmark-regression-gate`. It installs the identical `MappedStringAdmissionBenchmark` harness in compatible fixed baseline `87c74c2`, builds both WebGraph JMH jars, and compares `budgetedTransformedZeroHit` baseline-first against the candidate with the existing 15%/separated-CI rule plus candidate-first reverse-order confirmation. Its report/status is included in the aggregate fail-closed report. I also compiled the overlaid harness at `87c74c2` locally and executed the exact benchmark selector successfully.

.github/workflows/benchmark.yml
johnsonlee: Fixed in 2cf05cd. Every initial, raw, and confirmation latency filename is now shard-qualified. Shard artifacts were renamed from `benchmark-latency-shard-*` to `latency-shard-*`; only the intermediate combiner downloads that pattern. The final aggregate still downloads `benchmark-*`, so it receives only the canonical combined latency artifact and no longer re-merges shard artifacts. This preserves all 12 cases’ raw JMH/GC evidence without collisions.

.github/workflows/benchmark.yml
```diff
@@ -108,6 +110,97 @@ jobs:
...
+        java -jar "${BASELINE_JAR}" \
+          'io.johnsonlee.graphite.webgraph.MappedStringAdmissionBenchmark.budgetedTransformedZeroHit' \
+          -foe true -rf json -rff "${PWD}/benchmark-results/baseline-budgeted-string.json"
+        java -jar "${CANDIDATE_JAR}" \
+          'io.johnsonlee.graphite.webgraph.MappedStringAdmissionBenchmark.budgetedTransformedZeroHit' \
+          -foe true -rf json -rff "${PWD}/benchmark-results/candidate-budgeted-string.json"
+
+    - name: Compare budgeted lookup latency
+      run: |
+        if node candidate/.github/scripts/benchmark-gate.mjs compare-jmh \
```
johnsonlee, line 167: This blocking gate still does not catch the regression it was added for. I overlaid the exact candidate harness onto known-bad `a19540d` and ran it against `87c74c2` on the same machine: baseline was 1.891 ms/op, bad candidate 4.747 ms/op (+151.0%), but the 5-shot SingleShot 99.9% intervals overlapped, so this exact `compare-jmh` command returned `passed=true` / **NOISE** and never ran confirmation. The current CI artifact is similarly noisy (baseline CI 4.52–8.69 ms, candidate -0.89–13.66 ms). Please make this benchmark statistically stable (for example AverageTime/batched samples, more measurements/forks) and prove the known-bad negative control fails, or use a gate-specific comparator that confirms any >15% score regression and blocks when it repeats. A required green job that accepts the historical 2x regression is not regression protection.

johnsonlee: Fixed in b5fd2b9. The budgeted job now passes `--threshold-only` on both initial and confirmation comparisons: any score regression above 15% forces the reverse-order run regardless of CI overlap, and a repeated score regression above 15% blocks. Other JMH gates retain the separated-CI policy. Added focused comparator/confirmation tests (24/24 script tests pass). I also ran the requested real negative control with the exact overlaid harness: `87c74c2` 1.707 ms/op vs known-bad `a19540d` 4.911 ms/op (+187.8%); reverse order 1.859 vs 4.836 ms/op (+160.2%). Initial comparator exit=1, confirmation exit=1, final gate **FAIL** despite overlapping SingleShot intervals.

johnsonlee: Verified on pushed `b5fd2b9` with the real negative control, not only synthetic rows. Using identical overlaid harnesses on the same machine: initial `87c74c2` 1.891 ms/op vs known-bad `a19540d` 4.747 ms/op (+151.0%) now triggers reverse-order confirmation despite overlapping intervals; the reverse run was 1.723 vs 4.828 ms/op (+180.2%) and the final status is `passed=false` / FAIL. The current fixed candidate remains below the baseline in CI, and all 24 gate-script tests pass. Resolving this thread; latest-head CI remains under observation.

PR #98, opened 2026-08-28
Require successful budgeted queries in latency gates

johnsonlee:
## Summary
- execute current-base and PR latency JMH through the budget-aware `CrossGraphCypherExecutor` with an explicit 25,000,000-unit benchmark-only budget; work tracking remains enabled while the HTTP semaphore is excluded, and the production default remains 250,000
- fail every timed invocation on an exception or unexpected row count, and compare complete columns, ordered rows, values, and graph provenance digests across the fixed pre-PR-95 baseline, current base, and PR before timing
- build the benchmark corpus from every real fixture JAR and cover prefix, contains, mixed-operator, zero-hit, dense/sparse, early/middle/late, bimodal, skewed, and broadly distributed targets
- cache prepared fixture graphs by graph-building source fingerprint and run the wrapped-query cases in parallel shards
- retain the mapped full-scan guard as SingleShotTime, stabilized with 3 forks, 5 warmup invocations, and 20 measured invocations per fork

The latency benchmark calls the executor directly and never times an HTTP response, so a fast 429 cannot become a passing latency sample. Query failures, budget exhaustion, missing or truncated results, row-count mismatches, and digest mismatches fail the shard under `-foe true`. Production HTTP concurrency and 429 behavior remain covered by `graphite-explore` tests.

## Verification
- `./gradlew :cypher:test :explore:test :webgraph:test :webgraph:jmhJar detekt --no-daemon`
- `./gradlew :webgraph:test :webgraph:koverLog --no-daemon` — 98.0227% webgraph line coverage
- `node --test .github/scripts/benchmark-gate.test.mjs`
- `actionlint .github/workflows/benchmark.yml`
- identical same-runner fixed/base/PR correctness preflight and latency comparison
- required GitHub Actions unit/coverage and benchmark-regression-gate workflows: PASS

## Performance
The benchmark result comment is the canonical same-runner report.

- Wrapped real-fixture cases: 0.9275–3.677 s/op on the PR versus 16.329–22.945 s/op on the fixed pre-PR-95 baseline, a 350.7%–2129.4% speedup. Versus current main, changes range from -3.1% to +2.3%; no wrapped-query regression.
- Synthetic 17-graph cases: 15.080 ms/op cold and 0.4952 ms/op warm versus 104.523 ms/op and 104.615 ms/op pre-PR-95. Versus current main: -0.5% cold and +11.6% warm, both passing.
- Method-level: the mapped SingleShot full scan is 3.966 ms/op versus 3.963 ms/op on main (+0.1%); all method-level gates pass, with no confirmed regression.
- End to end: all Hive, Kotlin compiler, and Tika build/save/load/query/pipeline gates pass; query deltas versus main are +8.3%, +2.1%, and -2.9%, with no confirmed end-to-end regression.

[View the benchmark run and artifacts](https://github.com/johnsonlee/graphite/actions/runs/33234406373)

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -565,8 +565,6 @@ jobs:
           LEGACY_BUDGET_ARGS=()
           if [ "${REVISION}" = fixed ]; then
             LEGACY_BUDGET_ARGS=(-Dgraphite.benchmark.allowLegacyUnbudgetedExecutor=true)
-          elif [ "${REVISION}" = base ]; then
-            LEGACY_BUDGET_ARGS=(-Dgraphite.benchmark.allowLegacyBudgetDefault=true)
           fi
           java -Xmx8g -cp "${JMH_JAR}" \
             "${LEGACY_BUDGET_ARGS[@]}" \
@@ -599,15 +597,13 @@ jobs:
           -Dkotlin.compiler.graph.path=${PWD}/fixture-graphs/kotlin-compiler"
         FIXED_GRAPH_JVM_ARGS="${GRAPH_JVM_ARGS} \
           -Dgraphite.benchmark.allowLegacyUnbudgetedExecutor=true"
-        BASE_GRAPH_JVM_ARGS="${GRAPH_JVM_ARGS} \
-          -Dgraphite.benchmark.allowLegacyBudgetDefault=true"
         java -jar "${FIXED_JAR}" \
           "${BENCHMARK_FILTER}" \
           -jvmArgsAppend "${FIXED_GRAPH_JVM_ARGS}" -foe true -prof gc -rf json \
           -rff "${PWD}/benchmark-results/fixed-latency-${SHARD}.json"
         java -jar "${BASE_JAR}" \
           "${BENCHMARK_FILTER}" \
-          -jvmArgsAppend "${BASE_GRAPH_JVM_ARGS}" -foe true -prof gc -rf json \
+          -jvmArgsAppend "${GRAPH_JVM_ARGS}" -foe true -prof gc -rf json \
           -rff "${PWD}/benchmark-results/base-latency-${SHARD}.json"
         java -jar "${CANDIDATE_JAR}" \
           "${BENCHMARK_FILTER}" \
@@ -640,15 +636,13 @@ jobs:
             -Dkotlin.compiler.graph.path=${PWD}/fixture-graphs/kotlin-compiler"
           FIXED_GRAPH_JVM_ARGS="${GRAPH_JVM_ARGS} \
             -Dgraphite.benchmark.allowLegacyUnbudgetedExecutor=true"
-          BASE_GRAPH_JVM_ARGS="${GRAPH_JVM_ARGS} \
-            -Dgraphite.benchmark.allowLegacyBudgetDefault=true"
           java -jar "${CANDIDATE_JAR}" \
             "${BENCHMARK_FILTER}" \
             -jvmArgsAppend "${GRAPH_JVM_ARGS}" -foe true -prof gc -rf json \
             -rff "${PWD}/benchmark-results/candidate-latency-confirmation-${SHARD}.json"
           java -jar "${BASE_JAR}" \
             "${BENCHMARK_FILTER}" \
-            -jvmArgsAppend "${BASE_GRAPH_JVM_ARGS}" -foe true -prof gc -rf json \
+            -jvmArgsAppend "${GRAPH_JVM_ARGS}" -foe true -prof gc -rf json \
             -rff "${PWD}/benchmark-results/base-latency-confirmation-${SHARD}.json"
           java -jar "${FIXED_JAR}" \
             "${BENCHMARK_FILTER}" \
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -564,12 +564,14 @@ jobs:
         for REVISION in fixed base candidate; do
           JMH_JAR=$(find "${REVISION}/graphite-webgraph/build/libs" -name '*-jmh.jar' -print -quit)
           test -n "${JMH_JAR}"
-          LEGACY_BUDGET_ARG=()
+          LEGACY_BUDGET_ARGS=()
           if [ "${REVISION}" = fixed ]; then
-            LEGACY_BUDGET_ARG=(-Dgraphite.benchmark.allowLegacyUnbudgetedExecutor=true)
+            LEGACY_BUDGET_ARGS=(-Dgraphite.benchmark.allowLegacyUnbudgetedExecutor=true)
+          elif [ "${REVISION}" = base ]; then
+            LEGACY_BUDGET_ARGS=(-Dgraphite.benchmark.allowLegacyBudgetDefault=true)
           fi
           java -Xmx8g -cp "${JMH_JAR}" \
-            "${LEGACY_BUDGET_ARG[@]}" \
+            "${LEGACY_BUDGET_ARGS[@]}" \
             io.johnsonlee.graphite.webgraph.AllFixtureBenchmarkQueryCorrectness \
             "${PWD}/fixture-graphs" \
             | tee "benchmark-results/${REVISION}-fixture-correctness.log"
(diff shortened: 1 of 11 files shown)


graphite-webgraph/src/jmh/kotlin/io/johnsonlee/graphite/webgraph/AllFixtureWrappedDiscoveryLatencyBenchmark.kt
```diff
@@ -299,13 +272,27 @@ private val EXPECTED_DISTRIBUTIONS = mapOf(
...
+    BenchmarkFixtureQueryCase("skewedMixedClassMethodOperator", SKEWED_MIXED_OPERATOR_QUERY, 250)
 )
+
+/**
+ * Latency samples represent completed queries, not HTTP admission failures.
+ * Concurrency and work-budget rejection are tested by graphite-explore; this
+ * harness bypasses those request guards and validates each result explicitly.
+ */
+internal fun successfulLatencyExecutor(graphs: List<CypherGraph>): CrossGraphCypherExecutor =
+    CrossGraphCypherExecutor(graphs)
```
johnsonlee: **[P1] Keep the latency gate on the production budgeted execution path.** This constructor sets `workTrackingEnabled=false`, while `/api/cypher/graphs` constructs `CrossGraphCypherExecutor(..., executionContext)` and therefore exercises different candidate-lookup/work-accounting branches. A budgeted-only regression (the class of regression PR #97 addressed), or the fact that these real requests still exceed the production budget and return 429, can now pass this gate as long as the unbudgeted query succeeds. Per-invocation row checks do not cover that difference. Please execute the budgeted constructor directly (without the HTTP semaphore) using the intended production budget; if the corpus query cannot complete under that budget, that is a functional failure to fix or an intentional product-budget change to test, rather than something the benchmark should bypass.

.github/workflows/benchmark.yml
```diff
@@ -298,28 +298,40 @@ jobs:
...
+        key: ${{ runner.os }}-${{ runner.arch }}-wrapped-query-fixtures-v2-${{ hashFiles('candidate/gradle/libs.versions.toml', 'candidate/gradle/wrapper/gradle-wrapper.properties', 'candidate/settings.gradle.kts', 'candidate/**/build.gradle.kts', 'candidate/graphite-core/src/main/**', 'candidate/graphite-sootup/src/main/**', 'candidate/graphite-webgraph/src/main/**', 'candidate/graphite-webgraph/src/jmh/kotlin/io/johnsonlee/graphite/webgraph/BenchmarkCorpus.kt', 'candidate/graphite-webgraph/src/jmh/kotlin/io/johnsonlee/graphite/webgraph/AllFixtureBenchmarkGraphPreparation.kt') }}
+        restore-keys: |
+          ${{ runner.os }}-${{ runner.arch }}-wrapped-query-fixtures-v1-a7ddb452997f5699a3233189a81be5020d3d8dd2f52b6cabee950de13689ed10
```
johnsonlee: **[P1] Do not let the migration fallback defeat fixture invalidation.** On every future `v2-${hashFiles(...)}` miss (including a real change under `core`, `sootup`, or `webgraph/src/main`), this fixed v1 cache is restored, `.complete` makes `ready=true`, graph preparation is skipped, and the post-cache step can publish those stale graphs under the new v2 key. The benchmark would then compare all revisions on graphs produced by old graph-building/storage code, despite the primary key correctly changing. Since the v2 cache has now been populated, please remove this fallback, or constrain a one-time migration so it can only apply to the exact known-compatible source fingerprint and cannot seed later source hashes.

graphite-webgraph/src/jmh/kotlin/io/johnsonlee/graphite/webgraph/AllFixtureWrappedDiscoveryLatencyBenchmark.kt
johnsonlee: Confirmed with the PR history: the budgeted version at `4ab1f34` failed [Actions run 33211606769](https://github.com/johnsonlee/graphite/actions/runs/33211606769) during real-fixture correctness with `Cypher work budget exceeded after 250000 graph work units`, and all four real latency shards failed. The final commit became green by replacing that path with this unbudgeted constructor. Therefore the current green gate demonstrates unbudgeted engine latency, but the production-default request represented by these cases still rejects.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/CypherExecutionBudget.kt
```diff
@@ -2,6 +2,9 @@ package io.johnsonlee.graphite.cypher
 
 import io.johnsonlee.graphite.graph.GraphWorkConsumer
 
+/** Default per-request work limit used by the production Cypher endpoint. */
+const val DEFAULT_CYPHER_WORK_BUDGET = 25_000_000L
```
johnsonlee: **[P1] Do not raise the production resource limit 100x to make the benchmark pass.** The 250,000 default was introduced by #95 to bound broad-query CPU/heap, and the server admits two such queries concurrently. Raising it to 25,000,000 changes product behavior and permits up to 50M inspected/materialized units at once; the real-fixture JMH runs with `-Xmx8g` and its report does not gate allocation (the prior unbudgeted run allocated roughly 0.85–10.26 GB/op), so it is not evidence that this new default remains safe under the production concurrency/4 GiB constraints. This PR is supposed to make the latency gate reject failed samples, not weaken the guard being tested. Please keep the production default unchanged and give the benchmark an explicit sufficiently large `CypherExecutionBudget` so it still exercises `workTrackingEnabled=true`; if the intended product requirement is that these broad queries succeed by default, then the query/work accounting needs optimization or this policy change needs dedicated concurrent CPU/heap evidence and regression coverage.

.github/workflows/benchmark.yml
johnsonlee: Verified in `cc349b9`: the permanent v1 `restore-keys` fallback is removed, so a changed v2 content hash now forces fixture regeneration. This issue is resolved.

graphite-webgraph/src/jmh/kotlin/io/johnsonlee/graphite/webgraph/MappedStringAdmissionBenchmark.kt
```diff
@@ -48,6 +48,9 @@ open class MappedStringAdmissionBenchmark {
     fun earlyHitAfterLargeLimitScan(state: LargeLimitBenchmarkState): CypherResult = state.earlyHitWithoutIndexBuild()
 
     @Benchmark
+    @BenchmarkMode(Mode.AverageTime)
```
johnsonlee: **[P1] Preserve the SingleShot gate instead of changing the measured workload after it fails.** This job is documented in `docs/benchmark-regression-gate.md` as a SingleShot guard for the budget-aware full scan, and the failed run repeated the slowdown in reverse order (4.041 -> 6.283 ms/op, +55.5%). Overriding it to `AverageTime` with multi-second warmup changes the contract to JIT-warmed steady-state throughput and can make a first-query/full-scan regression disappear; it also leaves the gate documentation false. Please keep `SingleShotTime` and investigate/fix the repeated candidate slowdown. If the concern is variance, stabilize it without changing the workload (for example more SingleShot forks/iterations and evidence showing the original result was noise), then update the documented methodology deliberately.

johnsonlee: The latest PASS (`1.289 -> 1.182 ms/op`) was produced only after this method changed from SingleShot to warmed AverageTime, so it cannot invalidate the repeated SingleShot regression that triggered the change. Please restore the documented SingleShot workload and rerun/stabilize that workload itself.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/CypherExecutionBudget.kt
johnsonlee: The latest green run `33233611874` does not supply the missing production-safety evidence. Its 4 GiB heap rows are single-run, informational measurements for the separate large-corpus pipeline; the wrapped-query JMH still runs with `-Xmx8g`, does not gate allocation, and never exercises two concurrent 25M-budget requests. Please restore the 250,000 production default and keep the larger budget benchmark-local.

graphite-webgraph/src/jmh/kotlin/io/johnsonlee/graphite/webgraph/MappedStringAdmissionBenchmark.kt
johnsonlee: Verified fixed by `063a5f8`: the method again inherits `Mode.SingleShotTime`, now stabilized with 3 forks, 5 warmup invocations, and 20 measured invocations per fork; the methodology docs match. Latest CI measured fixed baseline 3.963 ms/op vs PR 3.966 ms/op (+0.1%, PASS), and an independent local run reported SingleShot mode with 60 samples.

graphite-webgraph/src/jmh/kotlin/io/johnsonlee/graphite/webgraph/AllFixtureWrappedDiscoveryLatencyBenchmark.kt
johnsonlee: Verified fixed at final head `063a5f8`: `budgetedLatencyExecutor` now constructs an explicit 25,000,000-unit `CypherExecutionBudget`, so current base and candidate both exercise `workTrackingEnabled=true`; only the fixed pre-PR-95 revision explicitly opts into the legacy unbudgeted constructor. Run `33234406373` passed result preflight and every synthetic/real latency shard.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/CypherExecutionBudget.kt
johnsonlee: Verified fixed by `0084a61` at final head `063a5f8`: the production/CLI/README default is restored to 250,000, the public 25M production constant is removed, and 25M exists only as a private JMH budget. The final net diff against main contains no production-code or CLI behavior change; local uncached tests and latest CI pass.

PR #100, opened 2026-08-29
Replace legacy Explorer node and call-site discovery with Cypher and refresh the web UI

johnsonlee:
## Summary

- replace the legacy node and call-site REST endpoints and MCP tools with Cypher-based discovery
- retain `/api/methods` and the MCP `methods` tool because declared return types and indexed methods without graph nodes cannot be represented with equivalent Cypher node queries
- redesign Graphite Explorer as a responsive Cypher-first workspace with informative graph cards and a centered, balanced multi-graph grid
- keep single-graph sessions focused on the class overview, hide unnecessary topology navigation, and make the inspector span the tablet layout
- add recoverable canvas requests, zero-graph handling, and latest-intent-wins cancellation for all canvas actions
- update OpenAPI, documentation, tests, and CI coverage for the supported API and browser state transitions

## Breaking change

The root and graph-scoped `/nodes` and `/call-sites` routes are removed. The matching MCP `nodes` and `call_sites` tools are also removed. Use the Cypher endpoint or MCP `cypher` tool for node and call-site discovery.

The `/methods` routes and MCP `methods` tool remain supported to preserve declared method metadata that Cypher node discovery cannot reproduce.

## Verification

- `./gradlew check -S --no-daemon --no-build-cache` — passed in a normal clone, including coverage gates and Android, Hive, Kotlin compiler, and Tika large-corpus tests
- `node --test graphite-explore/src/test/js/ui-state.test.js` — passed (7 state and concurrency tests)
- `npm ci && npm run build` in `graphite-mcp` — passed
- browser verification: empty registry reports 0 graphs and clears loading state
- browser verification: single graph hides navigation and the inspector spans the full 800 px tablet grid
- browser verification: a hot-unloaded graph produces a retryable canvas error with no unhandled console errors
- browser verification: four graphs render as a centered 2×2 grid with graph id and core metrics

## Performance

The Explorer benchmark workload is unchanged from `main`; `/api/methods?limit=200` was restored to both the measured session and warmup loop so the benchmark remains directly comparable.

- method-level same-runner regression comparison passed on the latest head
- paired Explorer HTTP-session benchmark passed with identical base and candidate workloads and guardrails
- the complete `benchmark-regression-gate`, including large-corpus, resource, budgeted latency, and real/synthetic wrapped-query shards, passed on the latest head


README.md
```diff
@@ -259,9 +258,10 @@ one graph. Every root non-Cypher result is grouped by `graphId`, while every
 cross-graph Cypher row includes `$metadata.graphIds` and returned graph elements include
 qualified identities such as `elementId = "orders:42"`.
 
-For agent-driven discovery, probe `/openapi.json` first. It describes the full
-root-all and graph-scoped REST surface, including the two explicit modes of
-`/api/cypher/graphs`.
+Use Cypher for agent-driven node, method, and call-site discovery. The legacy
```
johnsonlee: [P1] Preserve method-discovery parity before removing the old surface

Cypher can expose `ReturnNode.method` signatures for methods that have return nodes, but it does not expose the declared `MethodDescriptor.returnType`; `properties(n)` contains only `id`, `method`, and optional `actual_type`. The deleted `/api/methods` response and MCP `methods` tool returned structured `signature`, `class`, `name`, and `returnType`. The existing synthetic graph fixture also contains an indexed `quxMethod` with no corresponding node, which the old endpoint could discover but `MATCH (n)` cannot enumerate. This makes the documented replacement functionally incomplete. Please retain the endpoint/tool or expose equivalent method metadata through Cypher, and add a parity test that covers an indexed method with no node plus its declared return type.

graphite-explore/src/main/resources/web/app.js
```diff
@@ -77,22 +93,39 @@ function initCytoscape() {
...
+        dashboardInfo = info;
+
+        const totals = info.totals || info;
+        setMetric('stat-nodes', totals.nodes);
+        setMetric('stat-edges', totals.edges);
+        setMetric('stat-methods', totals.methods);
+        setMetric('stat-callsites', totals.callSites);
+
+        const graphs = info.graphs || [];
+        const graphCount = Number(info.count || graphs.length || 1);
```
johnsonlee: [P2] Preserve a valid zero graph count

Empty startup is supported and `/api/graphs` correctly returns `count: 0, graphs: []`. Because both values are falsy, this expression reports one graph. I reproduced it against the PR server: the badge showed “1 graph” while every metric was zero. Use `info.count ?? graphs.length` (or an explicit presence check) so zero remains zero, and cover the empty-registry state.

graphite-explore/src/main/resources/web/app.js
```diff
@@ -130,51 +172,25 @@ async function loadGraphTopology() {
...
     renderGraph(data, null);
-    document.getElementById('graph-info').textContent = `${data.nodes.length} nodes, ${data.edges.length} edges`;
+    setGraphInfo(`${data.nodes.length} classes, ${data.edges.length} calls`);
 }
 
 async function loadSingleGraphOverview(graphId) {
-    const res = await fetch('/api/graphs/' + encodeURIComponent(graphId) + '/overview');
-    const data = await res.json();
+    setCanvasState('Loading ' + graphId);
+    const data = await fetchJson('/api/graphs/' + encodeURIComponent(graphId) + '/overview');
```
johnsonlee: [P1] Turn interactive fetch failures into a recoverable canvas state

This loading overlay is never cleared or replaced when `fetchJson` rejects, and the direct click/double-tap callers do not catch the returned promise. I reproduced this by selecting a graph, unloading it through `DELETE /api/graphs/{id}`, and clicking “Explore classes”: the UI remained permanently on “Loading one” and logged an unhandled `Error: Graph not loaded`. `loadSubgraph` has the same pattern. Catch these interactive requests through a shared wrapper, call `setCanvasError` with a retry or dashboard-refresh path, and test hot-unload/network failure.

graphite-explore/src/main/resources/web/app.js
```diff
@@ -557,7 +639,12 @@ async function runCypher() {
             loadCypherResults(nodeRefs);
```
johnsonlee: [P1] Prevent stale query neighborhoods from replacing newer results

`loadCypherResults` is launched without awaiting, cancellation, or a generation token, yet it performs up to 50 sequential requests before rendering. The `finally` block immediately re-enables Run, so query A can still be loading when query B completes; A can then overwrite B's canvas and graph summary. Abort or tag each query/view load and discard stale renders. Simply awaiting would lock the UI for as many as 50 requests, so please add a delayed-response two-query regression test as well.

graphite-explore/src/main/resources/web/style.css
```diff
@@ -1,199 +1,327 @@
...
+    #cypher-panel { padding: 14px 18px 16px; }
+    .canvas-tip { display: none; }
+}
+
+@media (max-width: 820px) {
+    body { overflow: auto; }
+    .workspace { display: flex; flex: none; flex-direction: column; min-height: auto; overflow: visible; }
+    .main-stage { order: -1; display: flex; flex: none; flex-direction: column; width: 100%; overflow: visible; }
+    #graph-area { height: 58dvh; min-height: 420px; }
+    #sidebar { display: grid; flex: none; grid-template-columns: 1fr 1fr; width: 100%; overflow: visible; border-top: 1px solid var(--border-subtle); border-right: 0; }
```
johnsonlee, line 302: [P2] Let the single-graph inspector use the full tablet width

At 561–820 px this creates a two-column sidebar, while single-graph mode hides `#navigation-section`. The remaining `#detail-section` auto-places in column 1, leaving the entire right half blank until the 560 px breakpoint switches back to block layout. Make the detail section span both columns when navigation is hidden (or define the grid based on mode) and cover the single-graph tablet layout.

johnsonlee (commented):
Blocking review findings on the locked head `a290610`.

GitHub does not allow the PR author to submit `REQUEST_CHANGES` on their own PR, so this review is posted as comments and the PR must remain unmerged until all blocking threads are fixed and re-verified.

The no-cache Explorer tests/detekt and MCP build pass, but they do not exercise the replacement discovery semantics or these browser state transitions. I reproduced the empty-registry count and hot-unload loading failure against the PR server. Please also rename the PR so the breaking removal of the legacy discovery endpoints/tools is visible in the squash commit title (for example: “Replace legacy Explorer discovery endpoints with Cypher and refresh the web UI”).

graphite-explore/src/jmh/kotlin/io/johnsonlee/graphite/cli/ExplorerMemoryBenchmark.kt
```diff
@@ -102,8 +102,7 @@ open class ExplorerMemoryBenchmark {
     fun android_initialExplorerSession(counters: ExplorerMemoryCounters): Long =
         measureRetainedHeap(counters) {
             request("/api/graphs") +
-                request("/api/overview?limit=200") +
-                request("/api/methods?limit=200")
+                request("/api/overview?limit=200")
```
johnsonlee: [P1] Benchmark the replacement discovery work instead of deleting it

This changes `android_initialExplorerSession` from three requests to two, and the waterline setup removes the same `/api/methods` request without issuing an equivalent Cypher method-discovery query. Any before/after memory or latency result is therefore incomparable: the candidate simply performs less work. Replace the removed request with the documented Cypher workflow and preserve equivalent result volume/semantics. Also wire this Explorer benchmark into a paired base/head CI gate; the current benchmark workflow only runs the Cypher and webgraph JMH suites, so this modified workload is not exercised by PR checks.

johnsonlee (commented):
Additional benchmark-integrity blocker found while reviewing the performance evidence.

README.md
johnsonlee: Fixed in `b5d2c97`. I restored the root and graph-scoped `/api/methods` routes, their OpenAPI entries, and the MCP `methods` tool. The new parity regression test queries `com.example.Child.qux`, which is added to the method index without a corresponding node, and asserts its declared `returnType` is `int`. Node and call-site discovery remain migrated to Cypher.

graphite-explore/src/main/resources/web/app.js
johnsonlee: Fixed in `b5d2c97`. Interactive canvas loads now go through the shared `runCanvasAction` recovery path, including node taps, graph/node double-taps, reset, subgraph links, and retry buttons. A rejected request is converted into the retryable canvas error state without an unhandled rejection. I also reproduced a hot-unloaded graph in the browser: it showed `Graph not loaded: hot-unloaded`, exposed `Try again`, and logged no console error.

graphite-explore/src/jmh/kotlin/io/johnsonlee/graphite/cli/ExplorerMemoryBenchmark.kt
johnsonlee: Fixed in `b5d2c97`. Because `/api/methods` is retained, I restored `/api/methods?limit=200` in both the measured Explorer session and the warmup loop. `ExplorerMemoryBenchmark.kt` is now identical to `main`, so the paired benchmark remains directly comparable. The full no-build-cache check, including all large-corpus performance gates, passed locally; the refreshed same-runner benchmark gate is now running on this head.

graphite-explore/src/main/resources/web/app.js
johnsonlee: Fixed in `b5d2c97`. Cypher neighborhood rendering now uses an AbortController plus a monotonically increasing generation in `createLatestTaskRunner`; starting query B aborts and invalidates query A, and only the current generation may render. The delayed two-query regression deliberately lets A resolve after B and verifies that only B is applied, even when the simulated A task ignores cancellation until completion.

graphite-explore/src/main/resources/web/style.css
johnsonlee: Fixed in `b5d2c97`. At the tablet breakpoint, `#navigation-section[hidden] + #detail-section` spans `grid-column: 1 / -1`. Browser verification at an 800 px viewport measured the single-graph inspector at the same full width as the sidebar.

graphite-explore/src/main/resources/web/app.js
johnsonlee: Fixed in `b5d2c97`. The dashboard now uses nullish graph-count semantics through `GraphiteUiState.graphCount`, so an explicit `count: 0` remains zero. The new browser-state test covers both `{count: 0, graphs: []}` and an omitted count with an empty graph list; actual browser verification also showed `0 graphs` and a cleared loading state.

graphite-explore/src/main/resources/web/app.js
johnsonlee: Reopening after re-verification: one interactive fetch path is still outside the shared recovery wrapper. `showNodeDetail` renders `onclick="loadIncomingEdges(...)"`, and `loadIncomingEdges` sets the inspector to `Loading...` before awaiting `fetchJson` without a catch. If the graph is hot-unloaded at that point (or the network fails), this still produces an unhandled rejection and leaves the inspector permanently loading. The pure `runRecoverable` unit test does not execute this wiring. Please route this action through `runCanvasAction` and cover the actual incoming-edge action.

graphite-explore/src/main/resources/web/app.js
johnsonlee: Reopening after re-verification: the runner now protects query A from query B, but it is not invalidated by a newer non-Cypher canvas action. `exploreGraph`, `exploreSubgraph`, reset, and dashboard loads use `runCanvasAction` without cancelling `cypherNeighborhoodRunner`; an in-flight query neighborhood can therefore finish later and overwrite the graph/subgraph view the user just selected. The current test only exercises runner A followed by runner B. Please make every newer canvas-view action invalidate the active neighborhood generation and add the delayed query -> non-query view regression.

graphite-explore/src/jmh/kotlin/io/johnsonlee/graphite/cli/ExplorerMemoryBenchmark.kt
johnsonlee: Reopening after checking the workflow itself: restoring the original three-request workload makes `ExplorerMemoryBenchmark` comparable again, but no workflow references or runs this benchmark. The active paired gate only covers the Cypher/WebGraph jobs, and the `build.yml` increment adds only the Node UI-state test. Therefore an Explorer memory/latency regression still cannot fail `benchmark-regression-gate`; the running large-corpus checks cited above are not evidence for this workload. Please add an actual same-runner base/head Explorer benchmark job and include it in the aggregate gate.

graphite-explore/src/main/resources/web/app.js
johnsonlee: Fixed in `14cdfb9`. The generated `Load incoming` control now calls `exploreIncomingEdges`, which routes the request through the same interrupting/recoverable canvas action runner as the other interactive fetches. A rejection is consumed, rendered as a retryable error, and the retry retains the original action. The updated state test asserts that a failed canvas action returns without rejection and reports that exact action for retry.

graphite-explore/src/main/resources/web/app.js
johnsonlee: Fixed in `14cdfb9`. `runCanvasAction` is now created by `createInterruptingActionRunner`, which cancels/increments `cypherNeighborhoodRunner` before every newer non-Cypher canvas action. The new delayed regression starts an old query neighborhood, applies a newer graph view, then releases the old task and verifies it is discarded instead of overwriting the view.

graphite-explore/src/jmh/kotlin/io/johnsonlee/graphite/cli/ExplorerMemoryBenchmark.kt
johnsonlee: Fixed in `14cdfb9`. The workflow now has a separate `explorer-benchmarks` job on the cached persisted Android graph. It copies the base-defined Explorer workload into the candidate, builds both JMH jars, runs all four `ExplorerMemoryBenchmark` sessions as three cold forks in base→PR order, and reverse-confirms PR→base when any latency row exceeds 15%. The job validates the exact four-result set, uploads its report/status, is required by the aggregate job, and the aggregate fails closed if the Explorer artifact is absent. Locally the full real-Android four-session JMH completed successfully in 1m52s with the 4 GiB fork and all built-in waterline checks passing; the workflow/cache wiring and 37 JS gate/state tests also pass.

graphite-explore/src/main/resources/web/app.js
johnsonlee: Follow-up `14cdfb9` closes the remaining cross-view race: every non-query canvas action now cancels the active neighborhood generation before loading its own view. A second regression test starts a delayed query, applies a manual graph view, then resolves the old query and verifies the manual view remains.

graphite-explore/src/jmh/kotlin/io/johnsonlee/graphite/cli/ExplorerMemoryBenchmark.kt
johnsonlee: Follow-up `14cdfb9` also adds a dedicated paired `explorer-benchmarks` CI job. It builds base and candidate JMH jars with the identical base Explorer harness, runs both revisions on the same Android fixture and runner, compares all four session benchmarks, uploads the report/status, and makes the final aggregate gate fail closed if the Explorer artifact or job is missing.

graphite-explore/src/main/resources/web/app.js
johnsonlee: Follow-up in `42f1784`: incoming-edge loads now participate in the unified latest-intent controller. Each request has an ownership token, so cancellation restores the idle control only when it still owns that block (an older A cannot roll back a newer B). Current failures remain retryable; successful retry clears the current error overlay. This also prevents an aborted incoming request from leaving a permanent local `Loading...` state.

graphite-explore/src/main/resources/web/app.js
johnsonlee: Follow-up in `42f1784`: re-verification found that cancelling only Cypher neighborhoods was insufficient for graph A → graph B and async-detail → sync-detail races. All canvas/detail intents now share one generation plus AbortController. Post-await DOM/canvas mutations and layout-stop callbacks commit only while their intent is current; synchronous graph/topology/sidebar selections also invalidate older work; query neighborhoods start only if the originating query intent is still current. Seven state regressions cover delayed query replacement, query→view, ordinary action stale render/error suppression, retry clearing, and aborted-loading ownership.

graphite-explore/src/main/resources/web/app.js
johnsonlee: Verified on exact head `42f17847d76d228470a547214fb325d2eafb7bf5`: the seven UI-state regressions and full unit workflow passed, including retry/error ownership and stale-action suppression. Resolving.

graphite-explore/src/main/resources/web/app.js
johnsonlee: Verified on exact head `42f17847d76d228470a547214fb325d2eafb7bf5`: all latest-intent race regressions passed and the full unit workflow completed successfully. Resolving.

graphite-explore/src/jmh/kotlin/io/johnsonlee/graphite/cli/ExplorerMemoryBenchmark.kt
johnsonlee: Verified on exact head `42f17847d76d228470a547214fb325d2eafb7bf5`. The new Explorer gate passed on the cached Android graph: browser-forward 705.060→698.141 ms/op (-1.0%), initial session 722.303→725.953 (+0.5%), long-running waterline 1630.792→1655.571 (+1.5%), and incoming waterline 2820.327→1880.340 (-33.3%). `explorer-status.json` is PASS with no errors, and the final aggregate `benchmark-regression-gate` also passed. Resolving.

PR #101, opened 2026-08-29
Route method discovery through Cypher

johnsonlee:
## Summary

- remove root and graph-scoped `/methods` REST routes from Explorer and OpenAPI
- remove the MCP `methods` tool so method discovery is unified on Cypher
- add a virtual `Method` Cypher source backed by the complete metadata index, including indexed-only methods
- preserve general Cypher semantics while pushing exact, prefix, suffix, contains, regex, parameter, and return-type filters into a bounded fast path
- stream MAPPED method metadata without initializing the full metadata object
- use NCPU fork-join scans for bounded small and medium multi-graph queries; large result and grouped-count requests use a single streaming accumulator to bound CPU and RSS
- keep Method metadata scans outside the graph-work counter so complete 4/17/36-graph discovery does not become a false HTTP 429
- raise and centralize defaults at 4 concurrent Cypher queries and 1,000,000 graph work units
- add paired base-vs-head Method gates over real 4/17/36 mapped graphs, exact result manifests, wall time, CPU, final RSS, capacity, cancellation, and recovery
- reconstruct aggregate benchmark reports from only the latest attempt of each artifact producer, without mixing partial retries with stale files

## Verification

- `./gradlew check koverLog --no-daemon` — PASS, 84 tasks
- coverage: core 98.2047%, cypher 98.0828%, explore 98.0121%, query 98.9011%, sootup 98.4449%, webgraph 98.0596%
- `node --test .github/scripts/benchmark-gate.test.mjs` — 39/39, including the partial-retry stale-file regression
- `actionlint .github/workflows/benchmark.yml` — PASS
- `-Xmx256m` 8×100k grouped Method count — PASS
- both Method coordination tests forced through 5/5 reruns — PASS
- exact-head [unit run](https://github.com/johnsonlee/graphite/actions/runs/33279992585) for `b2569c46f9b1` — PASS
- exact-head [benchmark run](https://github.com/johnsonlee/graphite/actions/runs/33279992621) — PASS
- exact-head [benchmark report](https://github.com/johnsonlee/graphite/pull/101#issuecomment-5462269909) — PASS

## Benchmark conclusion

- Method compatibility: base and PR manifests are identical for 33/33 cases; all wall-time, CPU, and final-RSS rows pass the 15% gate
- Cypher capacity: -0.8% versus base, below the 15% limit; overload rejection, budget rejection, cancellation, and recovery pass at defaults 4 / 1,000,000
[...]

PR #102, opened 2026-08-29
Bound memory for filtered DATAFLOW queries

johnsonlee:
## Summary

- stream filtered relationship matches through WHERE, projection, DISTINCT, ORDER BY, SKIP, and LIMIT instead of retaining every intermediate binding
- keep ORDER BY memory bounded to O(SKIP + LIMIT) with a stable top-k heap after WHERE filtering, including hidden sort expressions omitted from RETURN
- prune unrelated graph sources when WHERE contains an exact source graphId equality
- keep fixed-hop and variable-length adjacency traversal lazy, and expand later patterns from their declared source variable
- traverse variable-length paths depth-first with an O(maxDepth) retained frame stack and a fixed-capacity dead-suffix memo that caches only zero-result, cycle-independent suffixes
- apply the target node pattern to variable-length endpoints even without WHERE, and additionally push the final WHERE only when all required relationship bindings are available
- preserve relationship-path multiplicity at reconvergence and preserve cycle-dependent paths under different ancestor contexts
- preserve the public `PathFinder.findPaths` breadth-first/shortest-first contract while keeping Cypher on its separate bounded lazy traversal
- apply relationship type unions, aliases, invalid-type rejection, and inline relationship properties to every variable-length hop
- support explicit aliases such as DATA_FLOW/DATAFLOW and CONTROL_FLOW/CONTROLFLOW; malformed underscore spellings remain unknown

## Verification

Exact candidate: `ab80322`.

- `./gradlew :cypher:test :cypher:detekt :cypher:filteredRelationshipMemoryTest :cypher:koverVerify :cypher:koverLog :cypher:jmhJar --no-daemon --rerun-tasks`
  - 1,054 regular tests and 7 isolated heap/budget tests, 0 failures
  - detekt clean
  - 98.0914% Cypher line coverage
  - JMH jar built successfully
- `git diff --check`
- GitHub unit/coverage workflow: PASS
- [Required benchmark-regression gate: PASS on base `b5808cd` and PR `ab80322`](https://github.com/johnsonlee/graphite/pull/102#issuecomment-5462688519)

The isolated `FilteredRelationshipMemoryTest` always runs with `-Xmx256m`. It covers:

- 4,000,000 generated single-hop DATAFLOW edges with zero WHERE hits
- 10,000,000 distinct generated variable-path targets with zero WHERE hits
- a 65-node/456-edge, 8-layer × 8-wide reconvergent DAG under the normal 250,000-work budget
- the same reconvergent DAG with no WHERE and an absent target label
- the same reconvergent DAG with no WHERE and an absent inline target property
- a 3,000-node chain whose terminal points back to every ancestor, preventing quadratic retained cycle-blocker state

Correctness controls cover DATAFLOW aliases/types, lazy LIMIT behavior, exact-depth revisits, cyclic ancestor-context changes, bounded memo eviction, public BFS order, and a diamond that must return two relationship-path rows without DISTINCT. A relationship-variable regression also verifies that WHERE is evaluated only after the path binding exists.

## Benchmark conclusions

The required gate runs base and PR on the same GitHub runners and publishes commands, environment, confidence handling, raw reports, and artifacts in the linked comment/run.

- Method-level: PASS. `CypherBenchmark.filteredVariableLengthPath` improved from 360.206 to 348.807 us/op (-3.2%); `filteredSingleHopRelationship` improved by 65.9%. No method-level row produced a confirmed blocking regression.
- End-to-end: PASS. All real-corpus JAR → build → save → mapped-load → Cypher rows passed. The 36-real-graph wrapped zero-hit query improved 4.336 → 4.188 s/op (-3.4%) versus current base and retained the required fixed-baseline speedup.
- Resources: PASS without threshold changes. On the 36-real-graph probe, allocation was +2.1%, query GC count +8.3%, query GC time +13.3%, retained heap delta -0.7%, and peak used heap -0.1%; all were inside the existing gate.

### DIFF
--- a/graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/PathFinder.kt
+++ b/graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/PathFinder.kt
@@ -140,6 +140,15 @@ object PathFinder {
         val start = SearchState(startNode, incomingEdge = null, parent = null, depth = 0)
         if (matchesTarget(start, options)) yield(PathMatch(start))
         if (options.maxDepth <= 0) return@sequence
+        if (options.maxDepth == 1) {
+            for (edge in edgesForDirection(graph, startNode.id, options)) {
+                val nextId = nextNodeId(edge, startNode.id, options.direction)
+                val nextNode = loadNode(graph, nextId, options.workTracker) ?: continue
+                val nextState = SearchState(nextNode, edge, start, depth = 1)
+                if (matchesTarget(nextState, options)) yield(PathMatch(nextState))
+            }
+            return@sequence
+        }
 
         val stack = ArrayDeque<SearchFrame>()
         val startEdges = edgesForDirection(graph, startNode.id, options).iterator()
@@ -166,9 +175,16 @@ object PathFinder {
             val nextState = SearchState(nextNode, edge, frame.state, frame.state.depth + 1)
             if (matchesTarget(nextState, options)) yield(PathMatch(nextState))
 
-            val expansion = expandableFrame(graph, nextId, nextState, options, expansionMemo)
-            frame.blockedAncestors.addAll(expansion.blockedAncestors)
-            expansion.frame?.let(stack::addLast)
+            if (nextState.depth < options.maxDepth) {
+                expandableFrame(
+                    graph,
+                    nextId,
+                    nextState,
+                    options,
+                    expansionMemo,
+                    frame.blockedAncestors
+                )?.let(stack::addLast)
+            }
         }
     }
 
@@ -178,23 +194,25 @@ object PathFinder {
         nodeId: NodeId,
         state: SearchState,
         options: SearchOptions,
-        expansionMemo: LinkedHashMap<Int, ExpansionRecord>
-    ): ExpansionDecision {
-        if (state.depth >= options.maxDepth) return ExpansionDecision()
+        expansionMemo: LinkedHashMap<Int, ExpansionRecord>,
+        blockedAncestors: MutableSet<Int>
+    ): SearchFrame? {
         if (hasAncestor(state.parent, nodeId)) {
-            return ExpansionDecision(blockedAncestors = setOf(nodeId.value))
+            blockedAncestors.add(nodeId.value)
+            return null
         }
 
         val edges = edgesForDirection(graph, nodeId, options).iterator()
-        if (!edges.hasNext()) return ExpansionDecision()
+        if (!edges.hasNext()) return null
         val previous = expansionMemo[nodeId.value]
         if (previous != null &&
             previous.depth == state.depth &&
             ancestorIds(state.parent).containsAll(previous.blockedAncestors)
         ) {
-            return ExpansionDecision(blockedAncestors = previous.blockedAncestors)
+            blockedAncestors.addAll(previous.blockedAncestors)
+            return null
         }
-        return ExpansionDecision(frame = SearchFrame(state, edges))
+        return SearchFrame(state, edges)
     }
 
     private fun ancestorIds(state: SearchState?): Set<Int> {
@@ -231,11 +249,6 @@ object PathFinder {
         val state: SearchState,
         val edges: Iterator<Edge>,
         val blockedAncestors: MutableSet<Int> = mutableSetOf()
-    )
-
-    private data class ExpansionDecision(
-        val frame: SearchFrame? = null,
-        val blockedAncestors: Set<Int> = emptySet()
     )
 
     private data class ExpansionRecord(
--- a/graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/PathFinder.kt
+++ b/graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/PathFinder.kt
@@ -51,7 +51,7 @@ object PathFinder {
     }
 
     /**
-     * Find paths from source nodes to target nodes with a depth-bounded traversal.
+     * Find all paths from source nodes to target nodes via breadth-first search.
      *
      * @param graph The graph to search
      * @param sources Source node IDs to start from
@@ -70,11 +70,13 @@ object PathFinder {
         minDepth: Int = 1,
         maxDepth: Int = 10,
         direction: Direction = Direction.OUTGOING
-    ): List<Path> = findPathMatches(
-        graph,
-        sources,
-        SearchOptions(targets, edgeType, minDepth, maxDepth, direction)
-    ).map { it.materialize(workTracker = null) }.toList()
+    ): List<Path> {
+        val options = SearchOptions(targets, edgeType, minDepth, maxDepth, direction)
+        return sources.asSequence().flatMap { source ->
+            val startNode = loadNode(graph, source, workTracker = null)
+            if (startNode == null) emptySequence() else breadthFirst(graph, startNode, options)
+        }.map { it.materialize(workTracker = null) }.toList()
+    }
 
     internal fun findPathMatches(
         graph: Graph,
@@ -87,11 +89,39 @@ object PathFinder {
         }
     }
 
+    /** Preserve the public [findPaths] shortest-first contract. */
+    private fun breadthFirst(
+        graph: Graph,
+        startNode: Node,
+        options: SearchOptions
+    ): Sequence<PathMatch> = sequence {
+        val visited = mutableSetOf<Int>()
+        val queue = ArrayDeque<SearchState>()
+        queue.add(SearchState(startNode, incomingEdge = null, parent = null, depth = 0))
+
+        while (queue.isNotEmpty()) {
+            val state = queue.removeFirst()
+            val current = state.node.id
+
+            if (state.depth == 0 && matchesTarget(state, options)) yield(PathMatch(state))
+            if (state.depth >= options.maxDepth) continue
+            if (!visited.add(current.value)) continue
+
+            for (edge in edgesForDirection(graph, current, options)) {
+                val nextId = nextNodeId(edge, current, options.direction)
+                val nextNode = loadNode(graph, nextId, options.workTracker) ?: continue
+                val nextState = SearchState(nextNode, edge, state, state.depth + 1)
+                if (matchesTarget(nextState, options)) yield(PathMatch(nextState))
+                queue.add(nextState)
+            }
+        }
+    }
+
     /**
      * Traverse one branch at a time so suspended lazy consumers retain at most [SearchOptions.maxDepth]
      * search states instead of a complete breadth-first frontier.
-     * The shallowest-depth map grows with expanded node IDs, but does not retain nodes or parent paths;
-     * only the stack's O(maxDepth) frames retain those objects.
+     * Nodes already present on the current branch are not expanded again, so cycles terminate without
+     * retaining a traversal-wide visited set. Only the stack's O(maxDepth) frames retain search state.
      *
      * Matches keep the graph's edge order within each branch. Cypher does not define a global result
      * order without ORDER BY, so depth-first traversal is safe for the lazy query pipeline while allowing
@@ -102,11 +132,9 @@ object PathFinder {
         startNode: Node,
         options: SearchOptions
     ): Sequence<PathMatch> = sequence {
-        val expandedAtDepth = mutableMapOf<Int, Int>()
         val start = SearchState(startNode, incomingEdge = null, parent = null, depth = 0)
         if (matchesTarget(start, options)) yield(PathMatch(start))
         if (options.maxDepth <= 0) return@sequence
-        expandedAtDepth[startNode.id.value] = 0
 
         val stack = ArrayDeque<SearchFrame>()
         stack.addLast(SearchFrame(start, edgesForDirection(graph, startNode.id, options).iterator()))
@@ -125,14 +153,21 @@ object PathFinder {
             val nextState = SearchState(nextNode, edge, frame.state, frame.state.depth + 1)
             if (matchesTarget(nextState, options)) yield(PathMatch(nextState))
 
-            val previousDepth = expandedAtDepth[nextId.value]
             if (nextState.depth < options.maxDepth &&
-                (previousDepth == null || nextState.depth < previousDepth)
+                !hasAncestor(frame.state, nextId)
             ) {
-                expandedAtDepth[nextId.value] = nextState.depth
                 stack.addLast(SearchFrame(nextState, edgesForDirection(graph, nextId, options).iterator()))
             }
         }
+    }
+
+    private fun hasAncestor(state: SearchState?, nodeId: NodeId): Boolean {
+        var cursor = state
+        while (cursor != null) {
+            if (cursor.node.id == nodeId) return true
+            cursor = cursor.parent
+        }
+        return false
     }
 
     private data class SearchFrame(
--- a/graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/PathFinder.kt
+++ b/graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/PathFinder.kt
@@ -99,7 +99,7 @@ object PathFinder {
             val state = queue.removeFirst()
             val current = state.node.id
 
-            if (state.depth >= options.minDepth && (options.targets == null || current in options.targets)) {
+            if (state.depth == 0 && matchesTarget(state, options)) {
                 yield(PathMatch(state))
             }
 
@@ -110,12 +110,17 @@ object PathFinder {
             for (edge in edges) {
                 val nextId = nextNodeId(edge, current, options.direction)
                 val nextNode = loadNode(graph, nextId, options.workTracker) ?: continue
-                queue.add(SearchState(nextNode, edge, state, state.depth + 1))
+                val nextState = SearchState(nextNode, edge, state, state.depth + 1)
+                if (matchesTarget(nextState, options)) yield(PathMatch(nextState))
+                queue.add(nextState)
             }
         }
     }
 
-    private fun edgesForDirection(graph: Graph, nodeId: NodeId, options: SearchOptions): List<Edge> =
+    private fun matchesTarget(state: SearchState, options: SearchOptions): Boolean =
+        state.depth >= options.minDepth && (options.targets == null || state.node.id in options.targets)
+
+    private fun edgesForDirection(graph: Graph, nodeId: NodeId, options: SearchOptions): Sequence<Edge> =
         when (options.direction) {
             Direction.OUTGOING -> filteredEdges(graph.outgoing(nodeId), options.edgeType, options.workTracker)
             Direction.INCOMING -> filteredEdges(graph.incoming(nodeId), options.edgeType, options.workTracker)
@@ -138,13 +143,11 @@ object PathFinder {
         edges: Sequence<Edge>,
         edgeType: Class<out Edge>?,
         workTracker: CypherWorkTracker?
-    ): List<Edge> {
-        val result = mutableListOf<Edge>()
+    ): Sequence<Edge> = sequence {
         for (edge in edges) {
             workTracker?.consume()
-            if (edgeType == null || edgeType.isInstance(edge)) result.add(edge)
+            if (edgeType == null || edgeType.isInstance(edge)) yield(edge)
         }
-        return result
     }
 
     enum class Direction { OUTGOING, INCOMING, BOTH }
--- a/graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/PathFinder.kt
+++ b/graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/PathFinder.kt
@@ -19,7 +19,8 @@ object PathFinder {
         val maxDepth: Int,
         val direction: Direction,
         val workTracker: CypherWorkTracker? = null,
-        val edgePredicate: ((Edge) -> Boolean)? = null
+        val edgePredicate: ((Edge) -> Boolean)? = null,
+        val nodePredicate: ((Node) -> Boolean)? = null
     )
 
     internal class SearchState(
@@ -121,11 +122,11 @@ object PathFinder {
      * Traverse one branch at a time so suspended lazy consumers retain at most [SearchOptions.maxDepth]
      * search states instead of a complete breadth-first frontier.
      * Nodes already present on the current branch are not expanded again, so cycles terminate without
-     * retaining a traversal-wide visited set. A fixed-size memo suppresses reconvergent expansion while
-     * recording the ancestors that actually blocked each completed suffix. A revisit is suppressed only
-     * when its ancestor set still contains every prior blocker. This keeps acyclic reconvergence cheap
-     * without dropping a suffix that becomes legal under a different branch. Eviction can only cause safe
-     * re-expansion. Only the stack's O(maxDepth) frames retain nodes and parent paths.
+     * retaining a traversal-wide visited set. A fixed-size memo suppresses only suffixes that completed
+     * without producing a match or encountering an ancestor back-edge. Result-producing suffixes are
+     * always replayed for every prefix, preserving Cypher row multiplicity; cycle-dependent suffixes are
+     * never memoized, preserving paths that become legal under a different ancestor set. Eviction can only
+     * cause safe re-expansion. Only the stack's O(maxDepth) frames retain nodes and parent paths.
      *
      * Matches keep the graph's edge order within each branch. Cypher does not define a global result
      * order without ORDER BY, so depth-first traversal is safe for the lazy query pipeline while allowing
@@ -136,17 +137,12 @@ object PathFinder {
         startNode: Node,
         options: SearchOptions
     ): Sequence<PathMatch> = sequence {
-        val expansionMemo = LinkedHashMap<Int, ExpansionRecord>()
+        val deadSuffixMemo = LinkedHashMap<ExpansionKey, Unit>()
         val start = SearchState(startNode, incomingEdge = null, parent = null, depth = 0)
         if (matchesTarget(start, options)) yield(PathMatch(start))
         if (options.maxDepth <= 0) return@sequence
         if (options.maxDepth == 1) {
-            for (edge in edgesForDirection(graph, startNode.id, options)) {
-                val nextId = nextNodeId(edge, startNode.id, options.direction)
-                val nextNode = loadNode(graph, nextId, options.workTracker) ?: continue
-                val nextState = SearchState(nextNode, edge, start, depth = 1)
-                if (matchesTarget(nextState, options)) yield(PathMatch(nextState))
-            }
+            yieldAll(singleHop(graph, start, options))
             return@sequence
         }
 
@@ -159,12 +155,17 @@ object PathFinder {
             val frame = stack.last()
             if (!frame.edges.hasNext()) {
                 val completed = stack.removeLast()
-                rememberBounded(
-                    expansionMemo,
-                    completed.state.node.id.value,
-                    ExpansionRecord(completed.state.depth, completed.blockedAncestors.toSet())
-                )
-                stack.lastOrNull()?.blockedAncestors?.addAll(completed.blockedAncestors)
+                if (!completed.foundMatch && !completed.cycleBlocked) {
+                    rememberBounded(
+                        deadSuffixMemo,
+                        ExpansionKey(completed.state.node.id.value, completed.state.depth),
+                        Unit
+                    )
+                }
+                stack.lastOrNull()?.let { parent ->
+                    parent.foundMatch = parent.foundMatch || completed.foundMatch
+                    parent.cycleBlocked = parent.cycleBlocked || completed.cycleBlocked
+                }
                 continue
             }
 
@@ -173,7 +174,10 @@ object PathFinder {
             val nextId = nextNodeId(edge, current, options.direction)
             val nextNode = loadNode(graph, nextId, options.workTracker) ?: continue
             val nextState = SearchState(nextNode, edge, frame.state, frame.state.depth + 1)
-            if (matchesTarget(nextState, options)) yield(PathMatch(nextState))
+            if (matchesTarget(nextState, options)) {
+                frame.foundMatch = true
+                yield(PathMatch(nextState))
+            }
 
             if (nextState.depth < options.maxDepth) {
                 expandableFrame(
@@ -181,10 +185,23 @@ object PathFinder {
                     nextId,
                     nextState,
                     options,
-                    expansionMemo,
-                    frame.blockedAncestors
+                    deadSuffixMemo,
+                    frame
                 )?.let(stack::addLast)
             }
+        }
+    }
+
+    private fun singleHop(
+        graph: Graph,
+        start: SearchState,
+        options: SearchOptions
+    ): Sequence<PathMatch> = sequence {
+        for (edge in edgesForDirection(graph, start.node.id, options)) {
+            val nextId = nextNodeId(edge, start.node.id, options.direction)
+            val nextNode = loadNode(graph, nextId, options.workTracker) ?: continue
+            val nextState = SearchState(nextNode, edge, start, depth = 1)
+            if (matchesTarget(nextState, options)) yield(PathMatch(nextState))
         }
     }
 
@@ -194,39 +211,22 @@ object PathFinder {
         nodeId: NodeId,
         state: SearchState,
         options: SearchOptions,
-        expansionMemo: LinkedHashMap<Int, ExpansionRecord>,
-        blockedAncestors: MutableSet<Int>
+        deadSuffixMemo: LinkedHashMap<ExpansionKey, Unit>,
+        parentFrame: SearchFrame
     ): SearchFrame? {
         if (hasAncestor(state.parent, nodeId)) {
-            blockedAncestors.add(nodeId.value)
+            parentFrame.cycleBlocked = true
             return null
         }
 
+        if (ExpansionKey(nodeId.value, state.depth) in deadSuffixMemo) return null
         val edges = edgesForDirection(graph, nodeId, options).iterator()
         if (!edges.hasNext()) return null
-        val previous = expansionMemo[nodeId.value]
-        if (previous != null &&
-            previous.depth == state.depth &&
-            ancestorIds(state.parent).containsAll(previous.blockedAncestors)
-        ) {
-            blockedAncestors.addAll(previous.blockedAncestors)
-            return null
-        }
         return SearchFrame(state, edges)
     }
 
-    private fun ancestorIds(state: SearchState?): Set<Int> {
-        val ids = mutableSetOf<Int>()
-        var cursor = state
-        while (cursor != null) {
-            ids.add(cursor.node.id.value)
-            cursor = cursor.parent
-        }
-        return ids
-    }
-
     private fun <K, V> rememberBounded(memo: LinkedHashMap<K, V>, key: K, value: V) {
-        if (!memo.containsKey(key) && memo.size >= EXPANSION_MEMO_CAPACITY) {
+        if (!memo.containsKey(key) && memo.size >= DEAD_SUFFIX_MEMO_CAPACITY) {
             val eldest = memo.entries.iterator()
             if (eldest.hasNext()) {
                 eldest.next()
@@ -248,16 +248,19 @@ object PathFinder {
     private data class SearchFrame(
         val state: SearchState,
         val edges: Iterator<Edge>,
-        val blockedAncestors: MutableSet<Int> = mutableSetOf()
+        var foundMatch: Boolean = false,
+        var cycleBlocked: Boolean = false
     )
 
-    private data class ExpansionRecord(
-        val depth: Int,
-        val blockedAncestors: Set<Int>
+    private data class ExpansionKey(
+        val nodeId: Int,
+        val depth: Int
     )
 
     private fun matchesTarget(state: SearchState, options: SearchOptions): Boolean =
-        state.depth >= options.minDepth && (options.targets == null || state.node.id in options.targets)
+        state.depth >= options.minDepth &&
+            (options.targets == null || state.node.id in options.targets) &&
+            (options.nodePredicate?.invoke(state.node) != false)
 
     private fun edgesForDirection(graph: Graph, nodeId: NodeId, options: SearchOptions): Sequence<Edge> =
         when (options.direction) {
@@ -292,5 +295,5 @@ object PathFinder {
 
     enum class Direction { OUTGOING, INCOMING, BOTH }
 
-    private const val EXPANSION_MEMO_CAPACITY = 16_384
+    private const val DEAD_SUFFIX_MEMO_CAPACITY = 16_384
 }
--- a/graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/PathFinder.kt
+++ b/graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/PathFinder.kt
@@ -122,8 +122,10 @@ object PathFinder {
      * search states instead of a complete breadth-first frontier.
      * Nodes already present on the current branch are not expanded again, so cycles terminate without
      * retaining a traversal-wide visited set. A fixed-size memo suppresses reconvergent expansion while
-     * preserving the memory bound; eviction can only cause safe re-expansion. Only the stack's O(maxDepth)
-     * frames retain nodes and parent paths.
+     * recording the ancestors that actually blocked each completed suffix. A revisit is suppressed only
+     * when its ancestor set still contains every prior blocker. This keeps acyclic reconvergence cheap
+     * without dropping a suffix that becomes legal under a different branch. Eviction can only cause safe
+     * re-expansion. Only the stack's O(maxDepth) frames retain nodes and parent paths.
      *
      * Matches keep the graph's edge order within each branch. Cypher does not define a global result
      * order without ORDER BY, so depth-first traversal is safe for the lazy query pipeline while allowing
@@ -134,7 +136,7 @@ object PathFinder {
         startNode: Node,
         options: SearchOptions
     ): Sequence<PathMatch> = sequence {
-        val expandedAtDepth = LinkedHashMap<Int, Int>()
+        val expansionMemo = LinkedHashMap<Int, ExpansionRecord>()
         val start = SearchState(startNode, incomingEdge = null, parent = null, depth = 0)
         if (matchesTarget(start, options)) yield(PathMatch(start))
         if (options.maxDepth <= 0) return@sequence
@@ -142,13 +144,18 @@ object PathFinder {
         val stack = ArrayDeque<SearchFrame>()
         val startEdges = edgesForDirection(graph, startNode.id, options).iterator()
         if (!startEdges.hasNext()) return@sequence
-        rememberExpansion(expandedAtDepth, startNode.id, depth = 0)
         stack.addLast(SearchFrame(start, startEdges))
 
         while (stack.isNotEmpty()) {
             val frame = stack.last()
             if (!frame.edges.hasNext()) {
-                stack.removeLast()
+                val completed = stack.removeLast()
+                rememberBounded(
+                    expansionMemo,
+                    completed.state.node.id.value,
+                    ExpansionRecord(completed.state.depth, completed.blockedAncestors.toSet())
+                )
+                stack.lastOrNull()?.blockedAncestors?.addAll(completed.blockedAncestors)
                 continue
             }
 
@@ -159,7 +166,9 @@ object PathFinder {
             val nextState = SearchState(nextNode, edge, frame.state, frame.state.depth + 1)
             if (matchesTarget(nextState, options)) yield(PathMatch(nextState))
 
-            expandableFrame(graph, nextId, nextState, options, expandedAtDepth)?.let(stack::addLast)
+            val expansion = expandableFrame(graph, nextId, nextState, options, expansionMemo)
+            frame.blockedAncestors.addAll(expansion.blockedAncestors)
+            expansion.frame?.let(stack::addLast)
         }
     }
 
@@ -169,27 +178,44 @@ object PathFinder {
         nodeId: NodeId,
         state: SearchState,
         options: SearchOptions,
-        expandedAtDepth: LinkedHashMap<Int, Int>
-    ): SearchFrame? {
-        if (state.depth >= options.maxDepth || hasAncestor(state.parent, nodeId)) return null
-        val previousDepth = expandedAtDepth[nodeId.value]
-        if (previousDepth != null && previousDepth <= state.depth) return null
+        expansionMemo: LinkedHashMap<Int, ExpansionRecord>
+    ): ExpansionDecision {
+        if (state.depth >= options.maxDepth) return ExpansionDecision()
+        if (hasAncestor(state.parent, nodeId)) {
+            return ExpansionDecision(blockedAncestors = setOf(nodeId.value))
+        }
 
         val edges = edgesForDirection(graph, nodeId, options).iterator()
-        if (!edges.hasNext()) return null
-        rememberExpansion(expandedAtDepth, nodeId, state.depth)
-        return SearchFrame(state, edges)
+        if (!edges.hasNext()) return ExpansionDecision()
+        val previous = expansionMemo[nodeId.value]
+        if (previous != null &&
+            previous.depth == state.depth &&
+            ancestorIds(state.parent).containsAll(previous.blockedAncestors)
+        ) {
+            return ExpansionDecision(blockedAncestors = previous.blockedAncestors)
+        }
+        return ExpansionDecision(frame = SearchFrame(state, edges))
     }
 
-    private fun rememberExpansion(expandedAtDepth: LinkedHashMap<Int, Int>, nodeId: NodeId, depth: Int) {
-        if (!expandedAtDepth.containsKey(nodeId.value) && expandedAtDepth.size >= EXPANSION_MEMO_CAPACITY) {
-            val eldest = expandedAtDepth.entries.iterator()
+    private fun ancestorIds(state: SearchState?): Set<Int> {
+        val ids = mutableSetOf<Int>()
+        var cursor = state
+        while (cursor != null) {
+            ids.add(cursor.node.id.value)
+            cursor = cursor.parent
+        }
+        return ids
+    }
+
+    private fun <K, V> rememberBounded(memo: LinkedHashMap<K, V>, key: K, value: V) {
+        if (!memo.containsKey(key) && memo.size >= EXPANSION_MEMO_CAPACITY) {
+            val eldest = memo.entries.iterator()
             if (eldest.hasNext()) {
                 eldest.next()
                 eldest.remove()
             }
         }
-        expandedAtDepth[nodeId.value] = depth
+        memo[key] = value
     }
 
     private fun hasAncestor(state: SearchState?, nodeId: NodeId): Boolean {
@@ -203,7 +229,18 @@ object PathFinder {
 
     private data class SearchFrame(
         val state: SearchState,
-        val edges: Iterator<Edge>
+        val edges: Iterator<Edge>,
+        val blockedAncestors: MutableSet<Int> = mutableSetOf()
+    )
+
+    private data class ExpansionDecision(
+        val frame: SearchFrame? = null,
+        val blockedAncestors: Set<Int> = emptySet()
+    )
+
+    private data class ExpansionRecord(
+        val depth: Int,
+        val blockedAncestors: Set<Int>
     )
 
     private fun matchesTarget(state: SearchState, options: SearchOptions): Boolean =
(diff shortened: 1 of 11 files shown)


graphite-cypher/src/test/kotlin/io/johnsonlee/graphite/cypher/FilteredRelationshipMemoryTest.kt
```diff
@@ -0,0 +1,45 @@
...
+                if (id == source.id && type == DataFlowEdge::class.java) {
+                    generateSequence { DataFlowEdge(source.id, target.id, DataFlowKind.ASSIGN) }
+                        .take(EDGE_COUNT) as Sequence<T>
+                } else {
+                    emptySequence()
+                }
+        }
+
+        val result = CypherExecutor(virtualGraph).execute(
+            "MATCH (a:IntConstant)-[r:DATA_FLOW]->(b:IntConstant) " +
```
johnsonlee, line 56: **[P1 blocking] Make the heap gate exercise matching DATAFLOW edges.** `resolveEdgeType("DATA_FLOW")` selects `DataFlowEdge` for the typed adjacency call, but `matchesRelConstraints` then compares the requested string `DATA_FLOW` with `CypherFunctions.type(edge) == DATAFLOW`; they are not equal, so every generated edge is rejected before a relationship binding reaches `WHERE`. I installed this exact test and dedicated 256 MiB task on base `5f9825d`; it also completed successfully, proving the gate does not distinguish the eager implementation from this change. Use `:DATAFLOW` here (or make the advertised alias semantically valid), add a positive control proving an edge can match, and demonstrate base fails the 1,000,000-edge heap test while the candidate passes. Otherwise the mandatory regression can stay green while the original retention bug returns.

johnsonlee: Follow-up discrimination run: correcting only the type to `:DATAFLOW` is still insufficient at the committed 1,000,000 edges—base `5f9825d` also passed the dedicated `-Xmx256m` task in 10 s. With the same probe at 4,000,000 edges, exact head `a1e5911` passed (`BUILD SUCCESSFUL`, 19.84 s) while base failed with `java.lang.OutOfMemoryError` (`BUILD FAILED`, 20.17 s). Please calibrate the gate so base demonstrably fails and head passes on the CI environment (4M here, or a stricter heap/allocation assertion), in addition to fixing the relationship type and adding the positive match control.

johnsonlee: Fixed in fcb22a5. The heap probe now uses :DATAFLOW and 4,000,000 matching edges, and a positive control verifies that :DATA_FLOW binds a real edge; the alias comparison is now normalized so the advertised alias is valid. Discrimination run with the identical test/task: main 5f9825d failed under -Xmx256m with OutOfMemoryError, while head passed in 13 s. The full :cypher:check gate also passes.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
```diff
@@ -1359,17 +1360,83 @@ class QueryPipeline private constructor(
...
+     */
+    @Suppress("ComplexCondition", "CyclomaticComplexMethod", "MagicNumber", "ReturnCount")
+    private fun tryStreamingFilteredMatchLimit(clauses: List<CypherClause>): CypherResult? {
+        if (clauses.size !in 4..5) return null
+        val match = clauses.getOrNull(0) as? CypherClause.Match ?: return null
+        val where = clauses.getOrNull(1) as? CypherClause.Where ?: return null
+        val ret = clauses.getOrNull(2) as? CypherClause.Return ?: return null
+        val skip = clauses.getOrNull(clauses.lastIndex - 1) as? CypherClause.Skip
+        val limit = clauses.lastOrNull() as? CypherClause.Limit ?: return null
+        if (clauses.size == 5 && skip == null) return null
```
johnsonlee: **[P1 blocking] Keep filtered ORDER BY queries heap-bounded too.** This rejects the valid `MATCH, WHERE, RETURN, ORDER BY, LIMIT` shape and sends it back to the eager pipeline, which materializes every relationship binding before `WHERE`. On exact head `a1e5911`, I ran the corrected 4,000,000-edge probe under the committed `-Xmx256m` task with `WHERE b.value = -1 RETURN r ORDER BY b.value LIMIT 1`; it failed with `java.lang.OutOfMemoryError` in 21.40 s, while the no-ORDER candidate probe passes. A zero-hit filter needs to retain no sort rows, so ORDER BY should not require retaining four million pre-filter bindings. Stream WHERE first and maintain a bounded top-(SKIP+LIMIT) structure for ordering, then add a discriminating heap regression for this common shape.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
```diff
@@ -1889,6 +1995,11 @@ class QueryPipeline private constructor(
         }
     }
 
+    private fun matchesRelationshipType(edge: Edge, actual: String?, requested: String): Boolean =
+        requested.equals(actual, ignoreCase = true) ||
+            requested.replace("_", "").equals(actual?.replace("_", ""), ignoreCase = true) ||
+            (edge is ResourceEdge && requested.equals("RESOURCE", ignoreCase = true))
```
johnsonlee: **[P2 blocking] Keep relationship-type aliases explicit instead of deleting every underscore.** This makes undeclared/malformed names such as `:D_A_T_A_F_L_O_W`, `:C_A_L_L`, and `:T_Y_P_E` match DATAFLOW/CALL/TYPE edges even though `resolveEdgeType` does not recognize them; an unknown type now silently returns real rows. It also allocates normalized strings per inspected edge on alias scans. Canonicalize only the documented aliases once (for example `DATA_FLOW` ↔ `DATAFLOW` and `CONTROL_FLOW` ↔ `CONTROLFLOW`) or compare via an explicit alias table, and add negative execution tests for malformed/unknown types alongside the positive `DATA_FLOW` case.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
johnsonlee: Reverified on exact head `0755aaa0d25641601c19ccdd669107275605526e`: the blocker remains. I temporarily changed the committed 4M/256 MiB heap probe to `WHERE b.value = -1 RETURN r ORDER BY b.value LIMIT 1` and reran the dedicated task with `--rerun-tasks`; it failed with `OutOfMemoryError` at the test execution (`BUILD FAILED in 25s`, `real 26.17`). The `0755aaa` increment only adds unrelated traversal tests and does not change the eager ORDER BY fallback. The temporary change was removed and the review clone is clean.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
johnsonlee: Reverified on exact head `0755aaa0d25641601c19ccdd669107275605526e`: replacing the positive alias in the committed targeted test with malformed `:D_A_T_A_F_L_O_W` still returns the real `ASSIGN` edge and the test task succeeds (`BUILD SUCCESSFUL in 22s`). This confirms arbitrary underscore deletion is still over-accepting unknown types. The temporary change was removed and the review clone is clean.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
johnsonlee: Fixed in 2cf3b71. MATCH WHERE RETURN ORDER BY [SKIP] LIMIT now filters lazily before projection and retains only a stable top-(SKIP+LIMIT) PriorityQueue; DISTINCT uses a bounded selected-key map and preserves provenance. Added execution coverage for 100,000 repeated edges with DISTINCT/ORDER BY/SKIP/LIMIT plus CASE/null and multi-field mixed ordering. With the exact 4,000,000-edge ORDER BY probe under -Xmx256m, base 5f9825d fails with OutOfMemoryError in 17 s while head passes in 17 s. The new orderedFilteredSingleHopRelationship JMH is 165.224 ± 4.629 us/op on base and 171.838 ± 4.816 us/op on head (+4.0%, overlapping 99.9% CIs, below the 15% gate).

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
johnsonlee: Fixed in 2cf3b71. Relationship matching now uses explicit edge-subtype comparisons without underscore deletion or per-edge normalized-string allocation. Only documented spellings are accepted (including DATAFLOW/DATA_FLOW and CONTROL_FLOW/CONTROLFLOW); malformed D_A_T_A_F_L_O_W, C_A_L_L, T_Y_P_E, C_O_N_T_R_O_L_F_L_O_W, and the previously non-semantic RESOURCE_EDGE return no rows. Positive execution tests cover all canonical edge families and every RESOURCE_* kind, and resolver tests reject the malformed names.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
```diff
@@ -1359,17 +1361,163 @@ class QueryPipeline private constructor(
...
+        val selectedDistinctRows = if (ret.distinct) {
+            HashMap<Map<String, Any?>, RankedProjectedRow>()
+        } else {
+            null
+        }
+        var encounterOrder = 0L
+
+        for (bindings in matches) {
+            if (evaluator.evaluate(where.condition, bindings) != true) continue
+            val projected = projectRow(ret.items, columns, bindings)
```
johnsonlee, line 1474: **[P1 blocking] Preserve ORDER BY expressions that reference bindings omitted from RETURN.** This projects each match before ranking, and `rankedRowComparator` later evaluates every non-alias expression against only that projected map. A valid query such as `MATCH (a:IntConstant)-[:DATAFLOW]->(b:IntConstant) WHERE a.value = 42 RETURN b.id AS id ORDER BY b.value ASC LIMIT 2` therefore cannot see `b`; every sort key becomes null and the queue keeps encounter order. On exact head `2cf3b71660e932c768bf9f87dce9915f1e98a569`, I added three targets encountered with values `2,0,1`: the expected ids for values `0,1` were `[22,23]`, but the result was the first two encountered ids `[21,22]` (`BUILD FAILED in 16s`). The existing new tests sort only by returned aliases, so they do not cover this standard Cypher shape—including the original OOM repro query `RETURN r ORDER BY b.value`. Retain/evaluate hidden sort keys from the original bindings while keeping only top-(SKIP+LIMIT), strip them from the response, and add a regression for an unprojected property/expression.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
johnsonlee: Independently verified the original memory blocker fixed on exact head `2cf3b71660e932c768bf9f87dce9915f1e98a569`: the committed 4,000,000-edge `ORDER BY b.value LIMIT 1` probe passes under the dedicated `-Xmx256m` task (`BUILD SUCCESSFUL in 11s`, real 11.71s), and the focused ordered top-K tests pass. This thread can remain resolved. A separate correctness regression for ORDER BY expressions omitted from RETURN is tracked on the current implementation in discussion r3886832913.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
johnsonlee: Independently verified fixed on exact head `2cf3b71660e932c768bf9f87dce9915f1e98a569`. Relationship matching now accepts only the explicit documented aliases, the malformed spellings return no rows, and the focused execution/resolver tests pass (`BUILD SUCCESSFUL in 19s`). This thread can remain resolved.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
johnsonlee: Fixed in 0a914e2. Each ORDER BY expression is now evaluated once against a short-lived context containing the original MATCH bindings plus projected aliases; the ranked row retains only the resulting scalar sort-value list, projected output, and encounter order. The top-(SKIP+LIMIT) memory bound is unchanged, hidden sort keys are never exposed in the response, and the comparator no longer needs the omitted bindings. Added regressions where RETURN contains only b.id while ORDER BY uses the omitted b.index property and a CASE expression over b.index; both now return the value-sorted ids rather than encounter order. Exact-head :cypher:test, the 4M/-Xmx256m ordered heap task, detekt, JMH build, and Kover all pass (98.0327%).

johnsonlee: Independently verified fixed on exact head `0a914e2a104cddffbc5de7405a56981d1045dfcc`. The implementation now computes scalar sort values from the original bindings plus projected aliases before discarding the bindings, while the queue retains only those values and the bounded top-K rows. The committed hidden-property/hidden-CASE test passes, and the exact prior probe with encounter values `2,0,1` now returns the ids for values `0,1` (`BUILD SUCCESSFUL in 19s`). The review clone is clean. This thread can remain resolved.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
```diff
@@ -1696,36 +1896,42 @@ class QueryPipeline private constructor(
...
+        rel: PatternElement.RelationshipPattern,
+        targetNodePattern: PatternElement.NodePattern,
+        bindings: Map<String, Any?>
+    ): Sequence<Map<String, Any?>> {
+        val sourceNode = sourceNodePattern.variable
+            ?.let(bindings::get)
+            ?.let(::nodeCursor)
+            ?: findLastBoundNode(bindings)
+            ?: return emptySequence()
+        val edgeClass = rel.types.singleOrNull()?.let { NodePropertyAccessor.resolveEdgeType(it) }
```
johnsonlee: **[P1] Preserve relationship type constraints on variable-length paths**

`singleOrNull()` turns both a type union and an unrecognized spelling into `null`, but the variable-length branch passes that value to `PathFinder`, where `edgeType = null` means “traverse every edge”; unlike the single-hop branch, it never calls `matchesRelConstraints`. On exact `ffc1be0`, a four-node graph with `DATAFLOW 1->2`, `CALL 1->3`, and `TYPE 1->4` produces `[{b.id=2}, {b.id=3}, {b.id=4}]` for `MATCH (a:IntConstant)-[:DATAFLOW|CALL*1..1]->(b:IntConstant) WHERE a.value = 3 RETURN b.id ORDER BY b.id LIMIT 10`, even though node 4 is reachable only via TYPE. Replacing the union with the malformed alias `D_A_T_A_F_L_O_W` still returns all three rows, while the equivalent single-hop query returns none. Please carry the requested type set/predicate into variable-length traversal (and reject unresolved types) rather than using null as both “untyped” and “cannot resolve”.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/PathFinder.kt
```diff
@@ -110,12 +110,17 @@ object PathFinder {
             for (edge in edges) {
                 val nextId = nextNodeId(edge, current, options.direction)
                 val nextNode = loadNode(graph, nextId, options.workTracker) ?: continue
-                queue.add(SearchState(nextNode, edge, state, state.depth + 1))
+                val nextState = SearchState(nextNode, edge, state, state.depth + 1)
+                if (matchesTarget(nextState, options)) yield(PathMatch(nextState))
+                queue.add(nextState)
```
johnsonlee, line 116: **[P1] Do not retain terminal variable-length states after yielding them**

The lazy `yield` does not keep a zero-hit variable-length query memory-bounded: after the consumer rejects each yielded row in `WHERE`, execution resumes here and enqueues every state even when `nextState.depth == maxDepth`, although those states can never be expanded. I reproduced this on exact `ffc1be0` with a generated DATAFLOW adjacency and `MATCH (a:IntConstant)-[:DATAFLOW*1..1]->(b:IntConstant) WHERE b.value = -1 RETURN b.id ORDER BY b.id LIMIT 1`: 10,000,000 edges under `-Xmx256m` fail with `OutOfMemoryError` from `projectOrderedFilteredRows` after ~3.7 s because the BFS queue grows linearly. The new heap gate covers only the single-hop matcher, so it misses this advertised lazy variable-length path. At minimum, avoid enqueueing states at `maxDepth`; to bound deeper searches as well, traversal needs a frontier strategy that does not retain the whole width.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
johnsonlee: Fixed in f303486. Variable-length traversal now applies the same `matchesRelConstraints` predicate on every followed edge, while all-unresolved requested types return an empty sequence instead of falling through to untyped traversal. The exact four-node regression now returns `[2, 3]` for `DATAFLOW|CALL*1..1` and excludes TYPE; `D_A_T_A_F_L_O_W*1..1` returns no rows. Full Cypher tests (1048), detekt, Kover (98.0408%), and the 256 MiB heap gate pass in an isolated clone.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/PathFinder.kt
johnsonlee: Fixed in f303486. PathFinder now uses an explicit depth-first iterator stack: terminal `maxDepth` matches are yielded without being retained, and deeper lazy traversal retains only O(maxDepth) SearchFrame/Node/parent state rather than an entire breadth frontier. A shallowest-expanded-depth map preserves completeness when a node is reached by a shorter path later. The exact 10,000,000-edge `*1..1`, rejected WHERE, ORDER BY/LIMIT repro now passes under the committed `-Xmx256m` task (pre-fix `ffc1be0` OOMs); added tests also prove deep-first lazy consumption and shallower-revisit expansion. Full Cypher tests, detekt, Kover, JMH build/smoke, and the heap gate pass.

johnsonlee: Rechecked exact head `f303486`. The original `*1..1`/repeated-target repro is fixed, but the claimed deeper memory bound is not: `expandedAtDepth` retains one map entry for every distinct expanded node. I changed only the topology to one source with 10,000,000 distinct DATAFLOW targets, each with no outgoing edges, and ran the same rejected `WHERE`, `ORDER BY ... LIMIT 1` query with `*1..2` through the committed `filteredRelationshipMemoryTest` task. Under its exact `-Xmx256m`, it fails with `OutOfMemoryError` after about 11 s. The iterator stack is O(maxDepth), but the map is O(traversal width); the committed `*1..1` test never inserts terminal targets into that map and repeats one target ID, so it cannot detect this. Keeping this thread unresolved until deeper distinct-width traversal is actually bounded and covered by the heap gate.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
johnsonlee: Verified fixed on exact head `f303486`: the focused committed regression passes, `DATAFLOW|CALL*1..1` returns only ids `[2, 3]`, and malformed `D_A_T_A_F_L_O_W*1..1` returns no rows. Static inspection confirms every variable-length edge now passes through `matchesRelConstraints`, while an all-unresolved requested type set returns an empty sequence. Resolving this scoped type-constraint thread.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/PathFinder.kt
```diff
@@ -82,45 +83,72 @@ object PathFinder {
     ): Sequence<PathMatch> = sequence {
         for (source in sources) {
             val startNode = loadNode(graph, source, options.workTracker) ?: continue
-            yieldAll(bfs(graph, startNode, options))
+            yieldAll(depthFirst(graph, startNode, options))
```
johnsonlee, line 89: **[P1 blocking] Preserve the public `PathFinder.findPaths` BFS result order while optimizing Cypher traversal.** Before this commit, this public API was explicitly documented and implemented as BFS, so its returned `List<Path>` was shortest-first. An exact three-node probe whose source adjacency is ordered `source->middle`, then `source->target`, with `middle->target`, `targets={target}`, and `maxDepth=2` returns edge lengths `[1, 2]` on `ffc1be0`; on `f303486`, this `yieldAll(depthFirst(...))` change returns `[2, 1]`. The query pipeline may not promise row order without `ORDER BY`, but that does not erase the separate public API contract. Keep the public BFS behavior (or preserve its shortest-first output) and use a distinct internal lazy traversal for Cypher; add the base/head probe as a regression.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/PathFinder.kt
johnsonlee: Fixed in 12934ff. The Cypher-internal depth-first iterator no longer keeps a traversal-wide expanded-node map. It suppresses only expansion of a node already on the current parent chain, so retained search state is O(maxDepth) even across distinct-width frontiers while a node reached through a different branch remains expandable. The committed heap gate now uses 10,000,000 distinct target IDs, *1..2, a rejected WHERE, and ORDER BY/LIMIT under -Xmx256m; it passes in 12 s. Full Cypher tests (1049), detekt, Kover (98.05%), and JMH build/smoke pass.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/PathFinder.kt
johnsonlee: Fixed in 12934ff. Public PathFinder.findPaths once again uses a dedicated breadth-first traversal and its KDoc explicitly preserves the breadth-first contract; the Cypher pipeline continues to use the separate internal depth-first lazy iterator. Added the exact long-branch-first regression: source->middle precedes source->directTarget and middle->deepTarget, and public results are asserted at edge lengths [1, 2] with endpoints [directTarget, deepTarget]. The existing guarded lazy-descent test still verifies the internal traversal does not consume a wide sibling frontier before the first deep match. Full Cypher tests (1049), detekt, the 256 MiB heap gate, Kover, and JMH build pass.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/PathFinder.kt
johnsonlee: Verified fixed on exact head `12934ff`: the committed 10,000,000-distinct-target `*1..2` probe passes the dedicated `-Xmx256m` task, and inspection confirms no traversal-wide expanded-node collection remains. The original width-retention/OOM thread is correctly resolved; the separate reconvergent re-expansion cost is tracked independently.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/PathFinder.kt
johnsonlee: Verified fixed on exact head `12934ff`: the focused public API regression passes and returns path edge lengths `[1, 2]`; `findPaths` now uses a dedicated breadth-first iterator while Cypher keeps its internal lazy traversal. This scoped BFS-contract thread is correctly resolved.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/PathFinder.kt
```diff
@@ -99,28 +103,87 @@ object PathFinder {
...
+
+            val edge = frame.edges.next()
+            val current = frame.state.node.id
+            val nextId = nextNodeId(edge, current, options.direction)
+            val nextNode = loadNode(graph, nextId, options.workTracker) ?: continue
+            val nextState = SearchState(nextNode, edge, frame.state, frame.state.depth + 1)
+            if (matchesTarget(nextState, options)) yield(PathMatch(nextState))
+
+            if (nextState.depth < options.maxDepth &&
+                !hasAncestor(frame.state, nextId)
```
johnsonlee: **[P1 blocking] Do not trade the retained-width fix for exponential reconvergent re-expansion.** Exact `12934ff` probe: a finite 65-node/456-edge layered DAG (8 nodes per layer, full 8x8 DATAFLOW adjacency across 8 layers) runs `MATCH (a:BooleanConstant)-[:DATAFLOW*1..8]->(b:IntConstant) WHERE b.value = -1 RETURN b.id ORDER BY b.id LIMIT 1` with the server default `CypherExecutionBudget(250_000)`. This head throws `CypherBudgetExceededException` at exactly 250,000 work units; `f303486` returns the correct empty result under the same budget. Ancestor-only suppression expands the same reconvergent suffix once per simple prefix, O(number of paths), even though the endpoint filter rejects every row. The committed heap probe has no edges beyond depth 1, and `filteredVariableLengthPath` benchmarks only `*1..1`, so both miss this regression. Bound reconvergent expansion without restoring the traversal-wide HashMap (or push the endpoint filter/deduplication before suffix enumeration where path multiplicity is unobservable), and add this default-budget layered-DAG case to the gate.

johnsonlee: **[P1 blocking] Restore the exact-head repository resource gate.** Required run `33270709408` on exact `12934ff` failed `wrapped-query-resources`. The uploaded status is fail-closed with no harness errors: `AllFixtureWrappedDiscoveryResourceBenchmark.allFixtureThirtySixGraphFootprint` allocation increased from `4,651,195,330.7` to `6,104,712,746.7 B/op` (`+31.2504%`, threshold 15%). Its built-in reverse-order confirmation reproduced essentially the same result (`4,651,195,368.0 -> 6,104,712,994.7 B/op`, `+31.2504%`) and remains blocked. This comment does not claim this line alone causes that delta; it records the repository benchmark gate that we agreed is the safety criterion. Please make the same exact-head gate pass (a clean passing rerun is sufficient under that rule) rather than bypassing or weakening it.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/PathFinder.kt
```diff
@@ -99,28 +103,87 @@ object PathFinder {
...
+            ) {
+                stack.addLast(SearchFrame(nextState, edgesForDirection(graph, nextId, options).iterator()))
```
johnsonlee: **[P1 blocking] Restore the exact-head large-corpus gate.** Required run `33270709408` on exact `12934ff` also failed the end-to-end real-corpus comparator. For `kotlin-compiler`, mapped load regressed from `254` to `351 ms` (`+38.2%`, threshold 30%); the automatic reverse-order confirmation repeated it at `248 -> 338 ms` (`+36.3%`) and left `large-corpus-status.json` at `passed:false` with no harness errors. This is independent of the wrapped-allocation and reconvergent-traversal threads, and does not assert this specific line is the causal site. Under the agreed repository-gate rule, please make this exact-head gate pass; a clean passing rerun is sufficient rather than extra ad hoc proof.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/PathFinder.kt
johnsonlee: Verified no longer blocking on the same exact head `12934ff` after failed-job rerun attempt 2. Artifact `benchmark-latency-resources-102-2` has `passed:true`, no errors, and no blocked rows; the 36-graph allocation is now `4,651,196,520 -> 4,651,195,504 B/op` (effectively unchanged), with GC count `-7.7%`, GC time `-12.5%`, retained heap `-0.6%`, and peak heap `-0.02%`. Per the agreed benchmark-gate safety criterion, this clean exact-head pass is sufficient. Resolving this resource thread.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/PathFinder.kt
johnsonlee: Verified no longer blocking on the same exact head `12934ff` after failed-job rerun attempt 2. Artifact `benchmark-large-corpus-102-2` has `passed:true`, no errors, and no blocked rows. The previously failing kotlin-compiler mapped load is now `264 -> 309 ms` (`+17.0%`, below the 30% threshold); all other enforced build/save/load/query/pipeline rows also pass. Per the agreed benchmark-gate criterion, this clean exact-head pass is sufficient. Resolving this large-corpus thread. The attempt-2 aggregate failure is the separate cross-attempt artifact reconstruction defect already tracked on #101, not a remaining verdict from this gate.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/PathFinder.kt
johnsonlee: Attempt 3 provides newer contradictory evidence, so this resource thread is blocking again. The exact same head `12934ff` now has `benchmark-latency-resources-102-3` with `passed:false` and no harness errors. Allocation is healthy in this run, but the 36-graph query GC count regressed from `2.000` to `4.333/op` (`+116.7%`), and the built-in reverse-order confirmation repeated `2.000 -> 4.000/op` (`+100%`), both above the 15% gate. Reopening because the latest full rerun fails the repository safety criterion; a later clean exact-head gate pass remains sufficient under the agreed rule.

johnsonlee: Attempt 3 has now completed on exact head `12934ff`. The authoritative aggregate is **FAIL**: every latency shard (including the 36-real-graph shard), method-level, large-corpus, Explorer, and budgeted sections passed, but the 36-graph resource probe still fails query GC count (`2.000` → `4.333`, `+116.7%`). This confirms the already reopened resource blocker; the final aggregate introduced no additional independent benchmark failure.

johnsonlee: Fixed in efb24e4. The lazy DFS now combines path-local cycle suppression with a fixed-capacity (16,384-entry) shallowest-depth expansion memo. Reconvergent suffixes within the memo are expanded once at the shallowest observed depth; bounded eviction can only cause safe re-expansion, never missed results, and retained memory cannot grow with traversal width. Leaf nodes are not memoized after the adjacency lookahead, so the 10,000,000-distinct-target *1..2 heap probe remains bounded. Added the exact 65-node/456-edge, width-8 x depth-8 layered DAG regression under CypherExecutionBudget(250_000); it returns the correct empty result. The 256 MiB task (all 4 probes), 1049 regular tests, detekt, Kover (97.9759%), JMH build, and filteredVariableLengthPath smoke all pass.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/PathFinder.kt
```diff
@@ -99,28 +103,118 @@ object PathFinder {
...
+    private fun expandableFrame(
+        graph: Graph,
+        nodeId: NodeId,
+        state: SearchState,
+        options: SearchOptions,
+        expandedAtDepth: LinkedHashMap<Int, Int>
+    ): SearchFrame? {
+        if (state.depth >= options.maxDepth || hasAncestor(state.parent, nodeId)) return null
+        val previousDepth = expandedAtDepth[nodeId.value]
+        if (previousDepth != null && previousDepth <= state.depth) return null
```
johnsonlee: **[P1 blocking] Do not memoize expansion by node/depth alone; this drops valid simple paths.** Exact-head `efb24e4` reproduction with edge order `S→A`, `S→B`, `A→X`, `A→T`, `B→X`, `X→A`: `MATCH (a:BooleanConstant)-[:DATAFLOW*4..4]->(b:IntConstant) RETURN b.id` returns empty here, while the identical test on parent `12934ff` returns `T`. The first branch `S-A-X` records `X` at depth 2, but `X→A` is correctly rejected because `A` is an ancestor. On the later branch `S-B-X`, `X→A→T` is legal, yet this line suppresses expanding `X` because it ignores the different ancestor set. The committed zero-hit layered-DAG test proves work reduction but cannot detect missing matches. Preserve the branch-dependent ancestor context in any memo key (within a fixed-capacity bound), or avoid this unsafe pruning, and add this positive-result regression.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/PathFinder.kt
johnsonlee: Verified on exact head `efb24e4`: the committed 65-node/456-edge width-8 × depth-8 zero-hit query now completes under the default `250_000` work budget in the dedicated 256 MiB task. This fixes the exponential reconvergent re-expansion reported in this thread, so I am resolving it. The separate positive-result correctness regression caused by node/depth-only memoization is tracked independently at discussion `r3887609696`.

johnsonlee: Verified the exact-head resource gate on `efb24e4`: job `99156023107` completed successfully and the uploaded `latency-resource-status.json` is `passed:true` with no errors. The 36-graph GC-count suspicion from the initial order (`3.000` → `4.333`) did not reproduce in reverse order (`6.000` → `6.000`); allocation, retained heap, peak heap, and all single-graph rows also pass. Per the repository benchmark-gate rule, the prior resource regression is no longer established, so I am resolving this thread.

graphite-cypher/src/test/kotlin/io/johnsonlee/graphite/cypher/FilteredRelationshipMemoryTest.kt
```diff
@@ -0,0 +1,143 @@
...
+
+        val result = CypherExecutor(virtualGraph).execute(
+            "MATCH (a:IntConstant)-[:DATAFLOW*1..2]->(b:IntConstant) " +
+                "WHERE b.value = -1 RETURN b.id ORDER BY b.id LIMIT 1"
+        )
+
+        assertTrue(result.rows.isEmpty())
+    }
+
+    @Test
```
johnsonlee, line 103: **[P1 blocking] Keep the required Cypher coverage gate green.** Exact-head run `33273463387` completed the tests but failed the required check with `cypher coverage is 97.9759%, below 98%`. This regression lives in `FilteredRelationshipMemoryTest`, which `graphite-cypher/build.gradle.kts` explicitly excludes from the standard `test` task, so it does not supply the main Kover execution data for the new `expandableFrame`/memo branches. Add a regular-sized regression in the standard test suite or merge this custom task coverage into Kover; the PR is not ready while the required coverage gate fails.

johnsonlee: Verified on exact head `92cb83e`: required unit run `33274125793` completed successfully, all tests/build steps passed, and the published Cypher line coverage is now `98.1%` (above the required `98%`). The new regular `PathFinderTest` also passes locally. Resolving this coverage thread.

johnsonlee: Fixed in `92cb83e` and retained in current head `84038e9`. I added a regular `PathFinderTest` that drives the fixed-capacity memo past 16,384 entries, so the eviction branch is now part of the standard Kover input. Required unit run 33274125793 passed on `92cb83e`; current-head local verification reports 1,052 regular tests + 4 isolated heap/budget tests, detekt clean, and 98.0691% Cypher line coverage.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/PathFinder.kt
johnsonlee: Fixed in current head `84038e9`. The memo no longer suppresses by node/depth alone: each completed suffix records the ancestor IDs that actually blocked traversal, and a same-depth revisit is suppressed only if the new branch still contains every recorded blocker. Thus the first `S→A→X` visit records `A`; the later `S→B→X` visit lacks that blocker and expands `X→A→T`. I added the exact Cypher regression from this comment and an unequal-depth exact-match regression. Both pass, while the 65-node/456-edge layered DAG still completes under 250,000 work units and the 10,000,000-target test still passes under `-Xmx256m`. Full local result: 1,052 regular + 4 isolated tests, detekt clean, Kover 98.0691%.

johnsonlee: `84038e9c` fixes the original ancestor-unlock example, but the thread remains blocking because the memo still collapses distinct legal paths when the prior blocker set is empty. Exact reproduction: diamond edges `S→L`, `S→R`, `L→X`, `R→X`, `X→T`, followed by `MATCH (a:BooleanConstant)-[:DATAFLOW*3..3]->(b:IntConstant) RETURN b.id AS id`. Parent `12934ff` returns two `T` rows (one per distinct relationship path); exact head `84038e9c` returns only one. After the first prefix completes `X`, its memo record has an empty `blockedAncestors`; every later same-depth ancestor set contains the empty set, so the second prefix is suppressed even though Cypher without `DISTINCT` must preserve both paths. Preserve result multiplicity across reconvergent prefixes (for example by replaying a safely memoized suffix per prefix), or restrict pruning to a case where it cannot remove result-producing paths, and add this exact regression.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/PathFinder.kt
```diff
@@ -99,28 +103,155 @@ object PathFinder {
...
+        stack.addLast(SearchFrame(start, startEdges))
+
+        while (stack.isNotEmpty()) {
+            val frame = stack.last()
+            if (!frame.edges.hasNext()) {
+                val completed = stack.removeLast()
+                rememberBounded(
+                    expansionMemo,
+                    completed.state.node.id.value,
+                    ExpansionRecord(completed.state.depth, completed.blockedAncestors.toSet())
```
johnsonlee: **[P1 blocking] Bound the retained blocker sets, not only the memo entry count.** Each completed frame copies its entire accumulated blocker union here, then line 158 propagates that union to every parent. A fixed 16,384-entry map therefore does not bound memory: one deep suffix can make every entry retain an O(depth) set. On exact head `84038e9c`, I added a 3,000-node chain whose last node has one edge back to each ancestor and consumed it through `findPathMatches` with the normal 250,000-work budget. The dedicated `filteredRelationshipMemoryTest` JVM (`-Xmx256m`) throws `OutOfMemoryError` after 2.9 s; the identical test on parent `12934ff` completes successfully in the same 256 MiB task. This graph uses only O(depth) nodes/edges/work but this memo retains O(depth²) boxed IDs, so the PR can replace the original wide-frontier OOM with a deep-cycle OOM. Store a bounded/compact blocker representation (or otherwise cap total retained bytes, not just record count) and add this exact low-heap regression alongside the reconvergent correctness cases.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/PathFinder.kt
```diff
@@ -99,28 +103,155 @@ object PathFinder {
...
+            }
+
+            val edge = frame.edges.next()
+            val current = frame.state.node.id
+            val nextId = nextNodeId(edge, current, options.direction)
+            val nextNode = loadNode(graph, nextId, options.workTracker) ?: continue
+            val nextState = SearchState(nextNode, edge, frame.state, frame.state.depth + 1)
+            if (matchesTarget(nextState, options)) yield(PathMatch(nextState))
+
+            val expansion = expandableFrame(graph, nextId, nextState, options, expansionMemo)
```
johnsonlee: **[P1 blocking] Restore the existing filtered variable-path latency gate.** Exact head `84038e9c`, run `33274557425`, job `99158978418` failed the repository's fail-closed Method gate on `CypherBenchmark.filteredVariableLengthPath`: base `358.967 us/op` versus candidate `456.710 us/op` (`+27.2%`, limit 15%). The required reverse-order confirmation reproduced the same benchmark at `355.702 -> 452.377 us/op` (`+27.2%`), so this is not accepted as run-order noise. That workload is a zero-hit `[:DATAFLOW*1..1]` query; this changed loop now calls `expandableFrame` and merges an `ExpansionDecision` for every candidate edge even though each next state is already at `maxDepth`, which is the adjacent new hot-path overhead to inspect. Please remove the regression without weakening the threshold and make this exact benchmark gate pass; per the agreed safety rule, a clean exact-head rerun of the existing gate is sufficient.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/PathFinder.kt
johnsonlee: Rechecked on exact head `a1daa30`: this commit only changes the depth-limit hot path and does not change memo result semantics. The same diamond regression still fails at `assertEquals(2, result.rows.size)` because the query returns one `T` row instead of the two distinct legal relationship paths. Keeping this thread unresolved.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/PathFinder.kt
johnsonlee: Rechecked on exact head `a1daa30` with the same dedicated 256 MiB task. The 3,000-node chain plus terminal back-edges still throws `OutOfMemoryError` at the terminal `count()`; `ExpansionRecord(completed.blockedAncestors.toSet())` and parent propagation are unchanged. The new max-depth fast path does not bound these retained blocker sets, so this memory thread remains unresolved.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/PathFinder.kt
johnsonlee: `a1daa30` is directionally targeted at this gate failure: it bypasses the memo/decision path for `maxDepth == 1` and avoids constructing an expansion result once the next state reaches the depth limit. Keeping the thread open until the exact-head repository gate reruns the same benchmark and passes; a clean gate result is sufficient.

johnsonlee: Verified fixed by the exact-head repository gate on `a1daa30`: job `99161210321` completed SUCCESS, the artifact is nonempty, `method-status.json` is `passed:true` with 14 rows and no errors, and `filteredVariableLengthPath` is now `342.155 -> 387.698 us/op` (`+13.3%`, below the 15% limit). Per the agreed benchmark-gate safety criterion, this is sufficient; resolving this latency thread. The separate path-multiplicity and deep-cycle-memory threads remain blocking.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/PathFinder.kt
johnsonlee: Fixed in `3080bc8`. The traversal no longer memoizes result-producing suffixes at all: a frame is eligible for the fixed-capacity dead-suffix memo only after it completes with `foundMatch == false` and `cycleBlocked == false`. Therefore both diamond prefixes replay `X→T` and the exact regression now returns two `T` rows without `DISTINCT`. The final WHERE/target predicate is pushed into the last unbound variable-length traversal, so the zero-hit layered DAG can still memoize genuinely dead suffixes and completes under 250,000 work units.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/PathFinder.kt
johnsonlee: Fixed in `3080bc8`. `ExpansionRecord` and all retained blocker sets are removed. Each frame now carries only two booleans (`foundMatch`, `cycleBlocked`), and any cycle-dependent suffix is simply not memoized. I added the exact 3,000-node chain with terminal back-edges to every ancestor to the dedicated `-Xmx256m` task with the normal 250,000-work budget; it passes together with the 10,000,000-target and layered-DAG regressions. The isolated task now has 5 tests, 0 failures.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
```diff
@@ -1743,27 +1966,34 @@ class QueryPipeline private constructor(
...
+        resultPredicate: ((Map<String, Any?>) -> Boolean)?
+    ): Sequence<Map<String, Any?>> = sequence {
         val direction = when (rel.direction) {
             Direction.OUTGOING -> PathFinder.Direction.OUTGOING
             Direction.INCOMING -> PathFinder.Direction.INCOMING
             Direction.BOTH -> PathFinder.Direction.BOTH
         }
 
         val workTracker = if (workTrackingEnabled) activeWorkTracker.get() else null
+        val nodePredicate = if (resultPredicate != null && rel.variable == null) {
```
johnsonlee: **[P1 blocking] Feed target-node matching into the dead-suffix memo even when there is no WHERE.** This predicate is currently created only from `resultPredicate`, so a plain zero-hit target label leaves it null. On exact head `3080bc8`, I built the same 8-layer × 8-wide reconvergent DAG used by the committed budget test, with one `BooleanConstant` source and only `IntConstant` descendants, then ran `MATCH (a:BooleanConstant)-[:DATAFLOW*1..8]->(b:CallSiteNode) RETURN b.id LIMIT 1` under the normal 250,000-work budget. It throws `CypherBudgetExceededException` instead of returning an empty result. With `targets = null` and no `nodePredicate`, `PathFinder.matchesTarget` marks every reachable Int node as a match, so every frame becomes `foundMatch` and the new dead-suffix memo is never populated. The committed layered test passes only because its `WHERE b.value = -1` supplies `resultPredicate`. Always provide a predicate for the target label/inline constraints, and additionally evaluate WHERE only when its required bindings are available (for example when there is no relationship variable); add this no-WHERE/absent-label case to the 256 MiB budget task.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/PathFinder.kt
johnsonlee: Independently verified on exact head `3080bc8`: the complete `PathFinderTest` passes, including the committed diamond assertion that the non-`DISTINCT` query returns two `T` rows. The dead-suffix memo now excludes result-producing frames, so the original multiplicity failure is fixed. Leaving this thread resolved; the no-WHERE target-label budget gap is tracked separately.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/PathFinder.kt
johnsonlee: Independently verified on exact head `3080bc8`: the full dedicated `:cypher:filteredRelationshipMemoryTest` task passes under its configured `-Xmx256m`, including the committed 3,000-node chain with terminal back-edges. The retained blocker sets that caused the original OOM are gone. Leaving this thread resolved; the separate no-WHERE target-label budget gap does not invalidate this memory fix.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/PathFinder.kt
```diff
@@ -99,28 +104,170 @@ object PathFinder {
...
+     * Matches keep the graph's edge order within each branch. Cypher does not define a global result
+     * order without ORDER BY, so depth-first traversal is safe for the lazy query pipeline while allowing
+     * LIMIT to stop before unrelated siblings are loaded.
+     */
+    private fun depthFirst(
+        graph: Graph,
+        startNode: Node,
+        options: SearchOptions
+    ): Sequence<PathMatch> = sequence {
+        val deadSuffixMemo = LinkedHashMap<ExpansionKey, Unit>()
```
johnsonlee, line 140: **[P1 blocking] Restore the existing wrapped-query resource gate on this traversal.** On exact head `3080bc8`, [run 33276692329 / job 99164563389](https://github.com/johnsonlee/graphite/actions/runs/33276692329/job/99164563389) produced a complete 10-row artifact with `latency-resource-status.json` at `passed:false`. For the 36-real-fixture query, query GC count went from base `4.000` to candidate `6.000` per operation (`+50%`), then repeated in reverse order at base `2.000` versus candidate `4.000` (`+100%`). The apparent allocation increase did not repeat and was correctly classified as noise; retained heap, peak heap, single-graph metrics, and the previous exact head `a1daa30` all pass, so this blocker is specifically the reverse-confirmed GC-frequency regression introduced by the current update. Please make the same exact repository gate pass without weakening its threshold; that passing gate is sufficient evidence.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
johnsonlee: Fixed in ab80322. Target NodePattern constraints (labels, inline properties, and an already-bound target variable) now participate in PathFinder.nodePredicate even when there is no WHERE. The final WHERE is additionally pushed only when the variable-length relationship is unbound, so predicates that require the relationship path are still evaluated after materialization/binding. Added the exact 65-node/456-edge, 8×8 reconvergent no-WHERE absent-CallSiteNode regression under the normal 250,000-work budget and 256 MiB heap, plus an inline-property miss case and a relationship-variable timing test. Exact-head verification: 1054 regular tests + 7 isolated memory/budget tests, detekt, Kover (98.0914%), and jmhJar all pass.

johnsonlee: Independently verified on exact head `ab80322`. The implementation now builds `nodePredicate` from target labels, inline properties, and an already-bound target even without `WHERE`, while it only pushes the final predicate before path materialization when the relationship variable is absent. I reran the complete `:cypher:filteredRelationshipMemoryTest` under its configured 256 MiB heap; all seven cases pass, including the exact 8×8 no-WHERE absent-`CallSiteNode` repro and the inline-property miss. I also reran the focused relationship-variable timing test and it passes. The original budget exception is fixed without evaluating `WHERE r ...` before `r` is bound; leaving this thread resolved.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/PathFinder.kt
johnsonlee: The unchanged resource gate now passes on exact head ab80322: [run 33277354935 / wrapped-query-resources](https://github.com/johnsonlee/graphite/actions/runs/33277354935/job/99166417362). No threshold or harness change was made. For `allFixtureThirtySixGraphFootprint`, the uploaded gate report records allocation +2.1% (4,556,575,309.3 -> 4,651,195,429.3 B/op), query GC count +8.3% (4.000 -> 4.333), query GC time +13.3% (5.000 -> 5.667 ms), retained heap delta -0.7%, and peak used heap -0.1%; every row is PASS and `latency-resource-status.json` has `passed: true`. The benchmark is the pure-node ZERO_HIT_QUERY and does not enter PathFinder, which is also consistent with the earlier exact `a1daa30` pass where base/candidate allocation differed by only 98.7 B/op. This exact-head passing repository gate is the requested evidence.

johnsonlee: Verified no longer blocking on exact head `ab80322`. [Wrapped-query resource job 99166417362](https://github.com/johnsonlee/graphite/actions/runs/33277354935/job/99166417362) completed successfully, and its nonempty artifact reports `passed:true`, 10 rows, no errors, and no blocked rows. The 36-real-fixture metrics are now allocation `+2.1%`, query GC count `4.000 -> 4.333/op` (`+8.3%`), GC time `+13.3%`, retained heap `-0.7%`, and peak heap `-0.1%`; all single-graph rows pass as well. Per the repository benchmark-gate safety criterion, this clean exact-head rerun is sufficient, so I am resolving the GC-frequency thread.

PR #105, opened 2026-08-29
Fix P0 Cypher read semantics

johnsonlee:
## Summary

- add public parameterized execution APIs and make unknown labels produce no matches
- align null comparison, `IN`, aggregate, `OPTIONAL MATCH ... WHERE`, and `ORDER BY` behavior with openCypher read semantics
- implement relationship-simple variable-length paths, relationship lists, multiple relationship types, per-hop properties, bound relationship reuse, and exact named-path materialization
- add focused compatibility and regression coverage and document the supported read-oriented Cypher subset

## Scope

This PR addresses P0 read semantics only. It intentionally excludes write clauses such as `CREATE`, `MERGE`, `SET`, and `DELETE`. It also does not claim a full openCypher TCK pass rate or introduce an official TCK runner; the added tests are focused regressions for the audited P0 gaps.

## Verification

- `./gradlew -POSSRH_USERNAME= :cypher:check :cypher:koverLog --rerun-tasks --no-daemon`
- latest Cypher module check: 1126 tests, 0 failures/errors; detekt passed; 98.0258% line coverage
- all module coverage gates pass: Core 98.1785%, Explore 98.0066%, Query 98.9011%, SootUp 98.4449%, WebGraph 98.042%
- `node --test .github/scripts/benchmark-gate.test.mjs`: 33/33 pass
- large-corpus gates pass for Hive, Kotlin compiler, and Tika
- focused local JMH smoke: `simpleNodeMatch` 27.923 us/op; `variableLengthPath` 40.911 us/op
- local same-host list-membership A/B: main 423.584 us/op; PR 386.528 us/op (-8.7%)
- local same-host method A/B after hot-path tuning: `aggregationCountGroupBy` +8.6%, `nodeMatchWithWhere` +3.5%, `withPipeline` +8.5%; all below the 15% regression gate

The required `Benchmark regression` CI workflow remains the authoritative base-versus-PR comparison.


graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/ExpressionEvaluator.kt
```diff
@@ -447,39 +456,58 @@ private fun appendWithCancellation(
...
+
+private fun evaluateMembership(
+    list: List<*>,
+    element: Any?,
+    checkCancelled: (() -> Unit)?
+): Boolean? {
+    var containsNullComparison = false
+    for ((index, candidate) in list.withIndex()) {
+        if ((index and CANCELLATION_POLL_MASK) == 0) checkCancelled?.invoke()
+        when (cypherEquals(element, candidate)) {
```
johnsonlee: **[P1 blocking] Keep openCypher `IN` semantics without regressing the budgeted membership hot path.** Exact-head CI reproduced a stable regression in `budgetedListMembership`: base `700.623 us/op` vs candidate `904.697 us/op` (`+29.1%`), with separated 99.9% confidence intervals; the reverse-order confirmation is base `697.674` vs candidate `876.021 us/op` (`+25.6%`). The sibling list-concatenation row is `-6.0%`, so this is localized to membership rather than runner noise. This loop now invokes recursive `cypherEquals` for every one of 100,000 numeric candidates, whereas the previous same-type list path used optimized `contains`. Please retain the required null/cross-number semantics while restoring a fast common numeric/list path, then rerun the existing fail-closed gate; no additional proof beyond a passing repository gate is required.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/ExpressionEvaluator.kt
```diff
@@ -447,39 +456,58 @@ private fun appendWithCancellation(
...
 
-private fun containsWithCancellation(
-    list: List<*>,
-    element: Any?,
-    checkCancelled: (() -> Unit)?
-): Boolean = when {
-    checkCancelled == null -> element in list
-    list is RandomAccess -> containsRandomAccessList(list, element, checkCancelled)
-    else -> containsSequentialList(list, element, checkCancelled)
+private fun cypherEquals(left: Any?, right: Any?): Boolean? = when {
```
johnsonlee: **[P1 blocking] Apply the new Cypher equality semantics to simple `CASE` as well.** `cypherEquals` correctly treats numeric values by value across runtime number types, but `evaluateCase` still compares its test and `WHEN` values with Kotlin `==`. On exact head `94c8673`, I added the end-to-end query `RETURN CASE 1 WHEN 1.0 THEN "same" ELSE "different" END AS value`; `:cypher:test` failed because the result was `different`. Simple CASE is defined in terms of equality, so this leaves the newly aligned `1 = 1.0` behavior internally inconsistent. Route simple CASE through the same three-valued equality helper and add integer/float plus null/nested-value regressions.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
```diff
@@ -1782,36 +1855,72 @@ class QueryPipeline private constructor(
...
+    @Suppress("UNCHECKED_CAST")
+    private fun matchedRelationships(bindings: Map<String, Any?>): Set<Edge> =
+        (bindings[INTERNAL_MATCHED_PATH_SEGMENTS_KEY] as? List<MatchedPathSegment>)
+            .orEmpty()
+            .flatMapTo(linkedSetOf()) { it.relationships }
+
+    private fun unavailableRelationships(
+        rel: PatternElement.RelationshipPattern,
+        bindings: Map<String, Any?>
+    ): Set<Edge> = matchedRelationships(bindings) - relationshipBinding(rel.variable?.let(bindings::get))
```
johnsonlee: **[P1 blocking] Do not exempt an already-used relationship just because the same variable is repeated.** On this exact head, a two-node graph with one undirected `DATAFLOW` edge returns a row for `MATCH (a {value: 1})-[r:DATAFLOW]-(b)-[r:DATAFLOW]-(c) RETURN c.value`; the only result traverses that single relationship twice. openCypher's default relationship-unique match mode still applies when a relationship variable is repeated (the repeated-variable pattern has no solution). The subtraction here removes `r` from the unavailable set on the second segment, which explicitly permits the invalid reuse. A separate second `MATCH` may reuse `r`; two occurrences in the same graph pattern may not.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
```diff
@@ -1508,33 +1542,62 @@ class QueryPipeline private constructor(
...
 
-        // If the pattern has a path variable, bind it to the matched path
-        if (pattern.pathVariable != null) {
-            return matches.map { bindings ->
-                val path = buildPathRepresentation(pattern, bindings)
-                bindings.toMutableMap().apply {
+        return matches.map { bindings ->
+            val path = pattern.pathVariable?.let { buildPathRepresentation(pattern, bindings) }
+            bindings.toMutableMap().apply {
+                remove(INTERNAL_MATCHED_PATH_SEGMENTS_KEY)
```
johnsonlee, line 1865: **[P1 blocking] Preserve relationship uniqueness across every comma-separated path in one `MATCH`.** A falsifiable probe on a two-node/single-edge graph—`MATCH (a {value: 1})-[r]-(b), (b)-[s]-(c) RETURN c.value`—returns `1` on this head by reusing the only relationship. The equivalent query with two separate `MATCH` clauses should return `1`, but the comma form is one graph pattern and must not reuse a relationship. Clearing the segment state at the end of each `CypherPattern` means `executeMatch` cannot enforce uniqueness across the clause's next comma-separated path. Carry the matched-relationship state across constituent paths, and reset it only after the whole MATCH clause.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/ExpressionEvaluator.kt
```diff
@@ -447,39 +456,58 @@ private fun appendWithCancellation(
...
+    left == null || right == null -> null
+    left is Number && right is Number -> left.toDouble() == right.toDouble()
```
johnsonlee: **[P1 blocking] Avoid `Double` coercion for integer equality.** Exact-head probe `RETURN 9007199254740993 IN [9007199254740992] AS member` returns `true`; these are distinct integer values and openCypher `IN` must return `false`. Both `Long`s collapse to the same IEEE-754 double here. This is a regression from the base membership implementation, which compared the list elements exactly. Please compare integral values without precision loss (and reuse that equality consistently for `=`, `IN`, lists/maps, and simple CASE).

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
```diff
@@ -1582,50 +1647,29 @@ class QueryPipeline private constructor(
...
+        val path = buildList {
+            if (firstNode != null) add(firstNode)
+            matchedSegments.forEach { addAll(it.tail) }
         }
-        if (!qualified) return path
+        return pathValue(path)
+    }
 
+    private fun pathValue(path: List<Any>): Any {
+        if (!qualified) return PathFinder.Path(path.filterIsInstance<Node>(), path.filterIsInstance<Edge>())
```
johnsonlee, line 2002: **[P1 blocking] Teach `length()` about the new unqualified path representation.** This branch now returns `PathFinder.Path`, but `CypherFunctions.size` only recognizes `String`, `List`, and `QualifiedPath`. On this exact head, `MATCH p=(a:IntConstant {value: 1})-[:DATAFLOW*0]->(a) RETURN length(p), nodes(p), relationships(p)` returns `null` for `length(p)` even though `nodes(p)` has one item and `relationships(p)` is empty; the expected length is `0`. Nonzero named paths fail the same way. Add `PathFinder.Path -> value.edges.size` so single-graph named paths retain the public path-function contract.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
```diff
@@ -164,20 +177,31 @@ class QueryPipeline private constructor(
...
+        clauses: List<CypherClause>,
+        parameters: Map<String, Any?>,
+        workTracker: CypherWorkTracker?
     ): CypherResult {
-        if (workTracker == null) return executeWithActiveBudget(clauses)
         val previousTracker = activeWorkTracker.get()
-        activeWorkTracker.set(workTracker)
+        val previousParameters = activeParameters.get()
+        if (workTracker == null) activeWorkTracker.remove() else activeWorkTracker.set(workTracker)
+        activeParameters.set(parameters.toMap())
```
johnsonlee, line 217: **[P1 blocking] Restore the existing method-level latency gate.** Exact-head artifact `benchmark-method-105-1` rejects three independent rows after reverse-order confirmation: `aggregationCountGroupBy` is `+31.4%` initially and `+22.3%` on confirmation; `nodeMatchWithWhere` is `+25.4%` / `+29.8%`; and `withPipeline` is `+256.2%` / `+244.8%`. The latter query performs numeric `ORDER BY`, so the new per-comparison `BigDecimal(a.toString())` path is one concrete changed hot path; the unconditional per-query parameter `ThreadLocal` setup is another shared path worth checking for the other rows. These are already falsified by the repository's paired gate, not a request for extra capacity evidence. Please preserve the new semantics while bringing the exact head back through the existing gate; a passing rerun is sufficient.

johnsonlee (commented):
Exact-head method-level benchmark gate has three confirmed regressions; inline evidence attached.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/ExpressionEvaluator.kt
johnsonlee: Verified fixed on exact head `c496b90`: the end-to-end probe `RETURN 9007199254740993 IN [9007199254740992] AS member` now returns `false`, while the added same-runtime numeric fast path preserves the cross-type coercion tests. This resolves the scoped `IN` regression; the separate general `=` precision issue is tracked in the equality/CASE thread.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/ExpressionEvaluator.kt
johnsonlee: Rechecked exact head `c496b90`: the simple CASE probe still returns `different`. The updated helper also still makes `RETURN 9007199254740993 = 9007199254740992 AS equal` return `true`, because distinct same-runtime `Long` values fall through to `toDouble()` and collapse above 2^53. Please make integral equality exact in `cypherEquals` and route simple CASE through it; the focused `IN` thread is resolved separately.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/ExpressionEvaluator.kt
johnsonlee: Correction after extending the exact-head probe: this thread is not fully fixed, so I am reopening it. `c496b90` fixes the same-runtime `Long` case, but `RETURN 9007199254740993 IN [9007199254740992.0] AS member` still returns `true`; the distinct integer and floating value collapse in the remaining cross-type `toDouble()` branch. Please use an exact mixed numeric comparison while retaining `2 IN [2.0] = true`, then cover both same-type and cross-type values around 2^53.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/ExpressionEvaluator.kt
johnsonlee: Verified fixed on exact head `c496b90` by the existing repository gate. `budgetedListMembership` is now base `780.682 us/op` vs candidate `817.936 us/op` (`+4.8%`, below the 15% limit), and `budgetedListConcatenation` is `-9.1%`; `budgeted-collection-latency` is terminal SUCCESS. No additional performance proof is needed, so I am resolving this thread.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
johnsonlee: Exact-head `c496b90` rerun confirms this blocker is unchanged despite the membership optimization. `aggregationCountGroupBy` is now `+34.6%` initially / `+29.7%` confirmation, `nodeMatchWithWhere` is `+28.4%` / `+20.7%`, and `withPipeline` is `+273.0%` / `+259.6%`; required `method-level-benchmarks` is terminal FAILURE. Keeping this thread unresolved until the existing gate passes.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/ExpressionEvaluator.kt
```diff
@@ -447,39 +456,155 @@ private fun appendWithCancellation(
...
         index++
     }
     checkCancelled()
-    return false
+    return if (containsNullComparison) null else false
+}
+
+private fun cypherMembershipEquals(element: Any, candidate: Any?): Boolean? = when {
+    candidate == null -> null
+    element is List<*> && candidate is List<*> -> cypherListsEqual(element, candidate)
```
johnsonlee: **[P1 blocking] Apply the exact numeric membership rule recursively inside lists and maps.** The new same-runtime-class guard is only used for top-level numeric candidates; these two branches delegate back to `cypherEquals`, which still coerces distinct `Long`s to `Double`. On exact head `c496b90`, both `RETURN [9007199254740993] IN [[9007199254740992]] AS member` and `RETURN {value: 9007199254740993} IN [{value: 9007199254740992}] AS member` return `true` (expected `false`). The small nested/null tests pass, so the defect is specifically the recursive 2^53 boundary. Please use the same exact numeric comparator at every recursion level.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/ExpressionEvaluator.kt
```diff
@@ -447,39 +456,155 @@ private fun appendWithCancellation(
...
 
-private fun containsRandomAccessList(list: List<*>, element: Any?, checkCancelled: () -> Unit): Boolean {
+@Suppress("ReturnCount")
+private fun evaluateRangeMembership(
+    list: CypherRangeValues,
+    element: Number,
+    checkCancelled: (() -> Unit)?
+): Boolean {
+    val target = element.toLong()
+    if (element.toDouble() != target.toDouble()) {
```
johnsonlee: **[P1 blocking] Reject out-of-range floating values before converting them to `Long`.** `Number.toLong()` saturates, and the subsequent `Double` check cannot distinguish `Long.MAX_VALUE` from 2^63 because `Long.MAX_VALUE.toDouble()` rounds to 2^63. Consequently, on exact head `c496b90`, `RETURN 9223372036854775808.0 IN range(9223372036854775807, 9223372036854775807) AS member` returns `true` instead of `false`. Check finite/integral/range bounds with an exact conversion (or the shared unlimited-precision numeric comparator) before using the `Long` fast path.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
johnsonlee: Root-cause refinement from a same-machine base/head allocation run: `aggregationCountGroupBy` allocates `154,640 -> 224,729 B/op`, `nodeMatchWithWhere` `210,961 -> 281,001 B/op`, and `withPipeline` `285,601 -> 355,695 B/op`—about `+70 KiB/op` on all three. The common source is `matchPattern` unconditionally cloning every returned binding at lines 1559–1569 to remove path-internal keys, even for a node-only pattern where `navigate=false`, `trackSegments=false`, and none of those keys can exist; these benchmarks scan 500 nodes, so that adds one map copy per row. For `withPipeline`, the new allocating `BigDecimal(a.toString())` numeric ORDER BY path adds the remaining cost; a short local same-type numeric fast-path ablation reduced this head's row from `188.9` to `138.0 us/op` (diagnostic only, not gate evidence). My earlier ThreadLocal note is not supported as the main cause. The authoritative merge criterion remains the paired repository gate.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/ExpressionEvaluator.kt
johnsonlee: Rechecked exact head `1aca072`: both nested exactness probes still fail. List and map membership each treat `9007199254740993` as equal to `9007199254740992`; keeping unresolved.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/ExpressionEvaluator.kt
johnsonlee: Rechecked exact head `1aca072`: `RETURN 9007199254740993 IN [9007199254740992.0] AS member` still returns `true`. Mixed numeric membership still loses precision; keeping unresolved.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
johnsonlee: Rechecked exact head `1aca072`: the zero-hop named-path probe still returns a non-`0`/null value for `length(p)` while `nodes(p)` has one item and `relationships(p)` is empty. The unqualified path-function blocker remains.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
johnsonlee: Rechecked exact head `1aca072`: the one-edge probe still returns a row for `MATCH (a {value: 1})-[r:DATAFLOW]-(b)-[r:DATAFLOW]-(c)`, so the same relationship is still reused within one path. Keeping unresolved.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/ExpressionEvaluator.kt
johnsonlee: Rechecked exact head `1aca072`: both focused probes still fail. `CASE 1 WHEN 1.0` returns `different`, and `9007199254740993 = 9007199254740992` returns `true`. This commit only changes allocation and ORDER BY fast paths; the equality/CASE blocker remains.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
johnsonlee: Rechecked exact head `1aca072`: the one-edge comma-pattern probe still returns a row for `MATCH (a {value: 1})-[r]-(b), (b)-[s]-(c)`. Relationship uniqueness is still reset between paths in one MATCH; keeping unresolved.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/ExpressionEvaluator.kt
johnsonlee: Rechecked exact head `1aca072`: `9223372036854775808.0 IN range(9223372036854775807, 9223372036854775807)` still returns `true`. The saturating conversion blocker remains.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
johnsonlee: Verified fixed by the repository gate on exact head `1aca072`. Artifact `benchmark-method-105-1` has terminal `method-level-benchmarks` SUCCESS: the three original blocking rows are now `aggregationCountGroupBy +5.3%`, `nodeMatchWithWhere +6.2%`, and `withPipeline +7.5%`, all below the 15% limit. The final paired status is `passed: true`; its only confirmation row was classified as noise by the existing gate. Per the agreed benchmark safety criterion, no additional performance proof is required. Resolving this performance thread; the independent correctness threads remain open.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
```diff
@@ -2375,8 +2489,8 @@ class QueryPipeline private constructor(
                 for (item in clause.items) {
                     val leftValue = evaluateOrderValue(item.expression, left)
                     val rightValue = evaluateOrderValue(item.expression, right)
-                    val comparison = compareNullable(leftValue, rightValue)
-                    if (comparison != 0) return@Comparator if (item.ascending) comparison else -comparison
+                    val comparison = compareOrderValues(leftValue, rightValue, item.ascending)
```
johnsonlee, line 3053: **[P1] Evaluate ORDER BY against the pre-projection row when the sort key is not returned**

The new comparator cannot restore a sort expression that projection has already discarded: `RETURN` is executed before this clause, and `evaluateOrderValue(n.value, row)` therefore sees no `n` binding when only `n.id` was projected. On exact `1aca072`, with IntConstants `(id=1,value=2)` then `(id=2,value=1)`, `MATCH (n:IntConstant) RETURN n.id ORDER BY n.value` returns `[1,2]`; ascending `n.value` requires `[2,1]`. The base has the same gap, but this PR changes and claims P0 `ORDER BY` alignment, and the added tests only sort returned expressions/aliases. Preserve the pre-projection bindings (unless DISTINCT/aggregation makes the key unavailable) or evaluate hidden sort values during projection, and add a result-order assertion with insertion order opposite to the sort key.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
```diff
@@ -1842,8 +1962,8 @@ class QueryPipeline private constructor(
         if (pattern.labels.size > 1) {
             val nodeLabels = CypherFunctions.labels(value).map { it.lowercase() }.toSet()
             val allMatch = pattern.labels.all { label ->
-                val labelClass = NodePropertyAccessor.resolveNodeLabel(label)
-                labelClass.isInstance(node) || label.lowercase() in nodeLabels
+                val labelClass = NodePropertyAccessor.resolveNodeLabelOrNull(label)
```
johnsonlee, line 2376: **[P1] Apply Cypher equality/null semantics to inline pattern properties**

Inline node and relationship properties still compare with Kotlin `==` at lines 1974/2008, bypassing the new three-valued/numeric equality behavior. On exact `1aca072`, a graph containing IntConstants `(id=1,value=2)`, `(id=2,value=1)` and one DATAFLOW edge gives: `(n:IntConstant {missing: null})` matches both nodes; `(n:IntConstant {value: 2.0})` matches none; and `[:DATAFLOW {missing: null}]` matches the edge. Cypher property-map predicates are equality predicates: a missing/null property must not pass WHERE-style matching, while integer `2` must equal floating `2.0`. This also predates the branch, but leaves the PR’s null/comparison P0 contract inconsistent between `WHERE n.value = ...` and the equivalent inline pattern. Route both constraint paths through the same Cypher equality helper and accept only an exact `true` result.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/CypherFunctions.kt
```diff
@@ -196,12 +196,12 @@ object CypherFunctions {
     // ========================================================================
 
     fun aggregate(name: String, values: List<Any?>): Any? = when (name.lowercase()) {
-        FUNCTION_COUNT -> values.size.toLong()
+        FUNCTION_COUNT -> values.count { it != null }.toLong()
```
johnsonlee, line 199: **[P1] Deduplicate aggregate/group values with Cypher numeric equality**

The aggregate now correctly ignores nulls, but its upstream DISTINCT filter and grouping maps still use Kotlin collection equality, which disagrees with the PR’s Cypher rule that numerically equal values compare by value across runtime types. On exact `1aca072`, `UNWIND [1, 1.0] AS x RETURN count(DISTINCT x) AS c` returns `2` instead of `1`; `RETURN DISTINCT x` likewise retains both values, and grouping by `x` creates two groups. Fixing the existing `cypherEquals` precision threads alone will not fix these hash/set paths. This behavior also exists on base, but it leaves the claimed aggregate/comparison alignment incomplete. Use a normalized Cypher value key/equality for DISTINCT and group keys, including nested list/map values, and cover cross-runtime numeric duplicates.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
johnsonlee: Rechecked exact merge head `05bcf900fa92`. The one-edge query with repeated `r` still returns a row, so one relationship is reused twice inside the same path. The post-merge focused test fails; keeping unresolved.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/ExpressionEvaluator.kt
johnsonlee: Rechecked exact merge head `05bcf900fa92` with the focused post-merge probe. `CASE 1 WHEN 1.0` still returns `different`, and the same test class also confirms `9007199254740993 = 9007199254740992` is still true. The #102 merge does not change either equality path; keeping unresolved.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
johnsonlee: Rechecked exact merge head `05bcf900fa92`. The one-edge comma-pattern query still returns a row by reusing the relationship across the two constituent paths in one MATCH. The post-merge focused test fails; keeping unresolved.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
johnsonlee: Rechecked exact merge head `05bcf900fa92`. The zero-hop named-path probe still does not return `length(p) = 0` while nodes has one item and relationships is empty. The post-merge focused test fails; keeping unresolved.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
johnsonlee: Rechecked exact merge head `05bcf900fa92`. With insertion order `(id=1,value=2)`, `(id=2,value=1)`, `MATCH (n:IntConstant) RETURN n.id ORDER BY n.value` still returns `[1,2]` instead of `[2,1]`. The #102 streaming/order changes do not fix the ordinary pre-projection binding path; keeping unresolved.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
johnsonlee: Rechecked exact merge head `05bcf900fa92`. The focused inline-property test still fails: missing/null node properties match, integer `1` does not match inline `1.0`, and a missing/null relationship property matches. The #102 merge does not align these constraints with Cypher equality; keeping unresolved.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/ExpressionEvaluator.kt
johnsonlee: Rechecked exact merge head `05bcf900fa92`. `9007199254740993 IN [9007199254740992.0]` still evaluates to true, so mixed numeric membership still loses precision. The post-merge focused test fails; keeping unresolved.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/ExpressionEvaluator.kt
johnsonlee: Rechecked exact merge head `05bcf900fa92`. `9223372036854775808.0 IN range(9223372036854775807, 9223372036854775807)` still returns true. Saturating float-to-Long conversion remains observable; keeping unresolved.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/CypherFunctions.kt
johnsonlee: Rechecked exact merge head `05bcf900fa92`. The aggregate probe still fails: `count(DISTINCT x)` for `[1, 1.0]` is not 1, and RETURN DISTINCT/grouping still retain separate numeric keys. The #102 merge does not change normalized group/distinct equality; keeping unresolved.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/ExpressionEvaluator.kt
johnsonlee: Rechecked exact merge head `05bcf900fa92`. Both nested list and nested map probes still treat the distinct integers around 2^53 as equal. The post-merge focused test fails; keeping unresolved.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/ExpressionEvaluator.kt
johnsonlee: Fixed in `2d4aaf6`. Simple CASE now detects the syntactic test expression (`expr.test != null`) and accepts only `cypherEquals(test, when) == true`. End-to-end regressions cover `CASE 1 WHEN 1.0` and a null simple-CASE test; the full repository `check` passes.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
johnsonlee: Fixed in `2d4aaf6`. Relationship uniqueness now uses an independent MATCH-scoped state with separate reserved and already-used sets. A relationship becomes unavailable after its first traversal even when the same variable is repeated. The exact two-node/one-edge repeated-`r` probe now returns no rows; separate MATCH clauses still permit reuse.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
johnsonlee: Fixed in `2d4aaf6`. Used relationships are now carried across every comma-separated path and cleared only after the whole MATCH/OPTIONAL MATCH clause. Named-path segments remain separate, so path reconstruction is not contaminated. Regressions cover single-hop and variable-length first paths, plus the allowed two-separate-MATCH control.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/ExpressionEvaluator.kt
johnsonlee: Fixed in `2d4aaf6`. Numeric equality no longer falls through to lossy `Double` coercion: integral values compare exactly, mixed finite numbers use decimal comparison, and non-finite values are handled explicitly. The 2^53 same-type and mixed integer/float probes now return false while `1 = 1.0` remains true.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
johnsonlee: Fixed in `2d4aaf6`. `length()`/`size()` now handles `PathFinder.Path` via `edges.size`. End-to-end named-path regressions cover zero-hop length 0 with one node/no relationships and a nonzero path length 1.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/ExpressionEvaluator.kt
johnsonlee: Fixed in `2d4aaf6`. Top-level and recursive membership now share the same exact `cypherEquals` implementation. Nested list and map probes around 2^53 both return false, while recursive cross-runtime numeric equality for ordinary values remains supported.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/ExpressionEvaluator.kt
johnsonlee: Fixed in `2d4aaf6`. The range fast path now requires an exact finite `Long` conversion; fractional, non-finite, and out-of-range values are rejected before membership lookup. The 2^63 probe returns false and `2.0 IN range(1, 3)` remains true.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
johnsonlee: Fixed in `2d4aaf6`. For non-DISTINCT/non-aggregate projection, hidden ORDER BY values are evaluated against pre-projection bindings plus projected aliases, retained internally, consumed by sorting, then removed. The insertion-order-opposite `RETURN n.id ORDER BY n.value` regression now produces the required value order. Simple alias-only ORDER BY keeps the allocation-free existing path.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
johnsonlee: Fixed in `2d4aaf6`. Inline node and relationship property constraints now require `cypherEquals(actual, expected) == true`. Integer property `2` matches `2.0`; missing-property/null constraints match neither nodes nor relationships. End-to-end regressions cover all three cases.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/CypherFunctions.kt
johnsonlee: Fixed in `2d4aaf6`. A recursive, exact Cypher value key now drives RETURN DISTINCT, aggregate DISTINCT, grouping, streaming/limited distinct paths, ordered distinct retention, fast property distinct, and UNION. Tests cover scalar/list/map cross-runtime numeric duplicates, grouping/count(DISTINCT), UNION, and the non-collapsing 2^53 boundary. The full repository check and local focused JMH smoke pass.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/ExpressionEvaluator.kt
johnsonlee: Verified fixed on exact head `2d4aaf6c0ccc`. The independent probe now returns `same` for `CASE 1 WHEN 1.0` and false for equality between the distinct integers around 2^53. The committed equality/CASE regressions and complete Cypher suite pass. Resolving this thread.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
johnsonlee: Verified fixed on exact head `2d4aaf6c0ccc`. The independent one-edge repeated-`r` query now returns no rows, while the committed regression also preserves reuse across separate MATCH clauses. Complete Cypher and bounded relationship-memory tests pass. Resolving this thread.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
johnsonlee: Verified fixed on exact head `2d4aaf6c0ccc`. The independent zero-hop named path now reports length 0 with one node and no relationships; the committed nonzero-path case also returns length 1. Resolving this thread.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
johnsonlee: Verified fixed on exact head `2d4aaf6c0ccc`. The independent comma-pattern one-edge query now returns no rows; relationship state is retained across constituent paths and cleared at the MATCH boundary. The committed single-hop and variable-path cases pass. Resolving this thread.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/ExpressionEvaluator.kt
johnsonlee: Verified fixed on exact head `2d4aaf6c0ccc`. `9007199254740993 IN [9007199254740992.0]` now returns false while small cross-type equality remains true. The independent and committed precision probes pass. Resolving this thread.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
johnsonlee: Verified fixed on exact head `2d4aaf6c0ccc`. Independent inline node and relationship probes now apply Cypher equality exactly: numeric `1` matches `1.0`, while missing/null properties match neither nodes nor edges. Resolving this thread.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/ExpressionEvaluator.kt
johnsonlee: Verified fixed on exact head `2d4aaf6c0ccc`. The out-of-range double no longer matches the Long.MAX_VALUE range, while exact integral doubles still use the range fast path. Independent and committed tests pass. Resolving this thread.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/CypherFunctions.kt
johnsonlee: Verified fixed on exact head `2d4aaf6c0ccc`. Independent `count(DISTINCT)`, RETURN DISTINCT, and grouping probes now collapse `1` with `1.0`; recursive list/map, UNION, and precision-boundary regressions are also committed and pass. Resolving this thread.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/ExpressionEvaluator.kt
johnsonlee: Verified fixed on exact head `2d4aaf6c0ccc`. The independent nested list and map membership probes around 2^53 now both return false, and recursive normalized-key tests pass. Resolving this thread.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
johnsonlee: Verified fixed on exact head `2d4aaf6c0ccc`. The independent non-projected sort probe now returns `[2,1]` for ascending `n.value`; hidden order values are removed after sorting, and the complete Cypher suite passes. Resolving this thread.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
```diff
@@ -44,9 +45,28 @@ private const val SINGLE_HOP_LIMIT_QUERY_CLAUSES = 3
...
+private const val INTERNAL_RELATIONSHIP_MATCH_STATE_KEY = "\u0000graphite.relationshipMatchState"
+private const val INTERNAL_ORDER_VALUES_KEY = "\u0000graphite.orderValues"
+
+private data class MatchedPathSegment(
+    val tail: List<Any>,
+    val relationships: List<Edge>
+)
+
+private data class RelationshipMatchState(
+    val reserved: Set<Edge>,
```
johnsonlee: **[P1 blocking] Qualify relationship identity by graph before enforcing MATCH-wide uniqueness.**

`RelationshipMatchState` stores raw `Edge` values, but cross-graph relationship identity is `graphId + edge` (as `QualifiedEdge` already models). Two different graphs can legitimately contain equal local edges; after the first comma-separated pattern uses one, `unavailableRelationships()` currently excludes the equal edge from every graph.

Exact-head reproduction on `2d4aaf6c0ccc`: build `orders` and `billing` with the same local `DataFlowEdge(1, 2, ASSIGN)`, then run:

```cypher
MATCH (a)-[r1:DATAFLOW]->(b), (c)-[r2:DATAFLOW]->(d)
RETURN graphId(a) AS firstGraph, graphId(c) AS secondGraph
ORDER BY firstGraph, secondGraph
```

The two relationships are globally distinct, so the valid results are `billing -> orders` and `orders -> billing`; the current code returns `[]`. I reproduced this as a focused JUnit test (expected 2 rows, actual 0). Track reserved/used identities with the owning graph namespace and only project them back to raw edges for the source graph being traversed; add this cross-graph regression alongside the single-graph uniqueness cases.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
```diff
@@ -2531,20 +2711,48 @@ class QueryPipeline private constructor(
         }
     }
 
+    private fun groupKey(
+        items: List<ReturnItem>,
+        groupByIndices: List<Int>,
+        row: Map<String, Any?>
+    ): Any = cypherValueKey(groupByIndices.map { index -> evaluator.evaluate(items[index].expression, row) })
```
johnsonlee: **[P1 blocking] Avoid per-row recursive key normalization regressing GROUP BY.**

The required exact-head benchmark gate fails this hot path, including its reverse-order confirmation:

- initial: `aggregationCountGroupBy` 80.093 → 111.060 µs/op (**+38.7%**)
- confirmation: 78.508 → 111.966 µs/op (**+42.6%**)

The benchmark groups `CallSiteNode.callee_class` (a string), so every row now pays for `groupByIndices.map { ... }` plus recursive `cypherValueKey(...)`/wrapper allocation even though native string equality already has the required semantics. Preserve exact cross-runtime numeric/nested-value grouping, but add a non-lossy fast path for already canonical scalar keys (and ideally avoid redundant outer wrappers). No additional performance proof is needed: this is resolved when the existing `benchmark-regression-gate` passes on the fixing head.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/ExpressionEvaluator.kt
```diff
@@ -167,22 +178,20 @@ class ExpressionEvaluator(
...
-            "<=" -> compareValues(left, right) <= 0
-            ">=" -> compareValues(left, right) >= 0
-            else -> throw CypherException("Unknown comparison: ${expr.op}")
+            "=" -> cypherEquals(left, right)
+            "<>" -> cypherEquals(left, right)?.not()
+            else -> if (left == null || right == null) {
+                null
+            } else {
+                when (expr.op) {
+                    "<" -> compareValues(left, right) < 0
```
johnsonlee, line 188: **[P1 blocking] Use exact numeric semantics for relational comparisons too.**

This branch still delegates `<`/`>`/`<=`/`>=` to `compareValues()`, whose numeric case converts both operands to `Double`. That contradicts the exact equality/DISTINCT/ORDER BY semantics added in this PR and collapses distinct integral values above 2^53.

Exact-head reproduction on `2d4aaf6c0ccc`:

```cypher
RETURN 9007199254740993 > 9007199254740992 AS greater,
       9007199254740992 < 9007199254740993 AS less
```

Expected `{greater=true, less=true}`; the focused JUnit probe returns `{greater=false, less=false}`. Reuse a shared exact numeric comparator (integral `Long` comparison, finite mixed-number decimal comparison, explicit non-finite handling) for predicate comparisons and add the boundary query as a regression.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
```diff
@@ -1642,7 +1701,7 @@ class QueryPipeline private constructor(
         inputRows: List<Map<String, Any?>>,
         limit: Int? = null
     ): List<Map<String, Any?>> {
-        var rows = inputRows
+        var rows = if (patterns.size > 1) inputRows.map(::startRelationshipMatchState) else inputRows
```
johnsonlee, line 1772: **[P1 blocking] Do not apply the final LIMIT to intermediate comma-pattern joins.**

Once MATCH-wide relationship uniqueness is tracked here, truncating each constituent pattern with `earlyLimit` becomes observably incorrect: the first retained binding may consume the only relationship that satisfies a later constrained pattern, while a discarded first-pattern binding would produce a valid completed row.

Exact-head reproduction on `2d4aaf6c0ccc`: a graph has edges `1 -> 2` and `3 -> 4` (in that encounter order), then:

```cypher
MATCH (a)-[first:DATAFLOW]->(b),
      (c:IntConstant {value: 1})-[second:DATAFLOW]->(d:IntConstant {value: 2})
RETURN a.value AS firstSource, c.value AS secondSource
LIMIT 1
```

The valid completed match is `{firstSource=3, secondSource=1}`; a focused JUnit probe expects that row but current code returns `[]`. Disable early LIMIT pushdown for multi-pattern MATCH, or make it join-aware and apply the cap only to completed pattern combinations; add this constrained two-pattern case as a regression.

johnsonlee: The same unsafe pushdown also crosses **separate MATCH clauses**, so fixing only `patterns.size > 1` would leave the root cause unresolved.

Exact-head focused probe:

```cypher
MATCH (a:IntConstant)
MATCH (a)-[:DATAFLOW]->(b)
RETURN a.value AS value
LIMIT 1
```

With nodes encountered as a dead-end value `1`, then source value `2 -> 3`, the valid result is `[{value=2}]`; current head truncates the first MATCH to the dead-end row and returns `[]`. `computeEarlyLimit()` currently treats another `CypherClause.Match` between the first MATCH and LIMIT as safe. The fix needs to reject pushdown across both multiple patterns in one MATCH and any later MATCH join (or apply LIMIT only after fully completed joined rows). Please cover both probes.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
johnsonlee: Fixed in 84afd4d. RelationshipMatchState now stores QualifiedEdge identities (graphId + edge); unavailableRelationships projects only identities owned by the source graph being traversed. Added the exact orders/billing collision regression and it returns billing→orders plus orders→billing. Full cypher check and the repository-wide check both pass.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
johnsonlee: Fixed in 84afd4d. DISTINCT/grouping now try native exact keys first and normalize only numeric-containing misses; the single-column GROUP BY path also avoids per-row List allocation. Same-host focused JMH: aggregationCountGroupBy base 33.593 us/op, candidate 32.272 us/op; returnDistinct candidate 110.415 us/op. Leaving this thread open until the exact-head benchmark gate on e39ff5a passes.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/ExpressionEvaluator.kt
johnsonlee: Fixed in e39ff5a. Predicate comparisons and mixed-number ORDER BY now share compareCypherNumbers: integral values compare as Long, finite mixed values as BigDecimal, and non-finite values explicitly as Double. Added the 2^53 boundary query plus finite/non-finite comparator regressions; cypher check passes.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
johnsonlee: Fixed in e39ff5a. computeEarlyLimit now rejects an initial MATCH with multiple patterns or inline WHERE and rejects every later MATCH (including optional) before LIMIT. Added both dead-end reproductions: comma-joined patterns and separate MATCH clauses now retain the valid value=2 row. Cypher check passes.

johnsonlee: Verified on exact head `e39ff5ae5e8049e37fd9ec1523983ba04204d91c`: both independent variants now pass—the constrained comma-pattern join returns `{firstSource=3, secondSource=1}`, and the separate-MATCH query skips the dead end and returns `{value=2}`. The committed regression also covers both pushdown boundaries. Resolving.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/ExpressionEvaluator.kt
johnsonlee: Verified on exact head `e39ff5ae5e8049e37fd9ec1523983ba04204d91c`: the independent boundary query now returns `{greater=true, less=true}`, and the focused committed regression passes. Resolving.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
johnsonlee: Verified on exact head `e39ff5ae5e8049e37fd9ec1523983ba04204d91c`: the independent two-graph probe now returns both valid cross-graph permutations, and the committed regression covers the same graph-qualified identity case. Resolving.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
johnsonlee: Verified from the exact-head `e39ff5ae5e8049e37fd9ec1523983ba04204d91c` method-level gate artifact: `aggregationCountGroupBy` is 57.898 us/op on base versus 57.317 us/op on PR (-1.0%), so the existing benchmark gate passes this hot path. Resolving.

johnsonlee: Exact-head gate is now green on e39ff5a: aggregationCountGroupBy 57.898 → 57.317 us/op (-1.0%, PASS), returnDistinct 141.484 → 154.067 us/op (+8.9%, PASS). Resolving as requested.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
```diff
@@ -1928,10 +2033,14 @@ class QueryPipeline private constructor(
...
         candidate: Any,
         nodePattern: PatternElement.NodePattern,
-        existingBindings: Map<String, Any?>
+        existingBindings: Map<String, Any?>,
+        navigate: Boolean,
+        trackSegments: Boolean
     ): Map<String, Any?>? {
         if (!matchesNodeConstraints(candidate, nodePattern, existingBindings)) return null
         return existingBindings.toMutableMap().apply {
+            if (trackSegments) putIfAbsent(INTERNAL_PATH_START_NODE_KEY, candidate)
```
johnsonlee, line 2090: **[P1 blocking] Restore the required Cypher capacity gate before merging.** On exact head `e6305be98a0e`, [`cypher-capacity-gate`](https://github.com/johnsonlee/graphite/actions/runs/33287428130/job/99193448193) completed both base and candidate measurements and failed on performance, not on 429s or missing output. The candidate preserved the functional envelope (4 active queries, 3 budget-exceeded responses, 1 cancellation, and successful recovery), but `processCpuNanos` increased from `5.46e9` to `6.47e9` (**+18.50%**) and `tailLatencyNanos` from `1.719123342e9` to `1.982926470e9` (**+15.35%**), both beyond the 15% gate. This benchmark drives four concurrent full node scans through this per-candidate binding/evaluation path, so the current authoritative evidence says the semantic changes introduced a cross-cutting scan regression. Please optimize it or rerun the required gate if this was runner variance; a green gate is sufficient and no additional proof is requested.

johnsonlee (commented):
Exact-head benchmark review found one blocking regression.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
johnsonlee: Control measurement: the concurrently running PR #104 used the same capacity harness and the same base→candidate order, while changing only workflow/gate code (runtime Cypher code is unchanged). Its artifact measured CPU `5.873e9 → 5.797e9` (**-1.31%**) and tail latency `1.7365e9 → 1.7848e9` (**+2.78%**). That control does not show a systematic candidate-second slowdown and makes the #105 exact-head +18.50% CPU / +15.35% tail result materially less consistent with simple ordering bias. The existing required gate remains the decision criterion.

johnsonlee: Fixed in 1d6c928 by streaming unbounded filtered single-node MATCH queries through predicate evaluation and projection instead of eagerly retaining one binding map per scanned node. The authoritative rerun on exact head 1d6c928 passed cypher-capacity-gate: processCpuNanos 6.020e9 -> 5.760e9 (-4.32%), tailLatencyNanos 1.842606532e9 -> 1.723825142e9 (-6.45%), and residentSetAfterBytes 1.084865195e9 -> 1.048989696e9 (-3.31%). The functional envelope remains 4 active queries, 3 budget exceeded, 1 cancellation, and successful recovery.

johnsonlee: Verified on exact head `1d6c9283ccd5`: the required [`cypher-capacity-gate`](https://github.com/johnsonlee/graphite/actions/runs/33289710079/job/99199101296) now passes with valid benchmark output. The functional envelope remains intact (active peak 4, 3 budget-exceeded responses, 1 cancellation, recovery 1); aggregate wall time improves `1915.8 -> 1797.4 ms/op` (-6.2%), CPU improves `18.06e9 -> 17.28e9 ns` (-4.3%), and tail latency improves `5.528e9 -> 5.171e9 ns` (-6.4%). No 429 or missing-result path was accepted. A clean exact-SHA `./gradlew :cypher:test --rerun-tasks --no-daemon` also passes. The reported regression is no longer present, so resolving.

PR #106, opened 2026-08-30
Accelerate benchmark gate and publish a Pages observatory

johnsonlee:
## Summary

- remove the ~30 minute PR gate bottleneck without deleting benchmark coverage
- replace the PR-path pre-PR-95 known-bad run with a pinned known-good anchor plus the exact current base; preserve the expensive historical proof in a scheduled/manual nightly workflow
- split the 33-case Method compatibility matrix into 12 balanced scenario shards
- build Explorer and wrapped-query JMH JARs once per revision, verify checksums, and fan them out through retry-stable artifacts
- classify every report using the PR #104 product/gate-system coverage model and show known gaps separately from the run verdict
- publish a self-contained Benchmark Observatory to GitHub Pages after every `main` push, with exact integration provenance, rolling history, paired PR evidence, search/filter controls, and responsive layout
- add a standalone release-tag workflow that compares each `v*` tag with its semantic predecessor, publishes a self-contained 7+2 release observatory at [`/releases/`](https://johnsonlee.github.io/graphite/releases/), and retains protocol-aware history without blocking publishing

## Bottleneck evidence and measured result

The latest successful gate before this change took about 29 minutes. The `real-36` wrapped-query shard was the critical path: the known-bad revision costs roughly 255 seconds per operation, and the harness performed one setup scan plus one warmup and three measured scans. Method compatibility at 36 graphs was the next long pole, while the matrix repeated dozens of identical Gradle JMH builds across jobs.

The final exact-head optimized PR run finished in **7m57s**, a **73.5% wall-clock reduction** from the ~30 minute baseline. The large-corpus job was the new critical path at 7m20s; method-level took 6m40s. All 12 Method compatibility shards and all 9 wrapped-latency shards passed. The nightly workflow retains the old exact correctness digest, 10x real-fixture speedup contract, complete nine-shard key manifest, and reverse-order confirmation. A later unchanged-production-tree run exposed a repeatable 120–123 ms single-sample Hive mapped-load spread. The harness now measures five loads of the same persisted graph and compares the median, records count/min/max for audit, and fails closed on protocol drift while retaining the original 30% plus 50 ms gate. A clean-clone real-corpus run completed in 1m14s with median/range values of Hive 138/133–201 ms, Kotlin compiler 80/76–123 ms, and Tika 94/88–144 ms.

## Gate contracts

- wrapped latency must not regress more than 50% from known-good commit `0b421f8` or more than 15% from the exact PR base
- synthetic deltas below 0.5 ms remain informational
- duplicate, missing, unexpected, non-positive, mixed-policy, or corrupt artifacts fail closed
- the transition fallback is allowed only when the base comparator matches the reviewed pre-anchor SHA-256 and candidate comparator tests pass; later bases select base-owned controls
- base-owned comparator commands, expected keys, workload harnesses, and fixture-preparation harnesses are overlaid exactly
- every JMH module sets `includeTests=false`; every workflow build compiles the full project test output and fails if any test class or resource intersects the finished fat JAR
- for old comparison revisions, the PR workflow accepts the external init/verifier only at pinned SHA-256 values; after merge it automatically uses base-owned controls, while release comparisons apply the exact current-tag controls to both tags
- the five-sample large-corpus rollout uses a one-time transition only when legacy-base, new-harness, and new-comparator SHA-256 pins plus candidate tests all match; later bases automatically return to base-owned controls
- shared build and shard artifacts use run-stable names plus overwrite so failed-job retries can reuse successful upstream artifacts
- all 11 Method scenarios at graph counts 4/17/36 remain required exactly once

## Reports, Pages, and release tags

The PR comment separates component PASS/FAIL from coverage completeness and groups evidence into semantic correctness, latency, capacity, resources, scalability, lifecycle, and gate-system sections. Uncovered families are explicit and non-blocking rather than being confused with executed failures.

`benchmark-pages.yml` is main-only and does not extend the PR gate. Paired evidence is accepted only when the PR run base matches the exact integration parent and the candidate tree matches the integration tree. Stale workflow reruns refuse to deploy over a newer `main`; all pushes are preserved with `concurrency.queue: max`.

`benchmark-tag-diff.yml` independently triggers on each `v*` tag, resolves the previous valid SemVer tag, builds both exact revisions, and runs six representative `CypherBenchmark` methods sequentially on one runner. The report first rolls measurements into seven product indicators — correctness integrity, latency, operational stability, capacity, resource efficiency, scalability, and lifecycle efficiency — plus two separate evidence dimensions: measurement confidence and coverage completeness. This exact workflow currently observes Latency only, so the other six product indicators stay explicitly unavailable rather than inheriting a conclusion. Latency uses an equal-weight geometric mean of current/previous ratios; method points, 99.9% confidence intervals, and the component-CI aggregate envelope remain visible, and an envelope crossing zero becomes an Attention item.

Each tag deployment appends or replaces a bounded 7+2 history entry and publishes the Observatory at [`/releases/`](https://johnsonlee.github.io/graphite/releases/). Trend lines break on predecessor-chain or harness-fingerprint changes so incompatible protocols are never connected. The main Pages workflow preserves the release path, while the tag workflow preserves the main Observatory; their shared serialized Pages concurrency prevents lost updates. The retained 90-day artifact still includes full SHAs, raw JMH JSON, and the PR #104 taxonomy. Workload fingerprints disable deltas when selected method bodies, shared setup, or effective JMH runtime/plugin/config drift, and run-attempt-scoped artifacts make reruns collision-safe. It is informational, has no release verdict, and has no dependency edge to `publish.yml`.

## Verification

- `node --test .github/scripts/benchmark-*.test.mjs` — 86/86 passed
- `actionlint` on PR, nightly, Pages, and tag workflows — passed (older actionlint schemas require a targeted ignore for the supported `concurrency.queue` key)
- `git diff --check` — passed
- exact-head clean clone: Cypher, Explorer, SootUp, and WebGraph `testClasses + jmhJar` builds — passed; all four archives are disjoint from their complete compiled test outputs
- old-revision transition proof: exact PR base `1c41db2`, wrapped-query reference `0b421f8`, and candidate all built under the pinned external isolation controls; no test output leaked
- workflow-equivalent large-corpus transition: both revisions emitted five-sample min/median/max markers; mapped-load deltas -2.1%, -3.5%, and +7.3%
- Pages JMH smoke: exact workflow filter produced and validated 14 rows; exact-head run `33299391407` published a byte-matched report/status pair with full SHA/run provenance, and `validatePairedEvidence` returned `available: true`
- tag resolver: current `v2.4.4` resolves exact predecessor `v2.4.3`; all six selected methods exist in both tags
- browser QA: release Observatory desktop and 390 px layouts verified; no page-level horizontal overflow, with navigation/table overflow scoped locally


.github/workflows/benchmark.yml
```diff
@@ -1315,33 +1486,40 @@ jobs:
           -jvmArgsAppend "${GRAPH_JVM_ARGS}" -foe true -prof gc -rf json \
           -rff "${PWD}/benchmark-results/candidate-latency-${SHARD}.json"
 
-    - name: Enforce fixed baseline speedup and current-base regression gate
+    - name: Enforce known-good anchor and current-base regression gate
       env:
         BENCHMARK_FILTER: ${{ matrix.filter }}
         BENCHMARK_PARAMETERS: ${{ matrix.parameters }}
         SHARD: ${{ matrix.shard }}
```
johnsonlee, line 1622: **[P1 blocking] Pass the candidate-test result into this step.** `CANDIDATE_GATE_TEST_JOB` is only defined in the preceding benchmark step environment; GitHub Actions step-level `env` values do not carry into later steps. This enforcement step reads the variable at line 1500, and the exact base `1c41db2` does not contain `compare-latency-anchor`, so every shard takes the bootstrap branch and evaluates `test "" = success`. Add `CANDIDATE_GATE_TEST_JOB: ${{ needs.candidate-gate-tests.result }}` to this step environment and cover the wiring in the workflow invariant test.

.github/workflows/benchmark-pages.yml
```diff
@@ -0,0 +1,210 @@
...
+            owner: context.repo.owner,
+            repo: context.repo.repo,
+            workflow_id: 'benchmark.yml',
+            event: 'pull_request',
+            status: 'completed',
+            head_sha: pull.head.sha,
+            per_page: 100
+          });
+          const run = runs
+            .filter((item) => item.conclusion === 'success')
```
johnsonlee, line 143: **[P1 blocking] Validate that paired evidence belongs to the exact merged integration before publishing PASS.** Filtering a successful run by `pull.head.sha` proves only the old PR head. A behind-base squash merge can apply that diff on a newer parent and produce a main tree that was never benchmarked; #104 is a concrete example. The downloaded `benchmark-status.json` currently contains no base/candidate metadata, and the renderer trusts `status.passed`, so the observatory can label stale evidence authoritative for a different main integration. Persist `baseSha` and `candidateSha` in the aggregate status, require candidate SHA/tree to match the merged main tree and base SHA to match the merge parent, and otherwise render paired evidence as unavailable. Add a behind-base squash test.

.github/workflows/benchmark.yml
```diff
@@ -33,7 +34,143 @@ jobs:
...
+      with:
+        cache-read-only: true
+
+    - name: Install base-owned wrapped-query harnesses
+      run: |
+        for HARNESS in \
+          BenchmarkCorpus.kt \
+          WrappedDiscoveryLatencyBenchmark.kt \
+          AllFixtureWrappedDiscoveryLatencyBenchmark.kt \
+          WrappedDiscoveryResourceBenchmark.kt; do
```
johnsonlee, line 179: **[P1 blocking] Keep the fixture-preparation harness base-owned.** The base workflow explicitly installed `AllFixtureBenchmarkGraphPreparation.kt` before building the candidate JMH JAR. This shared-build list omits it, while `prepare-latency-fixtures` later executes `AllFixtureBenchmarkGraphPreparation` and `AllFixtureBenchmarkDistributionCalibration` from the candidate JAR on a cache miss. A future PR can therefore accidentally redefine the persisted graphs used by reference, base, and candidate and self-certify the changed workload, contradicting the integrity claim that fixture-preparation controls come from the exact base. Include that source in this base-owned list and lock it with an invariant test.

.github/workflows/benchmark-pages.yml
```diff
@@ -0,0 +1,210 @@
+name: Main benchmark observatory
+
+on:
+  push:
+    branches: [main]
+  workflow_dispatch:
```
johnsonlee, line 6: **[P2] Prevent manual dispatch from publishing a non-main ref as main.** `workflow_dispatch` lets a maintainer select another branch or tag, but this workflow checks out `github.sha`, passes `--branch main`, updates the shared history, and deploys the production Pages site. A manual run can therefore overwrite the main observatory with a feature-ref snapshot while labeling it main. Reject or skip unless `github.ref == refs/heads/main` (including the deploy job), and add a workflow test for that guard.

docs/benchmark-regression-gate.md
```diff
@@ -18,6 +55,13 @@ keeps ordinary pull requests from accidentally changing the experiment and its p
 same revision. Candidate comparator tests run as non-authoritative test coverage in a separate job
 whose runner contains only the candidate checkout.
 
+The known-good-anchor rollout has one compatibility exception for a pull request whose base
+predates the anchor comparator command: its latency shards and latency-shard combiner may use the
+candidate comparator only when the base comparator matches the pinned reviewed SHA-256 and
+`candidate-gate-tests` passes. Any other pre-anchor base fails closed. The CODEOWNERS boundary must
+review that bootstrap change. When the base exposes the anchor command, the workflow always selects
```
johnsonlee: **[P1 blocking] The live ruleset does not guarantee this CODEOWNERS review.** Ruleset `12285695` still lists `johnsonlee` as a `pull_request` bypass actor. This owner-authored PR changes CODEOWNERS-owned gate files, currently has zero approvals, and can be merged through that bypass after checks. Because this one-time bootstrap executes candidate comparator code after candidate-owned tests, the statement that CODEOWNERS “must review” it is a false security premise. Either remove/narrow the bypass or state that this protects only non-bypass actors and that bypass credentials are inside the trusted boundary.

.github/workflows/benchmark.yml
johnsonlee: Exact-head CI reproduces this deterministically in run 33292897540. For example, job 99208084675 reaches “Enforce known-good anchor and current-base regression gate”; the logged step environment contains BENCHMARK_FILTER, BENCHMARK_PARAMETERS, and SHARD but no CANDIDATE_GATE_TEST_JOB, then exits 1 immediately. The completed synthetic-1/4/16/64 and real-b/c/d shards all failed the same rollout step, so this is workflow wiring—not benchmark variance.

.github/scripts/benchmark-gate.test.mjs
```diff
@@ -805,3 +981,61 @@ test("workflow component artifacts include the run attempt required by staging",
...
+    assert.match(workflow, /confirm-latency-anchor/);
+    assert.match(workflow, /Checkout candidate reporting code for anchor rollout/);
+    const anchorEnforcement = workflow.slice(
+        workflow.indexOf("    - name: Enforce known-good anchor and current-base regression gate"),
+        workflow.indexOf("    - name: Upload latency shard results"),
+    );
+    assert.match(
+        anchorEnforcement,
+        /CANDIDATE_GATE_TEST_JOB: \$\{\{ needs\.candidate-gate-tests\.result \}\}/,
+    );
```
johnsonlee, line 1196: **[P2] Fix the end marker so this invariant actually isolates the enforcement step.** The workflow step is named `Upload wrapped-query latency results`, but this test searches for `Upload latency shard results`. That string is absent, so `indexOf` returns `-1` and `slice(start, -1)` includes almost the entire remainder of the workflow. The assertion can therefore stay green if this env is removed again, because the later combine step also defines `CANDIDATE_GATE_TEST_JOB`. Use the exact next-step marker (and assert both boundaries are found, or parse the YAML) so the regression that broke all latency shards is genuinely covered.

.github/workflows/benchmark.yml
johnsonlee: Rechecked exact head `9a9e438`: the enforcement step now receives `CANDIDATE_GATE_TEST_JOB`, and completed synthetic-4/16/64 plus real-b/c latency shards are succeeding, so the runtime wiring defect is fixed. I am leaving this thread open only because the new invariant test does not yet isolate this step—the separate test-line comment identifies the mismatched end marker. Once that assertion is made effective, this thread can be resolved.

.github/workflows/benchmark-pages.yml
```diff
@@ -0,0 +1,210 @@
...
+
+permissions:
+  contents: read
+
+concurrency:
+  group: pages
+  cancel-in-progress: false
```
johnsonlee, line 13: **[P2] Queue every main snapshot instead of silently dropping intermediate pushes.** With the default concurrency queue (`single`), `cancel-in-progress: false` protects the running workflow but still allows a third push to cancel/replace the one pending run. A normal A-running / B-pending / C-pushed sequence therefore omits B entirely: it never uploads the forensic artifact or enters the rolling history, contradicting the stated “after every main push” contract. GitHub now supports `concurrency.queue: max` for FIFO pending runs (up to 100): https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency. Add `queue: max` and update the workflow invariant to require it rather than treating `cancel-in-progress: false` alone as sufficient.

.github/scripts/benchmark-gate.mjs
```diff
@@ -461,6 +563,77 @@ export function compareLatencyBaseline(
...
 
+export function compareLatencyAnchor(
+    anchorResults,
+    baseResults,
+    candidateResults,
+    regressionThreshold = 15,
+    anchorThreshold = 50,
+    expectedKeys = null
+) {
+    const baseComparison = compareJmh(baseResults, candidateResults, regressionThreshold);
```
johnsonlee, line 575: **[P1 blocking] Reject zero latency scores in the new anchor policy.** Both calls delegate to `compareJmh`, whose candidate validation allows `score == 0` (`candidateScore < 0`), while the previous `compareLatencyBaseline` explicitly required every latency score to be positive. On exact head `9a9e438`, a minimal anchor/base/candidate input of 100/100/0 ms/op returns `{passed:true, errors:[], delta:-100}`; the wrapped-query workflow performs no prior `> 0` validation. A zero/invalid candidate result can therefore self-report as a perfect optimization, contradicting the fail-closed artifact contract. Add latency-specific positive-score validation for anchor, base, and candidate (without breaking legitimate zero-valued non-latency secondary metrics), plus unit cases for zero in each revision.

.github/workflows/benchmark-pages.yml
```diff
@@ -0,0 +1,210 @@
...
+      pages: write
+      id-token: write
+    environment:
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}
+
+    steps:
+    - name: Deploy benchmark observatory
+      id: deployment
+      uses: actions/deploy-pages@v5
```
johnsonlee, line 308: **[P2] Do not let a re-run of an old main workflow roll the live observatory backward.** This is distinct from non-main `workflow_dispatch`: GitHub re-runs keep the original event’s `GITHUB_SHA` and `GITHUB_REF` (https://docs.github.com/en/actions/how-tos/manage-workflow-runs/re-run-workflows-and-jobs), so an old push run still passes a `github.ref == refs/heads/main` guard. It checks out/builds that old SHA, labels it `main`, and deploys unconditionally here. A renderer PoC with newer history SHA `b…` followed by a re-run of old SHA `a…` labels `a…` as main and makes it the newest history entry because the re-run timestamp is later. Resolve the current remote main tip immediately before deployment and skip/reject deployment when it differs from `github.sha`; add a stale-main-re-run invariant/test as well as the non-main-dispatch guard.

.github/workflows/benchmark.yml
johnsonlee: Fixed in 9a9e438 and hardened in 9a9b548. The anchor step now receives BASE_SHA/CANDIDATE_SHA/RUNNER/REPOSITORY/RUN_ID, and the invariant test slices the exact enforcement step through the wrapped-query upload boundary before asserting every required environment variable.

.github/workflows/benchmark-pages.yml
johnsonlee: Fixed in 9a9b548. The aggregate status now persists full base/candidate SHAs and run provenance. Pages derives the exact integration parent/tree, accepts only a successful PR run whose base equals that parent and whose candidate tree equals the integration tree, and revalidates provenance against the artifact status and Markdown before showing a verdict. Behind-base squash and mismatch cases now render UNAVAILABLE.

.github/workflows/benchmark.yml
johnsonlee: Fixed in 9a9b548. AllFixtureBenchmarkGraphPreparation.kt is now included in the exact base-owned wrapped-query harness overlay, alongside the corpus, latency, and resource harnesses. The workflow invariant test asserts the complete five-file overlay set exactly.

.github/workflows/benchmark-pages.yml
johnsonlee: Fixed in 9a9b548. snapshot, render, and deploy all explicitly require refs/heads/main, so workflow_dispatch from any other ref cannot benchmark or deploy as main. The Pages workflow test verifies the three main-only guards.

docs/benchmark-regression-gate.md
johnsonlee: Fixed in 9a9b548. The documentation now states that CODEOWNERS/ruleset protection applies to non-bypass actors and explicitly identifies repository owners and bypass credentials as part of the trusted boundary.

.github/scripts/benchmark-gate.test.mjs
johnsonlee: Fixed in 9a9b548. The regression test now finds the exact Upload wrapped-query latency results step as the end boundary, asserts both boundaries exist and are ordered, then checks the environment only inside that slice. This removes the previous false-positive boundary.

.github/workflows/benchmark-pages.yml
johnsonlee: Fixed in 9a9b548. Pages concurrency now uses queue: max with cancel-in-progress: false, preserving every main push rather than retaining only the oldest pending run. The deploy job independently rejects a stale main SHA immediately before publishing.

.github/scripts/benchmark-gate.mjs
johnsonlee: Fixed in 9a9b548. compareLatencyAnchor now rejects non-finite or non-positive primary latency scores for the known-good anchor, PR base, and candidate. Tests exercise a zero score independently in all three revisions; zero-valued secondary metrics retain their valid existing semantics.

.github/workflows/benchmark-pages.yml
johnsonlee: Fixed in 9a9b548. Immediately before deploy, the job fetches the current remote main branch and fails if it differs from context.sha, preventing an older rerun from rolling Pages backward. A workflow invariant test covers this guard.

.github/workflows/benchmark-tag-diff.yml
```diff
@@ -0,0 +1,321 @@
...
+          echo "current_tag=$(jq -r '.current.tag' tag-resolution/metadata.json)"
+          echo "previous_available=$(jq -r '.baselineAvailable' tag-resolution/metadata.json)"
+          echo "previous_sha=$(jq -r '.previous.sha // empty' tag-resolution/metadata.json)"
+          echo "previous_tag=$(jq -r '.previous.tag // empty' tag-resolution/metadata.json)"
+        } >> "${GITHUB_OUTPUT}"
+
+    - name: Upload immutable tag resolution
+      uses: actions/upload-artifact@v7
+      with:
+        name: tag-benchmark-resolution-${{ github.run_id }}
```
johnsonlee: **[P2] Make tag-diff artifacts retry-safe.** Every upload name is keyed by `github.run_id`, which is unchanged by `Re-run all jobs`, while `upload-artifact` defaults to `overwrite: false`. After a completed attempt, the rerun reaches this first upload with the same name and fails because that artifact already exists; after a partial attempt, the collision simply moves to whichever later artifact survived. Include `github.run_attempt` consistently in every upload/download name, or use an explicit safe overwrite policy, and add a workflow invariant covering reruns. GitHub documents the stable run ID/run attempt split in the [contexts reference](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context), and the action documents the overwrite default in [upload-artifact](https://github.com/actions/upload-artifact#inputs).

.github/workflows/benchmark-tag-diff.yml
```diff
@@ -0,0 +1,321 @@
...
+      with:
+        distribution: temurin
+        java-version: '17'
+
+    - name: Setup Gradle read-only cache
+      uses: gradle/actions/setup-gradle@v6
+      with:
+        cache-read-only: true
+
+    - name: Build exact previous JMH JAR
```
johnsonlee, line 151: **[P2] Do not publish signed release deltas when the benchmark workload changed between tags.** These two JARs compile each tag's own `CypherBenchmark`; the later validation proves only that six names, mode, params, unit, and numeric fields match. A release can change fixture setup, query/body, or JMH annotations while retaining all of those fields, and this renderer will attribute the workload change to product latency. The current `v2.4.3`→`v2.4.4` selected methods happen to be unchanged, but the permanent workflow needs a harness/config fingerprint and must mark deltas unavailable on drift (or build both product revisions with one reviewed common harness). Add a regression test where a selected method body differs but the JMH schema remains identical.

.github/workflows/benchmark-tag-diff.yml
johnsonlee: Fixed in 77e9c7e. Every tag-diff upload and matching download now includes both github.run_id and github.run_attempt, including the final 90-day report. The workflow invariant enumerates all five upload names, requires run_attempt in each, and requires matching downloads for the four intermediate artifacts.

.github/workflows/benchmark-tag-diff.yml
johnsonlee: Fixed in 77e9c7e. The resolver now SHA-256 fingerprints the shared fixture/setup, all six selected benchmark method bodies, and the fixed execution protocol while ignoring unrelated benchmark additions. The renderer emits signed deltas only when both fingerprints match; on drift it preserves both raw score sets, records workload-drift, and makes every delta unavailable. Tests cover an unchanged JMH schema with a selected method body change, plus the unrelated-benchmark case. The actual v2.4.3 to v2.4.4 pair resolves to the same fingerprint.

johnsonlee: Rechecked exact head `77e9c7e`: selected method-body drift is now handled, but the configuration half of the comparability issue is still open. `harnessFingerprint()` hashes only `CypherBenchmark.kt` plus the hard-coded `BENCHMARK_PROTOCOL`; it does not hash `gradle/libs.versions.toml` (`jmh = "1.37"`), the root `me.champeau.jmh` plugin version, or `graphite-cypher/build.gradle.kts`. A falsifiable case is a release whose only change is `jmh = "1.38"`: both tag-owned JARs then use different JMH runtimes/generators, yet these fingerprint inputs are byte-identical, so `workloadCompatible` remains true and signed deltas are still emitted. Include the effective JMH runtime/plugin/config inputs in the per-tag fingerprint and add this config-only drift regression case before resolving this thread.

johnsonlee: Completed the configuration half in 53592d7. The per-tag SHA-256 now also includes the effective JMH runtime version and core/generator catalog entries, the root me.champeau.jmh plugin version, the cypher module plugin/dependency declarations, and the exact jmh configuration block. A config-only 1.37 to 1.38 regression test changes the comparable fingerprint, while an unrelated Gradle test task does not. Full tests are 76/76, and the actual v2.4.3 to v2.4.4 pair still resolves to the same combined harness/config fingerprint.

johnsonlee: Verified on exact head `53592d7`: the combined fingerprint now changes for the config-only JMH 1.37→1.38 case, preserves equality for unrelated Gradle test-task changes, and still matches for the real `v2.4.3`→`v2.4.4` pair. Full local script tests pass 76/76, targeted actionlint and `git diff --check` pass. This thread is now resolved.

johnsonlee:
Benchmark safety is not established on exact head `53592d7`: [run 33295278300](https://github.com/johnsonlee/graphite/actions/runs/33295278300) failed the required gate on Hive mapped-load latency. Initial base→PR was 339→459 ms (+35.4%); reverse-order confirmation was 334→457 ms (+36.8%), exceeding both the 30% relative limit and 50 ms floor. All other product rows passed, but this repeated row correctly makes the aggregate FAIL.

This head changes only benchmark/report control files relative to the previously green head, so the result may be host/filesystem variance rather than a code-path regression; no additional ad-hoc proof is required. Keep the PR blocked and rerun the existing required benchmark gate for this exact head. A green exact-head gate is sufficient; another confirmed failure needs investigation before merge.

.github/scripts/benchmark-gate.mjs
```diff
@@ -55,7 +159,7 @@ const LATENCY_RESOURCE_EVENT_METRICS = new Set([
 const LARGE_CORPUS_METRICS = [
     { key: "buildMs", label: "build", threshold: 20, minimum: 500, unit: "ms" },
     { key: "saveMs", label: "save", threshold: 25, minimum: 250, unit: "ms" },
-    { key: "mappedLoadMs", label: "mapped load", threshold: 30, minimum: 50, unit: "ms" },
+    { key: "mappedLoadMs", label: "mapped load", threshold: 30, minimum: 150, unit: "ms" },
```
johnsonlee: **[P1 blocking] Stabilize the mapped-load measurement instead of widening the gate around one failed run.** The compared production tree is unchanged, so the 120/123 ms spread proves this single `loadMapped` sample is noisy, but it does not establish 150 ms as a safe floor. The three preceding green exact-head runs show base→PR mapped-load spreads up to +75, +70, and -77 ms; the direction and magnitude move between runs. With this change, a 200→350 ms slowdown (+75%) passes because the increase equals the 150 ms floor, and that 150 ms is also below the pipeline gate's 1,000 ms floor, so neither row catches it. Measure multiple mapped loads per persisted graph and compare a robust statistic (or otherwise remove the worktree/filesystem bias), then calibrate from that distribution. Do not convert observed harness noise into a permanent 3× loss of sensitivity.

johnsonlee: Fixed in `401ea1c`. I reverted the 150 ms floor entirely: mapped load keeps the original 30% + 50 ms gate, but the harness now loads the same persisted graph five times and compares the median. The marker records `mappedLoadSamples`, min, median, and max; the base-owned comparator requires exactly five positive ordered samples and fails closed on protocol drift. Boundary tests still block a 200→280 ms median regression. Clean-clone `:webgraph:largeCorpusTest` passed all three corpora in 1m14s: Hive median/range 138/133–201 ms, Kotlin compiler 80/76–123 ms, and Tika 94/88–144 ms. Node tests are 78/78 and clean-clone Kotlin test compilation passed.

graphite-webgraph/src/test/kotlin/io/johnsonlee/graphite/webgraph/LargeCorpusPerformanceGateTest.kt
```diff
@@ -167,21 +171,31 @@ abstract class LargeCorpusGate(private val baseline: CorpusBaseline) {
             closeQuietly(sourceGraph)
             sourceGraph = null
 
-            val loadStart = System.nanoTime()
-            loadedGraph = GraphStore.loadMapped(output)
-            val mappedLoadMillis = elapsedMillis(loadStart)
+            val mappedLoadSamples = LongArray(MAPPED_LOAD_SAMPLE_COUNT)
```
johnsonlee, line 174: **[P1 blocking] Make this five-sample protocol run in this PR’s required gate before merging it.** The implementation and its synthetic comparator tests are sound locally, but the current `large-corpus` job cannot execute this code: `.github/workflows/benchmark.yml:1166-1172` copies the base (`1c41db2`) `LargeCorpusPerformanceGateTest.kt` over the candidate checkout, and lines 1191/1206 invoke the base comparator. Because that base predates `mappedLoadSamples`, both revisions still emit one sample and the old comparator accepts it; `candidate-gate-tests` exercises only synthetic JS inputs. Thus a green exact-head benchmark run does not validate the new protocol, and its first real required execution would occur only after merge on a later PR. Add a fail-closed transition path/job that installs this exact candidate harness into both revisions and uses the exact candidate comparator (pin their hashes and require the comparator tests), then make that result part of `benchmark-regression-gate`. The artifact is an objective check: both logs must contain `mappedLoadSamples=5` plus min/median/max, and the rendered row must state `median of five loads`. This is validation by the existing gate, not a request for separate performance evidence.

johnsonlee:
Exact head `401ea1c` is not benchmark-green yet: run [33296379492](https://github.com/johnsonlee/graphite/actions/runs/33296379492) failed `wrapped-query-resources`. The AllFixture allocation row repeated in reverse order at base→candidate `4.651→6.105 GB/op` (+31.3%) and `4.557→6.105 GB/op` (+34.0%); query GC count repeated at `4.33→6` and `4→6`. I compared the downloaded base/candidate JMH JAR contents: all production and wrapped-query benchmark classes are byte-identical; only the intentionally changed `LargeCorpusPerformanceGateTest` classes differ, so this is more likely a resource-harness/JVM-layout variance than a product regression. Per the existing safety rule, no ad hoc proof is requested: keep the PR blocked and rerun the required benchmark gate on the exact next head after the bootstrap fix. A green exact-head gate is sufficient; another confirmed resource failure needs investigation before merge.

graphite-webgraph/src/test/kotlin/io/johnsonlee/graphite/webgraph/LargeCorpusPerformanceGateTest.kt
johnsonlee: Artifact verification confirms the bootstrap gap on run `33296379492`: all four `*-large-corpus*.log` files emit only `mappedLoadMs=...`; none contains `mappedLoadSamples`, `mappedLoadMinMs`, or `mappedLoadMaxMs`, and the report does not mention the five-load median. The job passed using the old one-sample protocol exactly as the workflow analysis predicted.

.github/workflows/benchmark.yml
```diff
@@ -33,7 +34,144 @@ jobs:
...
+          SOURCE="graphite-webgraph/src/jmh/kotlin/io/johnsonlee/graphite/webgraph/${HARNESS}"
+          test ! -L "source/${SOURCE}"
+          rm -f -- "source/${SOURCE}"
+          install -m 0644 "gate/${SOURCE}" "source/${SOURCE}"
+          cmp "gate/${SOURCE}" "source/${SOURCE}"
+        done
+
+    - name: Build comparable wrapped-query JMH JAR
+      run: |
+        source/gradlew -p source :webgraph:jmhJar --no-daemon
```
johnsonlee: **[P1 blocking] Keep test-output drift out of the wrapped-query JMH artifact.** The exact-head rerun failed the same resource row again: attempt 2 measured candidate `6,104,712,330.7 B/op` versus base `4,556,574,794.7 B/op` (+33.976%), and the reverse-order confirmation reproduced the same candidate value and delta. I unpacked the uploaded base/candidate JMH JARs and compared every entry: all production classes and all wrapped-query benchmark classes are byte-identical; the only differing entries are the 13 `LargeCorpusPerformanceGateTest`/support classes from `src/test`. Those test classes are unrelated to this resource probe but are being packaged by `:webgraph:jmhJar`, so this job is not isolating the code under measurement; an unrelated gate-test edit changes the fat JAR layout and repeatably changes the allocation path. Configure the JMH artifact to exclude `sourceSets.test.output` (and fail the build if any `*CorpusPerformanceGateTest*` entry appears), or otherwise neutralize test output before packaging. Then rerun the unchanged resource gate. Do not widen the resource threshold: the existing gate is correctly exposing artifact non-comparability.

johnsonlee: `96a2412` does not close this blocker. It overlays only the one test file changed by this PR, while the uploaded fat JAR contains the entire project test output, including `GraphStoreTest`, `MappedCypherBudgetTest`, `AndroidSdkIntegrationTest`, and `MethodQueryPersistenceTest` classes. A future edit to any of those files will again change the supposedly comparable resource artifact and can reproduce the same layout-sensitive allocation split. The new assertion only checks that this one pathname appears in YAML; it does not prove test output is excluded or generally neutralized. Remove `sourceSets.test.output` from `jmhJar` (with a packaging test), rather than maintaining a growing per-test overlay list. The blocker remains unresolved.

graphite-webgraph/src/test/kotlin/io/johnsonlee/graphite/webgraph/LargeCorpusPerformanceGateTest.kt
johnsonlee: `96a2412` does not change the `large-corpus` job at all, so this blocker is also still open: the job continues copying the base one-sample harness over candidate and invoking the base comparator. The next artifact will still lack `mappedLoadSamples=5` unless a separate fail-closed transition path is added.

johnsonlee: Exact-head artifact confirmation for `96a2412` / run `33296969448`: the successful large-corpus logs still contain only `mappedLoadMs` and no `mappedLoadSamples`, min, or max fields. This commit therefore remains green only under the old single-sample protocol.

.github/workflows/benchmark.yml
johnsonlee: Fixed structurally in `9b9d143`. `graphite-webgraph` now sets the JMH plugin `includeTests=false`; `verifyJmhJarExcludesTests` compiles the complete `sourceSets.test.output`, intersects every relative class/resource entry with the finished fat JAR, and fails on any leak. PR, budgeted webgraph, and historical JMH builds invoke that verification task. A clean-clone run passed, and explicit archive inspection found none of `GraphStoreTest`, `LargeCorpusPerformanceGateTest`, `MappedCypherBudgetTest`, `AndroidSdkIntegrationTest`, or `MethodQueryPersistenceTest`. The per-file overlay was removed.

graphite-webgraph/src/test/kotlin/io/johnsonlee/graphite/webgraph/LargeCorpusPerformanceGateTest.kt
johnsonlee: Fixed in `9b9d143` with a fail-closed transition path. When the base lacks the five-sample protocol, the job requires the exact reviewed legacy harness SHA-256, candidate five-sample harness SHA-256, candidate comparator SHA-256, and a successful candidate test job. It then installs the pinned harness into both base and candidate and selects the pinned comparator; any other legacy base fails. Once base exposes the protocol, selection automatically returns to base-owned controls. A workflow-equivalent local dual run produced `mappedLoadSamples=5` plus min/median/max for all six corpus records; mapped-load deltas were -2.1%, -3.5%, and +7.3% under the original 30% + 50 ms gate.

.github/workflows/benchmark.yml
```diff
@@ -33,7 +37,143 @@ jobs:
...
+          WrappedDiscoveryResourceBenchmark.kt; do
+          SOURCE="graphite-webgraph/src/jmh/kotlin/io/johnsonlee/graphite/webgraph/${HARNESS}"
+          test ! -L "source/${SOURCE}"
+          rm -f -- "source/${SOURCE}"
+          install -m 0644 "gate/${SOURCE}" "source/${SOURCE}"
+          cmp "gate/${SOURCE}" "source/${SOURCE}"
+        done
+    - name: Build comparable wrapped-query JMH JAR
+      run: |
+        source/gradlew -p source :webgraph:verifyJmhJarExcludesTests --no-daemon
```
johnsonlee: **[P1 blocking] Do not invoke a candidate-only Gradle task inside historical/base checkouts.** This matrix runs line 159 in three different repositories, but both `1c41db2` (the PR base) and `0b421f8` (the wrapped-query reference) lack `verifyJmhJarExcludesTests`, so their build jobs will fail with “task not found” before producing a JAR. The same regression exists at lines 1080/1081 for pinned lookup baseline `87c74c2` and in `benchmark-historical-latency.yml:72`; those older revisions also lack this task. Supply the exclusion and verifier from a trusted external transition control—e.g. a pinned/base-owned Gradle init script that sets `includeTests=false` and registers the verification against each checkout—rather than requiring the revision under test to already contain the new task. Keep the exact five-sample transition hashes; those three pins match the current files.

johnsonlee: Exact-head CI confirms the failure on run `33297650975`: `build-wrapped-query-jmh-base` reports `Cannot locate tasks that match :webgraph:verifyJmhJarExcludesTests`, and the reference build plus pinned budgeted-string baseline fail for the same reason; downstream latency/resource jobs are skipped. The candidate task itself passes locally and its JAR excludes all checked project test classes, so the remaining fix is specifically to inject that control into old revisions rather than invoke a task they do not define.

graphite-webgraph/src/test/kotlin/io/johnsonlee/graphite/webgraph/LargeCorpusPerformanceGateTest.kt
johnsonlee: Verified fixed on exact head `9b9d143` in [run 33297650975](https://github.com/johnsonlee/graphite/actions/runs/33297650975), job `large-corpus-benchmarks` (success). The job selected the pinned transition path after validating the legacy base harness and the candidate transition harness/comparator hashes. Downloaded artifact `benchmark-large-corpus-106-1` proves both base and candidate logs execute the new protocol for all three corpora: every audit marker contains `mappedLoadSamples=5`, `mappedLoadMinMs`, median `mappedLoadMs`, and `mappedLoadMaxMs`; the generated report states "median of five loads"; `large-corpus-status.json` is non-empty with `passed: true` and no errors. This closes the bootstrap coverage gap.

.github/workflows/benchmark.yml
johnsonlee: Fixed in `53cd285`. Comparable JMH builds no longer call a revision-local task: the workflow injects `benchmark-jmh-isolation.init.gradle`, compiles each project’s complete `testClasses`, and runs the external archive verifier. During this pre-merge transition, candidate controls are accepted only when both pinned SHA-256 values match; once the base contains them, selection automatically returns to base-owned copies. The same mechanism now covers wrapped-query reference/base/candidate, both pinned budgeted baselines, Explorer/Cypher builds, historical nightly, Pages, and both release tags. I validated exact base `1c41db2`, reference `0b421f8`, and candidate builds locally; all completed and all generated JMH JARs were disjoint from compiled test classes/resources. A clean clone of `53cd285` also built and verified Cypher, Explorer, SootUp, and WebGraph.

johnsonlee: Implementation verified on exact head `53cd285`: old revisions now receive `includeTests=false` through the external init script and are checked by the standalone JAR/test-entry intersection verifier, so they no longer invoke a candidate-only Gradle task. I independently built and verified both base `1c41db2` and wrapped reference `0b421f8` with these controls; both passed. Exact-head CI also has `build-wrapped-query-jmh-base`, `build-wrapped-query-jmh-reference`, and candidate build green, while the two pinned control hashes match the exact files. Keeping this thread open only until the existing full exact-head benchmark gate completes; no additional ad hoc proof is requested.

johnsonlee: Final exact-head verification is complete on `53cd285`: [run 33298146533](https://github.com/johnsonlee/graphite/actions/runs/33298146533) is fully green, including base/reference/candidate wrapped-query builds, both pinned budgeted baselines, resource guardrails, all nine latency shards and their aggregate, all 12 Method shards and the 33-case aggregate, large-corpus, Explorer, capacity, and the final regression gate. Downloaded `benchmark-report-106` contains a non-empty `benchmark-status.json` with `passed: true` and `errors: []`, explicitly comparing base `1c41db2` with PR `53cd285`. Together with the independent old-revision builds/verifier checks above, this proves the candidate-only-task compatibility blocker is fixed. Resolving.

.github/workflows/benchmark.yml
```diff
@@ -1648,8 +1980,29 @@ jobs:
...
+            --directory benchmark-results \
+            --base-sha '${{ github.event.pull_request.base.sha }}' \
+            --candidate-sha '${{ github.event.pull_request.head.sha }}' \
+            --runner '${{ runner.os }}-${{ runner.arch }}' \
+            --run-url '${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}' \
+            --report benchmark-results/benchmark-report.md \
+            --status benchmark-results/benchmark-render-status.json
+          node -e "const a=require('./benchmark-results/benchmark-authoritative-status.json');const r=require('./benchmark-results/benchmark-render-status.json');if(a.passed!==r.passed||JSON.stringify(a.errors)!==JSON.stringify(r.errors))process.exit(1)"
+          install -m 0644 benchmark-results/benchmark-authoritative-status.json \
+            benchmark-results/benchmark-status.json
```
johnsonlee, line 2011: **[P1 blocking] Keep the rendered report and its status body as one evidence pair.** In this transition branch, `benchmark-report.md` comes from the candidate renderer, but these lines replace `benchmark-status.json` with the base aggregator status, whose `body` contains the different base-rendered report. `benchmark-pages.mjs:151-154` deterministically rejects paired evidence unless `status.body === reportMarkdown`; this branch is entered precisely because the base report lacks `### Coverage summary`, so the two bodies cannot be equal. After merge, the Pages observatory will therefore show paired PR evidence as unavailable even though the gate passed. Keep enforcement on `benchmark-authoritative-status.json`, compare candidate status verdict **and provenance** against it, but publish the matching render status alongside the rendered report (or otherwise make Pages consume a matching pair). Add a transition-path test that passes the exact produced report/status pair through `validatePairedEvidence`.

johnsonlee: Exact-head artifact evidence from [run 33298825270](https://github.com/johnsonlee/graphite/actions/runs/33298825270) confirms this blocker and a second schema gap in the same path. Downloaded `benchmark-report-106` has a rendered `benchmark-report.md` containing `### Coverage summary`, while `benchmark-status.json.body` does not (`cmp` differs). Passing the exact pair to `validatePairedEvidence` returns `{available:false, reason:"The paired benchmark status identifies different evidence."}`. In addition, the produced status has no `baseSha`, `candidateSha`, or `runUrl` fields (all read as null/missing), although `benchmark-pages.mjs:151-152` requires all three to match the provenance file. Therefore merely publishing `benchmark-render-status.json` is insufficient unless aggregate status also persists these provenance fields. Please test the actual workflow-produced report/status schema end-to-end through `validatePairedEvidence`, including matching body, base SHA, candidate SHA, and source run URL.

johnsonlee: Fixed in `8f88cfa` and verified against the exact-head artifact from [run 33299391407](https://github.com/johnsonlee/graphite/actions/runs/33299391407). The published `benchmark-report.md` now matches `benchmark-status.json.body` byte-for-byte; the render status persists `baseSha=1c41db2ad720...`, `candidateSha=8f88cfa5f17f...`, and the exact source `runUrl`, while required-check enforcement remains isolated on `benchmark-authoritative-status.json`. Both statuses are PASS with zero errors, and the actual produced report/status pair passes `validatePairedEvidence` with `available: true`. The transition-path regression test now exercises that exact Pages-compatible pairing contract.

johnsonlee: Verified resolved on exact head `8f88cfa5f17f5c8e60d25ab78c8e0ed00d5b833f`. The complete benchmark run [33299391407](https://github.com/johnsonlee/graphite/actions/runs/33299391407) passed, and its downloaded `benchmark-report-106` artifact now publishes `benchmark-status.json` exactly equal to `benchmark-render-status.json`; `status.body` is byte-for-byte equal to `benchmark-report.md`; `baseSha`, `candidateSha`, `runner`, and `runUrl` match the separately retained authoritative status; and verdict/errors match. Passing this exact workflow-produced report/status pair with the matching run provenance through `validatePairedEvidence` returns `available: true`. Enforcement still reads `benchmark-authoritative-status.json`. The new transition test passes as part of 79/79 script tests, and the exact-head unit workflow also passed. Resolving.

johnsonlee:
**[P1 blocking] Re-run the existing gate against the tree that can actually merge.** Exact head `39dbfe7` is still rooted at `1c41db2`, while remote `main` is now `c8ea8375` after #107. The current potential merge commit has parents `c8ea8375 + 39dbfe7`, but the newly launched unit and benchmark runs `33305681564` and `33305681578` both record PR base `1c41db2`; they therefore validate neither the current-main comparison nor the actual combined merge tree. The `main` branch is not protected, so `BEHIND` does not itself prevent a squash merge. Please update or rebase this branch onto current `main` and let the existing required benchmark gate pass on that new exact head. No additional proof or new benchmark is requested; a complete green exact-head gate whose base is current `main` is sufficient.

.github/workflows/benchmark-tag-diff.yml
```diff
@@ -0,0 +1,433 @@
...
+        ')
+        STATUS=$(curl --silent --show-error --location --max-redirs 4 \
+          --proto '=https' --proto-redir '=https' \
+          --output _site/index.html --write-out '%{http_code}' "${ROOT_URL}")
+        if [ "${STATUS}" != 200 ] && [ "${STATUS}" != 404 ]; then
+          echo "::error::Published main observatory returned HTTP ${STATUS}"
+          exit 1
+        fi
+        if [ "${STATUS}" = 404 ]; then
+          install -m 0644 tag-report/index.html _site/index.html
```
johnsonlee, line 388: **[P1 blocking] Keep the tag-first bootstrap compatible with the main observatory.** This 404 fallback installs the release report as the root Pages document. The next `benchmark-pages.yml` run receives HTTP 200 for that root and unconditionally runs `benchmark-pages.mjs extract-history`, which expects `<script id="benchmark-history">...`; the release page contains only `benchmark-release-history`. I reproduced the exact sequence on `39dbfe7`: render a tag report, use it as the published root, then run the main extractor; it exits 1 with `Published benchmark page has no compatible history payload`. Shared `queue: max` serializes deployments but does not guarantee the main workflow acquires the group before a tag workflow, so a tag created around the initial merge can permanently make the main observatory fail until Pages is manually repaired. Please make the root-404 bootstrap main-history-compatible or have the main workflow recognize this release bootstrap as empty main history, and add a tag-first -> main regression test.

johnsonlee: Rechecked on exact head `a9e5501`: this remains reproducible. I rendered the updated release report from this head, used it as the tag-first root fallback, then ran the actual main-observatory `extract-history` command. It still exits 1 with `Published benchmark page has no compatible history payload`. This commit changes only the release renderer/tests and leaves both workflow branches unchanged, so the bootstrap sequence is not fixed; keeping the thread unresolved. The full local benchmark-script suite passes 82/82, so no separate blocker was found in the new forest/trend rendering.

johnsonlee: Verified fixed on exact head `e2a88c1`. I reran the original tag-first sequence with the real updated release renderer and main `extract-history`: it now exits 0 and returns `[]`, so the next main observatory can bootstrap normally. The fallback validates the embedded release history through the shared full 7+2 schema before accepting it; the committed malformed-payload case still fails closed and writes no recovered file. The complete benchmark-script suite passes 84/84 locally. Resolving this Pages bootstrap blocker; the separate current-main/base-provenance blocker remains open because this branch is still rooted at `1c41db2`.

johnsonlee:
Rerun update: attempt 2 of [run 33306953837](https://github.com/johnsonlee/graphite/actions/runs/33306953837) is genuinely green. I downloaded the final authoritative artifact: `passed=true`, `errors=[]`, so the earlier capacity/resource failures on this exact head are no longer blockers. The remaining blocker is only base provenance: that artifact still records `baseSha=1c41db2`, while the current merge ref has parents `c8ea8375 + e2a88c1`. Please sync/rebase onto current `main` and let the existing gate run once on the resulting exact head; a green run with current-main provenance is sufficient, with no additional proof requested.

johnsonlee:
Verified resolved on current-main exact head `7723157`. GitHub updated the branch with merge parents `e2a88c1 + c8ea8375`; `main` is now an ancestor and the run provenance is base `c8ea8375`, candidate `7723157`. [Run 33308347957](https://github.com/johnsonlee/graphite/actions/runs/33308347957) attempt 2 is fully green. I downloaded the final authoritative artifact: `passed=true`, `errors=[]`, with those exact base/candidate SHAs. The one initial mapped-string failure was rerun through the existing gate; attempt artifact `benchmark-budgeted-string-106-2` passed at `4.836 -> 4.506 ms/op` (-6.8%), and the aggregate correctly consumed that latest-attempt artifact. Unit tests passed, all 40 checks are green, all review threads are resolved, and the PR is now `CLEAN`/mergeable. The current-main provenance blocker is closed; no additional proof is requested.

johnsonlee:
Reverified after the new report-polish head `d05c330`. The increment is limited to `benchmark-tag-diff.mjs` and its test; the full benchmark-script suite passes 85/85 locally, and a property matrix across point counts 1–500 / label limits 2–20 confirms the thinning helper always returns the expected number of unique ordered labels with both endpoints. The DOM/CSS selectors match positioned elements and existing SVG text styling. Current-base [run 33309453103](https://github.com/johnsonlee/graphite/actions/runs/33309453103) attempt 2 is fully green: its authoritative artifact has `passed=true`, `errors=[]`, base `c8ea8375`, candidate `d05c330`. The initial capacity noise was rerun through the existing gate; attempt 2 passed (candidate primary latency -11.6% versus base), and the aggregate consumed `benchmark-cypher-capacity-106-2`. Unit tests and all 40 checks pass, no review thread is unresolved, and the PR is `CLEAN`/mergeable. No additional proof is requested.

.github/scripts/benchmark-tag-diff.mjs
```diff
@@ -0,0 +1,1212 @@
...
+    let previous = null;
+    for (const entry of history) {
+        const metric = historyMetric(entry, id);
+        const connected = previous !== null && entry.previousTag === previous.tag &&
+            entry.harnessFingerprint === previous.harnessFingerprint && entry.comparisonStatus === "available";
+        if (!connected) {
+            segment++;
+            latencyIndex = 100;
+        }
+        if (metric?.currentValue === null || metric === null || entry.comparisonStatus !== "available") {
```
johnsonlee: **[P1] Preserve absolute evidence observations across comparison breaks**

`comparisonStatus` controls whether two releases can be compared/connected, but it does not determine whether the current release has an absolute observation. On this exact head, generating the first semantic tag produces `comparisonStatus: "baseline-unavailable"` while the manifest contains `measurement-confidence: { state: "observed", currentValue: 2.9167 }` and `coverage-completeness: { state: "observed", currentValue: 0 }`; this condition drops both, so the new 7+2 trend section renders all nine cards as “No comparable numeric observations yet.” The same loss occurs at a harness/protocol break, even though the copy says the line should break rather than discard the point. Please let `metric.currentValue` decide whether an absolute point exists and use comparison compatibility only to start a new segment (latency remains safe because its value is null when no delta exists), with regression coverage for a first tag and a protocol-break entry.

johnsonlee: Reverified on exact head `c3e4a0e`; this is not fixed. The change hides empty cards and changes the copy, but the data-loss condition in `trendSeries` is unchanged. Reproduction results:

- First tag: manifest reports Measurement confidence `observed` at `2.9167` and Coverage completeness `observed` at `0`, while `renderedTrendMetrics` is `[]`.
- Workload/harness drift at `v2.4.5`: both evidence dimensions are again `observed` with current values, but their exact-value tables contain only `v2.4.4`; the current release is discarded instead of starting a new segment.

The full script suite still passes 85/85 because the updated assertions cover only an `available` comparison and still add no first-tag or protocol/workload-break regression case. Please fix the observation/connection distinction requested above rather than hiding the symptom, and add those two tests. Keeping this P1 open.

johnsonlee: Fixed in `60559cb`. `trendSeries` now treats absolute observation existence and cross-release connectivity independently: a non-null `metric.currentValue` always emits a point, while predecessor/fingerprint/comparison compatibility only selects the segment. Latency remains absent when its comparison-derived `currentValue` is null. Added two regression cases: the first semantic tag now renders Measurement confidence and Coverage completeness, and a workload/harness drift preserves both current evidence points while producing a protocol break. The complete benchmark script suite passes 86/86 locally. I will resolve this thread after exact-head CI completes.

johnsonlee: Exact-head verification is complete on `60559cbd5cd757d2bc48bd83ba8836763789fd0e`. [Benchmark run 33314140664](https://github.com/johnsonlee/graphite/actions/runs/33314140664) and [unit run 33314140639](https://github.com/johnsonlee/graphite/actions/runs/33314140639) are green; all 40/40 checks succeeded. The script suite now has 86/86 passing tests, including the requested first-tag and protocol-break cases. Resolving.

johnsonlee: Verified fixed on exact head `60559cbd5cd757d2bc48bd83ba8836763789fd0e`.

- The implementation now emits every non-null absolute observation independently of `comparisonStatus`; compatibility only controls segment connectivity.
- The complete benchmark script suite passes independently: 86/86.
- The added first-tag regression renders both Measurement confidence and Coverage completeness for `v2.4.4`.
- The drift regression retains both `v2.4.4` and `v2.4.5` evidence observations and renders one protocol break.
- Exact-head CI is 40/40 green. The authoritative benchmark artifact reports `passed: true`, no errors, base `c8ea83750394e85585b23a6a10e28ef759bac9cf`, candidate `60559cbd5cd757d2bc48bd83ba8836763789fd0e`.

The reported data-loss bug is resolved.

PR #107, opened 2026-08-30
Bound and accelerate broad Cypher searches

johnsonlee:
## Summary

- add cross-graph fast paths for the non-`DISTINCT` broad string-disjunction searches seen in production, plus `A AND (B OR C)` pushdown while preserving source order, duplicates, provenance, and `LIMIT`
- make mapped raw string-match state a bounded access-order LRU so long backtests do not fall off a cache-capacity cliff after many unique predicates
- replace server work-unit limiting with a 60-second maximum timeout; clients can request `timeoutMs`, with `effectiveTimeout = min(clientTimeout, serverMaxTimeout)`, and timed-out work is cancelled/interrupted with HTTP 504
- keep `--cypher-work-budget` parseable but deprecated and ignored by `serve`; direct library callers retain finite `CypherExecutionBudget` behavior
- expose timeout configuration through CLI, OpenAPI, metrics, README, and the MCP `timeout_ms` argument

## Verification

- `JAVA_OPTS='-Xmx3g -XX:MaxMetaspaceSize=1g' ./gradlew check --no-daemon` — passed, including detekt, coverage verification, unit/integration tests, and Tika/Hive/Kotlin large-corpus gates
- `npm ci && npx tsc --noEmit && npm run build` in `graphite-mcp` — passed
- `node --test graphite-explore/src/test/js/ui-state.test.js` — 7 passed
- `node --test .github/scripts/benchmark-gate.test.mjs` — 46 passed
- final head `38479a3`: full `:explore:test`, `:explore:detekt`, and all 26 GitHub checks passed

## Performance evidence

Environment: Apple Silicon macOS, OpenJDK 17.0.18, JMH 1.37, one fork/thread, `main` at `1c41db2` and candidate at `472a23a`, run sequentially on the same host.

Targeted command (identical temporary harness on both revisions):

```text
JAVA_OPTS='-Xmx3g -XX:MaxMetaspaceSize=1g' ./gradlew :cypher:jmh -Pjmh.filter='.*CrossGraphSearchBenchmark.*' --no-daemon
```

| Benchmark | main (ms/op) | candidate (ms/op) | change |
| --- | ---: | ---: | ---: |
| `CrossGraphSearchBenchmark.nonDistinctBroadOrZeroHit` | 161.732 | 22.392 | -86.2% |
| `CrossGraphSearchBenchmark.requiredPredicateWithResidualOrZeroHit` | 117.089 | 37.082 | -68.3% |

The targeted method-level comparison shows a large improvement and no regression for the changed search shapes.

Required repository `CypherBenchmark` command, shortened locally to reduce duplicate CI time:

```text
java -jar graphite-cypher/build/libs/cypher-1.0.0-SNAPSHOT-jmh.jar '^io\.johnsonlee\.graphite\.cypher\.CypherBenchmark\..*$' -wi 1 -i 2 -w 500ms -r 500ms -f 1
```

| Benchmark | main (us/op) | candidate (us/op) | change |
| --- | ---: | ---: | ---: |
| `aggregationCountGroupBy` | 32.562 | 33.493 | +2.9% |
| `countStar` | 2.725 | 3.033 | +11.3% |
| `filteredSingleHopRelationship` | 57.722 | 57.285 | -0.8% |
| `filteredVariableLengthPath` | 176.590 | 175.447 | -0.6% |
| `functionCalls` | 27.511 | 27.693 | +0.7% |
| `nodeMatchWithWhere` | 58.948 | 59.254 | +0.5% |
| `orderedFilteredSingleHopRelationship` | 252.152 | 251.988 | -0.1% |
| `regexFilter` | 24.905 | 24.979 | +0.3% |
| `returnDistinct` | 113.841 | 113.058 | -0.7% |
| `simpleNodeMatch` | 23.263 | 22.629 | -2.7% |
| `singleHopRelationship` | 34.929 | 37.061 | +6.1% |
| `variableLengthPath` | 32.509 | 33.271 | +2.3% |
| `withPipeline` | 86.887 | 86.391 | -0.6% |

The short local general-purpose suite is mixed around run-to-run noise and does not indicate a coherent method-level regression; the protected `benchmark-regression-gate` with its full repetitions and confirmation rules is authoritative.

The first CI runs exposed and confirmed a finite-budget hot-loop regression after cancellation started carrying a typed cause. The current revision keeps the original atomic-boolean hot poll and stores the first typed cause on the cold cancellation path; the unlimited-budget shortcut was also removed. The exact protected collection filter was then rerun locally with its full 2 forks, 3 warmups, and 5 measurements:

| Benchmark | main (us/op) | candidate (us/op) | change |
| --- | ---: | ---: | ---: |
| `BudgetedCypherBenchmark.budgetedListConcatenation` | 280.678 | 273.156 | -2.7% |
| `BudgetedCypherBenchmark.budgetedListMembership` | 386.099 | 382.600 | -0.9% |

The corrected finite-budget path shows no collection-latency regression; the final same-runner CI rerun on `38479a3` is authoritative.

End-to-end conclusion: local candidate large-corpus gates passed (`queryMs`: Hive 1604, Kotlin compiler 1086, Tika 1288). The required same-runner base/candidate [`benchmark-regression-gate`](https://github.com/johnsonlee/graphite/pull/107#issuecomment-5467506108) also passed all method-level, real-corpus, latency, capacity, and resource checks. It indicates no method-level or end-to-end performance regression.

### DIFF
--- a/graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/CypherExecutionBudget.kt
+++ b/graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/CypherExecutionBudget.kt
@@ -4,6 +4,7 @@ import io.johnsonlee.graphite.graph.GraphWorkConsumer
 import java.util.concurrent.CancellationException
 import java.util.concurrent.atomic.AtomicBoolean
 import java.util.concurrent.atomic.AtomicLong
+import java.util.concurrent.atomic.AtomicReference
 
 internal const val CANCELLATION_POLL_MASK = 1_023
 
@@ -38,18 +39,31 @@ class CypherCancellationSignal(
     private val checkObserver: (() -> Unit)? = null
 ) {
     private val cancelled = AtomicBoolean()
+    private val cancellation = AtomicReference<CypherQueryCancelledException?>()
 
     val isCancelled: Boolean get() = cancelled.get()
 
-    fun cancel(): Boolean = cancelled.compareAndSet(false, true)
+    fun cancel(): Boolean = cancel(CypherQueryCancelledException())
+
+    fun cancel(reason: CypherQueryCancelledException): Boolean {
+        if (!cancellation.compareAndSet(null, reason)) return false
+        cancelled.set(true)
+        return true
+    }
+
+    fun cancellationException(): CypherQueryCancelledException =
+        cancellation.get() ?: CypherQueryCancelledException()
 
     fun throwIfCancelled() {
         checkObserver?.invoke()
-        if (isCancelled) throw CypherQueryCancelledException()
+        if (isCancelled) throw cancellationException()
     }
 }
 
-class CypherQueryCancelledException : CancellationException("Cypher query cancelled")
+open class CypherQueryCancelledException(message: String = "Cypher query cancelled") : CancellationException(message)
+
+class CypherQueryTimeoutException(val timeoutMillis: Long) :
+    CypherQueryCancelledException("Cypher query timed out after $timeoutMillis ms")
 
 class CypherBudgetExceededException(
     val maxWorkUnits: Long
--- a/graphite-explore/src/main/kotlin/io/johnsonlee/graphite/cli/OpenApiSpecBuilder.kt
+++ b/graphite-explore/src/main/kotlin/io/johnsonlee/graphite/cli/OpenApiSpecBuilder.kt
@@ -467,22 +467,39 @@ internal class OpenApiSpecBuilder {
         API_PARAM_TIMEOUT_MILLIS,
         TYPE_INTEGER,
         false,
-        CYPHER_TIMEOUT_DESCRIPTION
+        CYPHER_TIMEOUT_DESCRIPTION,
+        minimum = 1
     )
 
-    private fun queryParameter(name: String, type: String, required: Boolean, description: String): Map<String, Any?> =
-        parameter("query", name, type, required, description)
+    private fun queryParameter(
+        name: String,
+        type: String,
+        required: Boolean,
+        description: String,
+        minimum: Int? = null
+    ): Map<String, Any?> = parameter("query", name, type, required, description, minimum)
 
     private fun pathParameter(name: String, type: String, description: String): Map<String, Any?> =
         parameter("path", name, type, true, description)
 
-    private fun parameter(location: String, name: String, type: String, required: Boolean, description: String): Map<String, Any?> =
+    private fun parameter(
+        location: String,
+        name: String,
+        type: String,
+        required: Boolean,
+        description: String,
+        minimum: Int? = null
+    ): Map<String, Any?> =
         mapOf(
             "in" to location,
             API_FIELD_NAME to name,
             FIELD_REQUIRED to required,
             FIELD_DESCRIPTION to description,
-            "schema" to mapOf(API_FIELD_TYPE to type)
+            "schema" to if (minimum == null) {
+                mapOf(API_FIELD_TYPE to type)
+            } else {
+                mapOf(API_FIELD_TYPE to type, "minimum" to minimum)
+            }
         )
 
     private fun response(description: String): Map<String, Any?> =
--- a/graphite-explore/src/main/kotlin/io/johnsonlee/graphite/cli/OpenApiSpecBuilder.kt
+++ b/graphite-explore/src/main/kotlin/io/johnsonlee/graphite/cli/OpenApiSpecBuilder.kt
@@ -60,12 +60,14 @@ internal class OpenApiSpecBuilder {
                     parameters = listOf(
                         pathParameter(API_FIELD_GRAPH_ID, TYPE_STRING, "Unique graph id"),
                         queryParameter(API_PARAM_QUERY, TYPE_STRING, true, "Cypher query text"),
-                        queryParameter(API_PARAM_LIMIT, TYPE_INTEGER, false, "Server-side maximum result rows")
+                        queryParameter(API_PARAM_LIMIT, TYPE_INTEGER, false, "Server-side maximum result rows"),
+                        cypherTimeoutParameter()
                     ),
                     responses = mapOf(
                         "200" to response("Cypher result"),
                         "400" to response("Missing or invalid query"),
-                        "429" to response("Cypher concurrency or work budget exceeded"),
+                        "429" to response("Cypher concurrency limit exceeded"),
+                        "504" to response("Cypher query timeout exceeded"),
                         "503" to response("Cypher cancelled while the client remained connected"),
                         "404" to response("Graph not loaded")
                     )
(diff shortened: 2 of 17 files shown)


graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/CypherExecutionBudget.kt
```diff
@@ -71,6 +80,7 @@ internal class CypherWorkTracker(
     fun consume(workUnits: Long) {
         require(workUnits >= 0) { "workUnits must be non-negative" }
         checkCancelled()
+        if (unlimited) return
```
johnsonlee: **[P1 blocking] Preserve the finite-budget hot path while adding unlimited server execution.** The required `budgeted-collection-latency` artifact on this exact head has already failed both the initial and reverse-order confirmation rounds: `budgetedListConcatenation` regressed from 563.306 to 913.200 us/op (+62.1%), then 562.849 to 920.045 (+63.5%); `budgetedListMembership` regressed from 530.672 to 762.669 us/op (+43.7%), then 529.798 to 765.073 (+44.4%). Both confidence intervals are separated and the 15% gate blocks them, so this is the authoritative same-runner evidence that the server optimization shifted substantial cost onto retained finite-`CypherExecutionBudget` callers. This new per-`consume` unlimited check (together with the cancellation-state change in the same hot path) is the only production delta shared by these base-owned benchmarks; please isolate/profile the cause and keep the `Long.MAX_VALUE` fast path without imposing this regression on finite budgets. Re-run the required gate to prove the fix.

graphite-explore/src/main/kotlin/io/johnsonlee/graphite/cli/OpenApiSpecBuilder.kt
```diff
@@ -441,6 +463,13 @@ internal class OpenApiSpecBuilder {
             )
         )
 
+    private fun cypherTimeoutParameter(): Map<String, Any?> = queryParameter(
+        API_PARAM_TIMEOUT_MILLIS,
+        TYPE_INTEGER,
+        false,
+        CYPHER_TIMEOUT_DESCRIPTION
```
johnsonlee: **[P2] Encode the positive timeout constraint in the query-parameter schema too.** The request-body schemas correctly add `minimum: 1`, and the server deterministically rejects zero/negative values, but this helper emits only `{type: integer}` through the generic `queryParameter`. As a result, every GET endpoint (and POST callers choosing the query string) advertises values that the API rejects, so generated clients cannot enforce the same contract. Add `minimum: 1` to this parameter schema and assert it structurally for each Cypher route, rather than only checking that the OpenAPI text contains `timeoutMs`.

graphite-webgraph/src/main/kotlin/io/johnsonlee/graphite/webgraph/MappedWebGraphBackedGraph.kt
```diff
@@ -710,21 +710,30 @@ internal class RawStringMatchStates(
...
+        maxEntries.coerceAtLeast(1) + 1,
+        STRING_PROPERTY_INDEX_LOAD_FACTOR,
+        true
+    )
     private var bytes = 0L
 
     @Synchronized
     fun stateFor(key: RawStringMatchKey, stringCount: Int): ByteArray? {
-        val existing = states[key]
+        var state = states[key]
```
johnsonlee, line 722: **[P1 blocking] The LRU rewrite must not regress the retained mapped zero-hit path.** On this exact head, the required `budgeted-mapped-string-latency` artifact from run `33300530306` fails the base-owned `MappedStringAdmissionBenchmark.budgetedTransformedZeroHit`: 3.539 -> 4.378 ms/op (+23.7%) initially and 3.502 -> 4.059 ms/op (+15.9%) in reverse-order confirmation, above the 15% gate in both rounds. This benchmark clears match state and performs the same 50k-node transformed zero-hit scan on base and candidate, so the current state-map rewrite has shifted cost onto an existing mapped lookup shape even though the new eviction behavior is intended to help long backtests. Preserve bounded LRU behavior, but restore this existing path below the authoritative gate and re-run the complete benchmark workflow.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/CypherExecutionBudget.kt
johnsonlee: Follow-up on exact head `fbf39ca`: removing the `unlimited` branch did **not** clear this blocker. Run [33300530306](https://github.com/johnsonlee/graphite/actions/runs/33300530306) still fails `budgetedListMembership` at 536.542 -> 759.573 us/op (+41.6%), with reverse-order confirmation 538.482 -> 780.871 us/op (+45.0%). `budgetedListConcatenation` is +31.3% in this run but correctly classified as noise because its intervals overlap; the membership failure alone is sufficient. This narrows the regression away from the removed branch and toward another change on the retained graph-free cancellation/budget path (notably the cancellation-state representation/check), but the required criterion remains the same: keep this thread unresolved until a complete exact-head gate passes.

johnsonlee: Verified resolved on exact head `472a23a0fc42d266bbcdb61093286459d19cac90` with required run [33300901385](https://github.com/johnsonlee/graphite/actions/runs/33300901385). The base-owned `budgeted-collection-latency` gate now passes: list concatenation is 568.943 -> 565.315 us/op (-0.6%) and list membership is 545.630 -> 554.718 us/op (+1.7%), both below the 15% threshold with overlapping confidence intervals. The affected Cypher/cancellation tests also pass on this exact head. Resolving this performance thread.

graphite-webgraph/src/main/kotlin/io/johnsonlee/graphite/webgraph/MappedWebGraphBackedGraph.kt
johnsonlee: Verified resolved by the latest required gate on exact head `472a23a0fc42d266bbcdb61093286459d19cac90`. In run [33300901385](https://github.com/johnsonlee/graphite/actions/runs/33300901385), `budgeted-mapped-string-latency` passes at 4.461 -> 4.540 ms/op (+1.8%), below the 15% threshold; the WebGraph LRU tests also passed on the reviewed implementation. Per the repository benchmark gate as the authoritative safety criterion, resolving this thread.

graphite-explore/src/main/kotlin/io/johnsonlee/graphite/cli/OpenApiSpecBuilder.kt
johnsonlee: Implemented on `38479a3`: every Cypher timeout query parameter now emits an OpenAPI integer schema with `minimum: 1`, matching request-body validation and runtime rejection. The endpoint test structurally checks the schema for GET and POST on `/api/cypher`, `/api/cypher/graphs`, and `/api/graphs/{graphId}/cypher`. Targeted test, full `:explore:test`, and `:explore:detekt` pass locally.

johnsonlee: Verified on exact head `38479a3de2aac3024ff72719a1731eb2c4528fdc`: `cypherTimeoutParameter()` now emits `schema: {type: integer, minimum: 1}`, and `ExploreCommandTest` structurally walks every operation on `/api/cypher`, `/api/cypher/graphs`, and `/api/graphs/{graphId}/cypher`, asserting both fields for each timeout query parameter. I independently ran the full `ExploreCommandTest` class and `:explore:detekt`; both pass. This resolves the deterministic OpenAPI contract mismatch.

PR #110, opened 2026-08-31
Optimize ordered distinct graph ID queries

johnsonlee:
## Summary

- add a strict cross-graph fast path for `MATCH (n) RETURN DISTINCT n.graphId AS graphId ORDER BY graphId LIMIT 100`
- enumerate qualified source metadata instead of materializing and sorting every node
- preserve `MATCH` semantics by excluding empty graphs, and probe only one node when a backend cannot provide `nodeCount`
- preserve ASC/DESC, aliases, literal LIMIT, external maxRows, provenance, cancellation checks, and single-graph fallback behavior
- add execution tests and verify the exact query with a supplementary fail-closed JMH harness

## Verification

- `node --test graphite-explore/src/test/js/ui-state.test.js`
- `JAVA_OPTS=-Xmx8g JVM_OPTS=-Xmx8g ./gradlew :cypher:check :cypher:koverXmlReport :cypher:jmhJar check -S --no-daemon`
- Cypher line coverage: 5226/5331 (98.03%)
- full repository check passed, including Hive, Kotlin compiler, and Tika large-corpus performance gates

## Supplementary benchmark

Temporary harness command (run sequentially against `main` at c237af0 and the production change; the harness itself is intentionally not part of the PR diff):

`./gradlew :cypher:jmh -Pjmh.filter='CrossGraphCypherBenchmark.orderedDistinctGraphIds' --no-daemon`

Environment: Apple M3 Max, macOS arm64, OpenJDK 17.0.18. Exact benchmark: `io.johnsonlee.graphite.cypher.CrossGraphCypherBenchmark.orderedDistinctGraphIds` (16 graphs x 5,000 nodes).

| Revision | Average time | 99.9% CI |
| --- | ---: | ---: |
| main | 15.658 ms/op | 15.210–16.106 ms/op |
| candidate | 0.064 ms/op | 0.063–0.066 ms/op |

The exact target query is approximately 244.7x faster (99.59% lower latency), so this targeted method comparison indicates no performance regression. The required [`benchmark-regression-gate`](https://github.com/johnsonlee/graphite/pull/110#issuecomment-5480786934) passed all 9/9 component reports on the same Linux-X64 runner. The standard `CypherBenchmark` method comparison found no confirmed regression, and all Tika, Hive, and Kotlin compiler end-to-end `JAR -> build -> save -> mapped load -> Cypher` comparisons passed. Therefore, neither the method-level nor end-to-end comparison indicates a blocking performance regression.

### DIFF
--- a/graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
+++ b/graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
@@ -52,6 +52,7 @@ import java.util.concurrent.atomic.AtomicInteger
 
 private const val COUNT_QUERY_CLAUSES = 2
 private const val LABEL_HISTOGRAM_QUERY_CLAUSES = 5
+private const val ORDERED_DISTINCT_GRAPH_ID_QUERY_CLAUSES = 4
 private const val DISTINCT_LIMIT_QUERY_CLAUSES = 3
 private const val FILTERED_LIMIT_QUERY_CLAUSES = 4
 private const val SINGLE_HOP_LIMIT_QUERY_CLAUSES = 3
@@ -242,6 +243,7 @@ class QueryPipeline private constructor(
         }
         val fastResult = tryFastNodeCount(clauses)
             ?: tryFastLabelHistogram(clauses)
+            ?: tryFastOrderedDistinctGraphIdLimit(clauses)
             ?: tryFastOrderedPropertyLimit(clauses)
             ?: tryFastDistinctPropertyLimit(clauses)
             ?: tryFastFilteredStringCount(clauses)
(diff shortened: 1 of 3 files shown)


graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
```diff
@@ -609,6 +611,65 @@ class QueryPipeline private constructor(
...
+        if (node.labels.isNotEmpty() || node.properties.isNotEmpty()) return null
+
+        val item = ret.items.single()
+        val property = item.expression as? CypherExpr.Property ?: return null
+        if (property.expression != CypherExpr.Variable(variable) || property.propertyName != GRAPH_ID_PROPERTY) {
+            return null
+        }
+        val column = item.alias ?: item.expression.toCypherString()
+        val sort = orderBy.items.single()
+        val sortsProjectedGraphId = sort.expression == item.expression ||
```
johnsonlee: This branch makes the fast path accept an unaliased `ORDER BY n.graphId`, but that query does not currently have the same result as the generic path. On this exact head, with populated sources supplied in encounter order `[空, Ω, graph-2, graph-10, a:colon, z]`, `MATCH (n) RETURN DISTINCT n.graphId ORDER BY n.graphId LIMIT 1` returns `a:colon`, while the semantically equivalent query with `SKIP 0` (which disables this fast path) returns `空`. A review-only differential test covering counted/uncounted and empty sources fails here; the same test passes in full on base `c237af0`, where both forms use the generic path. This means adding/removing `SKIP 0` changes results only after this optimization. Please either restrict this fast path to the alias-based shape actually targeted by the PR, or make the generic `ORDER BY` behavior consistent and cover both equivalent forms.

johnsonlee: Rechecked exact head `58d9fc2`. Its only delta from `23a82fe` is removal of the candidate-owned supplementary JMH method; the production fast path and tests are unchanged. The review-only differential suite still fails at `Pr110DifferentialTest.kt:61` with the same unaliased fast-path vs `SKIP 0` generic-path mismatch, so this blocker remains unresolved.

johnsonlee: Rechecked exact head `8f26248d8906dd3d5507a3e0da96b6896907de8c`. The blocker remains: `./gradlew :cypher:test --tests io.johnsonlee.graphite.cypher.Pr110DifferentialTest --no-daemon` still fails `catalog fast path matches generic execution across ordering and limits` at line 61 for the unaliased `ORDER BY n.graphId` case. Moving `tryFastOrderedDistinctGraphIdLimit(clauses)` inside `tryFastOrderedPropertyLimit()` does not narrow dispatch, because `tryFastOrderedPropertyLimit()` still invokes that graphId fast path unconditionally before compiling/checking its own ordered-property shape. Please either constrain the graphId fast path to the supported aliased shape or make its result semantics match generic execution, and commit the differential regression coverage.

johnsonlee: Fixed in ae4268e. The catalog fast path now accepts only ORDER BY on the projected alias, which is the PR target shape. Unaliased ORDER BY n.graphId falls back to generic execution. Added a committed differential regression test using [空, Ω, graph-2, graph-10, a:colon, z] and SKIP 0; CrossGraphCypherExecutorTest, detekt, and Kover all pass in an isolated clone.

PR #111, opened 2026-08-31
Always write resource stores and diagnose missing sidecars

johnsonlee:
## Summary

- always materialize `graph.resources`, including a valid zero-entry store for graphs without supported text resources
- distinguish a valid empty store from a graph whose persisted resource sidecar is missing
- return HTTP `409` with a rebuild instruction instead of silently reporting a missing store as empty
- preserve the existing two-method `ResourceAccessor` JVM ABI by exposing diagnostics through a separate capability
- cover both actual-JAR branches through `BuildCommand`: non-empty persisted content and an 8-byte loadable empty store
- verify the non-empty CLI path byte-for-byte with the published `spring-jcl-6.1.0.jar`

## Root cause

The behavior was introduced by `ca1870b093b851038863959399a2535b1ce73033` with the resource graph/API feature. It first shipped in `v1.2.0-alpha.1`, first appeared in a stable release in `v1.2.0`, and remained present through `v2.4.5`, including `v2.3.0`.

Version 2.3.0 invokes `PersistedResourceStore.save`, but that implementation returns before opening `graph.resources` whenever resource collection is empty. Collection only keeps `.properties`, `.yml`, `.yaml`, `.json`, `.xml`, and `.txt`, and read failures are silently dropped. Consequently, a successful 2.3.0 build can produce no `graph.resources` when its input has no matching readable entries.

Loading then compounded the problem: a missing sidecar was mapped to `EmptyResourceAccessor`, so the resources API could not distinguish a graph with zero resources from a graph whose resource store had never been written.

## Behavior

- graph with no supported resources: `graph.resources` exists as a valid 8-byte zero-entry store and `/api/resources` returns `200` with an empty result
- graph whose sidecar is missing: other graph APIs remain usable; resources endpoints return `409` with a current-CLI rebuild instruction
- graph built from a resource-bearing JAR: persisted content is returned by `/api/resources/{path}`

A graph without the sidecar cannot recover resource bytes that were never persisted; it must be rebuilt from its source archive with a fixed CLI.

## Verification

- final head `8d9a6c35b159`: required unit and benchmark checks pass
- old-binary/new-runtime compatibility: a committed test compiles an accessor against the old two-method API, loads it against the current runtime, and verifies resource availability lookup without `AbstractMethodError`; candidate `javap` exposes only `list` and `open`
- published non-empty JAR CLI E2E: `spring-jcl-6.1.0.jar` to `BuildCommand` to non-empty `graph.resources` to `GraphStore.load`; `META-INF/license.txt` path, source name, and bytes match the source JAR
- actual empty JAR E2E: compiled class-only JAR to `BuildCommand` to 8-byte `graph.resources` to `GraphStore.load`; resources are available and empty
- generated non-empty JAR API E2E: compiled fixture JAR to `BuildCommand` to MAPPED load to HTTP resource content
- published Android fixture: `android-all-14-robolectric-10818077.jar` to graph to save/load; `AndroidManifest.xml` matched byte-for-byte
- targeted compatibility, GraphStore, API, empty-JAR, and published-JAR tests pass in a clean clone
- all review threads and ordinary blocking comments are resolved

## Performance

Exact head `8d9a6c35b159` passed the required [benchmark regression gate](https://github.com/johnsonlee/graphite/actions/runs/33444863652), 9/9 component reports, against `main` at `e9410a4f7424`. A failed-job rerun cleared two isolated Method CPU measurements; all 12 Method shards and the aggregate gate finished green. Method-level and end-to-end results show no confirmed regression, and the large-corpus JAR-to-build-to-save-to-mapped-load-to-query gate passed.

Empty-resource saves add one 8-byte sequential sidecar write; no query execution path changes.

### DIFF
--- a/graphite-core/src/main/kotlin/io/johnsonlee/graphite/input/ResourceAccessor.kt
+++ b/graphite-core/src/main/kotlin/io/johnsonlee/graphite/input/ResourceAccessor.kt
@@ -11,16 +11,6 @@ import java.io.InputStream
  * Spring Boot fat JAR / WAR.
  */
 interface ResourceAccessor {
-
-    /**
-     * Explains why resources cannot be queried, or `null` when this accessor is usable.
-     *
-     * An available accessor may still contain no matching resources. This distinction lets
-     * persisted graphs report a missing legacy resource store instead of silently appearing empty.
-     */
-    val unavailableReason: String?
-        get() = null
-
     /**
      * List resource entries matching a glob pattern.
      *
@@ -43,6 +33,25 @@ interface ResourceAccessor {
     @Throws(java.io.IOException::class)
     fun open(path: String): InputStream
 }
+
+/**
+ * Capability implemented by accessors that cannot serve resources.
+ *
+ * This is deliberately separate from [ResourceAccessor] so implementations compiled against an
+ * older Graphite release remain binary compatible.
+ */
+interface UnavailableResourceAccessor {
+    val unavailableReason: String
+}
+
+/**
+ * Explains why resources cannot be queried, or `null` when this accessor is usable.
+ *
+ * An available accessor may still contain no matching resources. This distinction lets persisted
+ * graphs report a missing legacy resource store instead of silently appearing empty.
+ */
+val ResourceAccessor.unavailableReason: String?
+    get() = (this as? UnavailableResourceAccessor)?.unavailableReason
 
 /**
  * A resource file entry within a project archive.
--- a/graphite-core/src/main/kotlin/io/johnsonlee/graphite/input/ResourceAccessor.kt
+++ b/graphite-core/src/main/kotlin/io/johnsonlee/graphite/input/ResourceAccessor.kt
@@ -13,6 +13,15 @@ import java.io.InputStream
 interface ResourceAccessor {
 
     /**
+     * Explains why resources cannot be queried, or `null` when this accessor is usable.
+     *
+     * An available accessor may still contain no matching resources. This distinction lets
+     * persisted graphs report a missing legacy resource store instead of silently appearing empty.
+     */
+    val unavailableReason: String?
+        get() = null
+
+    /**
      * List resource entries matching a glob pattern.
      *
      * Supported patterns:
--- a/graphite-query/src/test/kotlin/io/johnsonlee/graphite/cli/GraphiteCommandTest.kt
+++ b/graphite-query/src/test/kotlin/io/johnsonlee/graphite/cli/GraphiteCommandTest.kt
@@ -1,5 +1,6 @@
 package io.johnsonlee.graphite.cli
 
+import io.johnsonlee.graphite.input.unavailableReason
 import io.johnsonlee.graphite.webgraph.GraphStore
 import picocli.CommandLine
 import java.io.ByteArrayOutputStream
--- a/graphite-query/src/test/kotlin/io/johnsonlee/graphite/cli/GraphiteCommandTest.kt
+++ b/graphite-query/src/test/kotlin/io/johnsonlee/graphite/cli/GraphiteCommandTest.kt
@@ -7,8 +7,11 @@ import java.io.PrintStream
 import java.io.PrintWriter
 import java.nio.file.Files
 import java.nio.file.Path
+import java.util.jar.JarEntry
+import java.util.jar.JarOutputStream
 import kotlin.test.Test
 import kotlin.test.assertEquals
+import kotlin.test.assertNull
 import kotlin.test.assertTrue
 import kotlin.test.fail
 
@@ -143,6 +146,56 @@ class GraphiteCommandTest {
         } finally {
             classesDir.toFile().deleteRecursively()
             outputDir.toFile().deleteRecursively()
+        }
+    }
+
+    @Test
+    fun `build from jar without persistable resources writes loadable empty store`() {
+        val root = Files.createTempDirectory("build-empty-resource-jar-test")
+        val classesDir = Files.createDirectories(root.resolve("classes"))
+        val fixtureJar = root.resolve("class-only-fixture.jar")
+        val outputDir = root.resolve("graph")
+        try {
+            val javaFile = root.resolve("ClassOnlySample.java")
+            Files.writeString(
+                javaFile,
+                """
+                package sample;
+                public class ClassOnlySample {
+                    public int value() { return 42; }
+                }
+                """.trimIndent()
+            )
+            val compiler = requireNotNull(javax.tools.ToolProvider.getSystemJavaCompiler())
+            assertEquals(
+                0,
+                compiler.run(null, null, null, "-d", classesDir.toString(), javaFile.toString()),
+                "Java fixture compilation should succeed"
+            )
+
+            JarOutputStream(Files.newOutputStream(fixtureJar)).use { jar ->
+                jar.putNextEntry(JarEntry("sample/ClassOnlySample.class"))
+                Files.copy(classesDir.resolve("sample/ClassOnlySample.class"), jar)
+                jar.closeEntry()
+            }
+
+            val cmd = BuildCommand().apply {
+                input = fixtureJar
+                output = outputDir
+                includePackages = listOf("sample")
+            }
+            val (_, err, code) = captureOutput { cmd.call() }
+            assertEquals(0, code, "JAR graph build should succeed, stderr: $err")
+
+            val resourceStore = outputDir.resolve("graph.resources")
+            assertTrue(Files.isRegularFile(resourceStore))
+            assertEquals(8L, Files.size(resourceStore), "An empty persisted resource store is header-only")
+
+            val loaded = GraphStore.load(outputDir)
+            assertNull(loaded.resources.unavailableReason)
+            assertEquals(emptyList(), loaded.resources.list("**").toList())
+        } finally {
+            root.toFile().deleteRecursively()
         }
     }
 
--- a/graphite-query/src/test/kotlin/io/johnsonlee/graphite/cli/GraphiteCommandTest.kt
+++ b/graphite-query/src/test/kotlin/io/johnsonlee/graphite/cli/GraphiteCommandTest.kt
@@ -8,8 +8,10 @@ import java.io.PrintWriter
 import java.nio.file.Files
 import java.nio.file.Path
 import java.util.jar.JarEntry
+import java.util.jar.JarFile
 import java.util.jar.JarOutputStream
 import kotlin.test.Test
+import kotlin.test.assertContentEquals
 import kotlin.test.assertEquals
 import kotlin.test.assertNull
 import kotlin.test.assertTrue
@@ -196,6 +198,39 @@ class GraphiteCommandTest {
             assertEquals(emptyList(), loaded.resources.list("**").toList())
         } finally {
             root.toFile().deleteRecursively()
+        }
+    }
+
+    @Test
+    fun `build from published jar preserves persisted resource bytes`() {
+        val fixtureJar = Path.of(requireNotNull(System.getProperty("spring.jcl.jar.path")))
+        val resourcePath = "META-INF/license.txt"
+        val expectedContent = JarFile(fixtureJar.toFile()).use { jar ->
+            val entry = requireNotNull(jar.getJarEntry(resourcePath)) {
+                "Published fixture does not contain $resourcePath"
+            }
+            jar.getInputStream(entry).use { it.readBytes() }
+        }
+        val outputDir = Files.createTempDirectory("build-published-resource-jar-test")
+        try {
+            val cmd = BuildCommand().apply {
+                input = fixtureJar
+                output = outputDir
+                includePackages = listOf("org.apache.commons.logging")
+            }
+            val (_, err, code) = captureOutput { cmd.call() }
+            assertEquals(0, code, "Published JAR graph build should succeed, stderr: $err")
+
+            val resourceStore = outputDir.resolve("graph.resources")
+            assertTrue(Files.size(resourceStore) > 8L, "Published JAR must produce a non-empty resource store")
+
+            val loaded = GraphStore.load(outputDir)
+            val entry = loaded.resources.list(resourcePath).single()
+            assertEquals(resourcePath, entry.path)
+            assertEquals(fixtureJar.fileName.toString(), entry.source)
+            assertContentEquals(expectedContent, loaded.resources.open(resourcePath).use { it.readBytes() })
+        } finally {
+            outputDir.toFile().deleteRecursively()
         }
     }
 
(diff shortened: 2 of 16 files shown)


graphite-query/src/test/kotlin/io/johnsonlee/graphite/cli/GraphiteCommandTest.kt
```diff
@@ -120,6 +121,7 @@ class GraphiteCommandTest {
             Files.createDirectories(sampleDir)
             val compileResult = compiler.run(null, null, null, "-d", classesDir.toString(), javaFile.toString())
             assertEquals(0, compileResult, "Java compilation should succeed")
+            Files.writeString(classesDir.resolve("application.properties"), "feature.mode=shadow\n")
```
johnsonlee, line 130: This does not cover the JAR path claimed by the PR: `cmd.input` is still a classes directory, so the test bypasses `JavaProjectLoader` archive discovery and `ArchiveResourceAccessor`. A regression that builds nodes from a JAR but drops its resource entries would still pass this test. I verified the missing path independently on exact head with review-only tests: (1) CLI-build the real `spring-jcl-6.1.0.jar`, reload the graph, and byte-compare `META-INF/license.txt` with the original JAR entry; (2) CLI-build the real `picocli-4.7.6.jar`, which has no persistable suffixes, and require an 8-byte loadable empty `graph.resources`. Both pass on `07f239c`, but uncommitted reviewer checks are not regression coverage. Please add deterministic fixture-JAR tests for both non-empty resource round-trip and the empty-resource sidecar through `BuildCommand` + `GraphStore.load`; the current directory test is not evidence for real JAR build-to-load behavior.

johnsonlee: Rechecked exact head `ac82ed72a4167d6662475033e459f60a02df9ceb`. The new generated-JAR HTTP test and the three large-corpus fixture checks now cover the non-empty archive path, and they pass locally. The second requested case is still absent from committed coverage: the empty-store tests call `GraphStore.save(DefaultGraph.Builder().build(), ...)` directly rather than building a real JAR with no persistable resource suffixes through `BuildCommand` and then loading the 8-byte sidecar. My reviewer-only picocli JAR test still passes on this head, but it remains uncommitted. Keeping this thread unresolved until both real-JAR branches are regression-covered in the PR.

graphite-webgraph/src/test/kotlin/io/johnsonlee/graphite/webgraph/LargeCorpusPerformanceGateTest.kt
```diff
@@ -35,6 +36,7 @@ data class CorpusBaseline(
     val persistedEdgeCount: Long,
     val methodCount: Long,
     val callSiteCount: Long,
+    val resourcePath: String,
```
johnsonlee: Blocking exact-head gate integrity: changing this transition harness changes its pinned SHA-256, but `.github/workflows/benchmark.yml` still contains the previous `LARGE_CORPUS_TRANSITION_HARNESS_SHA256=66feedea...`. The actual SHA-256 of this file on `ac82ed7` is `a2d7354ae7cf9bd6a7b9c527c7a90e5dc063617e8b0b9c499e4e499aa437a601`. The required `candidate-gate-tests` fails deterministically at `benchmark-gate.test.mjs:1145` on that mismatch, and the large-corpus job would reject the changed harness as well. Update the workflow pin to the exact new harness hash and keep the fail-closed test green.

graphite-query/src/test/kotlin/io/johnsonlee/graphite/cli/GraphiteCommandTest.kt
johnsonlee: Rechecked exact head `07af7787777eed752bdb83e40252eda2d334b55b`. The Android resource assertion itself passes in a forced local execution (47s), but it calls `JavaProjectLoader` + `GraphStore.save/load` directly, so it does not cover the requested `BuildCommand` branch. This commit also removes the Tika/Hive/Kotlin byte-for-byte checks rather than updating their harness pin; the PR body still claims those three fixture checks exist, which is now stale. Most importantly, there is still no committed real/generated JAR with zero persistable resources built through `BuildCommand` and reloaded as an 8-byte empty sidecar. The existing generated non-empty JAR test is valid, but the second branch remains uncovered; keeping this thread unresolved.

johnsonlee: The PR description has now been corrected to distinguish the committed Android fixture test from local Tika/Hive/Kotlin spot checks, so that stale-description concern is addressed. The remaining blocker in this thread is only the missing committed zero-persistable-resource JAR path through `BuildCommand` and reload of the 8-byte sidecar.

johnsonlee: Addressed in e19f43f7b66e4fac9e3d4415ce1262924367f263. The committed test now compiles a class-only JAR with no persistable resource suffixes, runs it through BuildCommand, asserts graph.resources is the 8-byte header-only store, reloads it through GraphStore.load, and verifies resources are available and empty. The full GraphiteCommandTest class passes 11/11 in a clean clone.

johnsonlee: Verified exact head `e19f43f7b66e4fac9e3d4415ce1262924367f263`. The committed class-only JAR test now exercises the complete missing branch through `BuildCommand`, asserts the 8-byte `graph.resources`, reloads it with `GraphStore.load`, and verifies an available empty accessor. It passes together with the reviewer-only non-empty and empty real dependency-JAR controls. This blocker is resolved.

johnsonlee:
**RESOLVED:** a subsequent failed-job rerun on the same exact head produced a complete green result: `benchmark-regression-gate` **PASS (9/9)**, all 12 Method compatibility shard statuses pass with zero errors, and the final status is tied to candidate `e19f43f7b66e4fac9e3d4415ce1262924367f263` and base `c237af093870a63e021b3198df03e6ea2c21bf36`. This supersedes the failed attempts documented below; no benchmark blocker remains.

---

Blocking exact-head gate result on `e19f43f7b66e4fac9e3d4415ce1262924367f263`: the required `benchmark-regression-gate` is **FAIL (8/9)**, not pending or green.

The failed component is `method-compatibility`. A failed-job rerun did not clear it:

- attempt 1: 4-graph early CPU `+18.7%` after reverse-order confirmation; 36-graph middle CPU `+22.9%` after confirmation;
- attempt 2: 4-graph middle CPU `+15.7%` after confirmation; 36-graph early CPU `+34.1%` and zero CPU `+16.7%` after confirmation.

All corresponding wall-time and final-RSS rows pass, every benchmark request succeeded (this is not a 429/failure-latency artifact), and the failing scenario moves between attempts, so the data is compatible with hosted-runner CPU variance rather than a resource-store code-path regression. The rerun did clear the prior real-36 wrapped-latency failure. However, per the repository's existing safety criterion, a red required aggregate cannot be treated as ready-to-merge. No bespoke proof is requested: make the exact-head benchmark gate pass (or fix a falsifiable defect in the gate), then update the PR description's final-head verification claim.

graphite-query/src/test/kotlin/io/johnsonlee/graphite/cli/GraphiteCommandTest.kt
```diff
@@ -132,12 +137,68 @@ class GraphiteCommandTest {
...
+                loaded.resources.open("application.properties").bufferedReader().use { it.readText() }
+            )
         } finally {
             classesDir.toFile().deleteRecursively()
             outputDir.toFile().deleteRecursively()
         }
     }
 
+    @Test
+    fun `build from jar without persistable resources writes loadable empty store`() {
```
johnsonlee, line 156: This synthetic class-only JAR proves only that an empty header can be written and loaded. It does not prove the PR title's real-JAR path or that persisted resource content survives the actual CLI build. Add a real fixture JAR case here (or in an isolated integration test): invoke BuildCommand, require a non-empty graph.resources, call GraphStore.load, list a known fixture resource, and compare its bytes with the source JAR. Reviewer reproduction: Tika tika-app-2.9.2.jar + log4j2_batch_process.properties passes with 4 GiB in ~1m40s, but OOMs with the current default test heap.

johnsonlee (commented):
Blocking: this PR is not ready to merge because its committed tests do not exercise the user-facing real-JAR CLI path. I reproduced the missing path against the resolved Tika fixture: BuildCommand -> graph.resources -> GraphStore.load -> read known resource bytes. It OOMs under the current :query:test default heap, and passes with a 4 GiB test heap in about 1m40s. Please commit a deterministic real-fixture integration test that invokes BuildCommand, asserts graph.resources is non-empty, reloads the output with GraphStore, and byte-compares a known resource from the source JAR. Configure that isolated build test with 4 GiB heap so CI proves the supported path instead of silently covering only synthetic input.

graphite-query/src/test/kotlin/io/johnsonlee/graphite/cli/GraphiteCommandTest.kt
johnsonlee: Addressed in 6c7ecaf31299eedc2c22900551977ad670453cb6. A dedicated non-transitive fixture now resolves the published spring-jcl-6.1.0.jar. The committed test invokes BuildCommand on that JAR, requires graph.resources > 8 bytes, reloads with GraphStore.load, lists META-INF/license.txt, verifies the source JAR name, and byte-compares the persisted content with the source JarFile entry. The full GraphiteCommandTest passes 12/12 in a clean clone in 17 seconds.

graphite-core/src/main/kotlin/io/johnsonlee/graphite/input/ResourceAccessor.kt
```diff
@@ -12,6 +12,15 @@ import java.io.InputStream
  */
 interface ResourceAccessor {
 
+    /**
+     * Explains why resources cannot be queried, or `null` when this accessor is usable.
+     *
+     * An available accessor may still contain no matching resources. This distinction lets
+     * persisted graphs report a missing legacy resource store instead of silently appearing empty.
+     */
+    val unavailableReason: String?
```
johnsonlee: [P1] Preserve binary compatibility for existing ResourceAccessor implementations

This default-bodied Kotlin property is emitted as an abstract JVM interface method under the current compiler mode. On base `e9410a4`, `javap ResourceAccessor` has only `list` and `open`; on exact head `3f1f6ef`, it adds `public abstract String getUnavailableReason()`. I compiled a `LegacyAccessor` against the base core JAR implementing only the two old methods, then ran `accessor.getUnavailableReason()` with the candidate core JAR; it deterministically fails with:

`java.lang.AbstractMethodError: Receiver class LegacyAccessor does not define or inherit ... ResourceAccessor.getUnavailableReason()`

`ExploreRoutes` now invokes this getter for every resources request, so an existing extension/accessor binary that worked on the current 2.x release can crash after this PR. The real-JAR persistence tests do not cover an implementation compiled against the previous public ABI, and the benchmark gate cannot detect this.

Please preserve the existing interface ABI—for example, expose availability through a new marker/capability interface implemented by `MissingPersistedResourceAccessor`, with a cast/extension helper returning null for legacy accessors—or otherwise use a JVM-default mechanism proven by a committed old-binary/new-runtime compatibility test. The resource diagnosis itself does not require adding an abstract method to the existing public interface.

johnsonlee: Addressed in 8d9a6c35b1597a69da7cc21046040d59015251d9. ResourceAccessor is back to its original two-method JVM ABI; unavailable diagnostics now use a separate UnavailableResourceAccessor capability plus a static extension lookup, so legacy implementations are never asked to invoke a new interface method. The committed regression test compiles a LegacyAccessor against an explicit old two-method ResourceAccessor stub, removes that stub, loads the implementation against the current core runtime, and verifies the diagnostic lookup returns null without AbstractMethodError. javap on the candidate core JAR now shows only list(String) and open(String). The core compatibility test, missing-store GraphStore/API tests, empty-JAR CLI test, and published spring-jcl CLI test all pass in a clean clone.

johnsonlee: Verified fixed on exact head `8d9a6c35b1597a69da7cc21046040d59015251d9`. `ResourceAccessor` again declares only `list(String)` and `open(String)` in the candidate JAR; availability is obtained through the separate `UnavailableResourceAccessor` capability and a static extension lookup. I independently reran the exact committed old-stub/new-runtime compatibility test with `--rerun-tasks`, and it passed; `javap` independently confirms that no `getUnavailableReason` method remains on the old public interface. This removes the deterministic `AbstractMethodError` reproduction, so I am resolving this thread. The new exact-head CI/benchmark run remains a separate PR-level requirement.

johnsonlee:
**RESOLVED:** failed-job rerun attempt 2 on the same exact head `8d9a6c35b1597a69da7cc21046040d59015251d9` passed both previously failing Method shards. `method-compatibility-gate` and the required `benchmark-regression-gate` then passed, with the final benchmark report at **9/9 component reports passed** against base `e9410a4f74246c470752c47f14aad7738fa4c9dd`. This supersedes the initial failed measurements below; no benchmark blocker remains.

---

Exact-head benchmark blocker on `8d9a6c35b1597a69da7cc21046040d59015251d9`: two authoritative Method CPU shards failed and repeated above the existing 15% threshold in reverse-order confirmation.

- `17-string / late`: `2.36B → 2.76B ns` (**+16.9%**) initially; confirmation `2.42B → 3.95B ns` (**+63.2%**).
- `36-aggregate / order`: `3.45B → 4.02B ns` (**+16.5%**) initially; confirmation `3.81B → 4.53B ns` (**+18.9%**).

Wall time, RSS, and semantic execution pass in both shards, so these are successful queries rather than failure/429 latency. The ABI fix has no obvious production connection to Method execution, but under the agreed policy the existing exact-head benchmark gate is the safety criterion; a twice-confirmed red gate cannot be overridden by plausibility. No additional custom proof is requested: rerun or fix until the exact-head required benchmark gate passes. Jobs: [17-string](https://github.com/johnsonlee/graphite/actions/runs/33444863652/job/99663159543), [36-aggregate](https://github.com/johnsonlee/graphite/actions/runs/33444863652/job/99663159550).

johnsonlee:
**RESOLVED on the same exact head `8d9a6c35b1597a69da7cc21046040d59015251d9`.** The failed shards were rerun through the existing gate and did not reproduce the blocking regressions:

- `17-string / late` CPU is now `1.96B → 1.85B ns` (**-5.6%**). Its other initial CPU spikes were reverse-confirmed at **+9.2%** and **-3.1%**, so the gate classifies them as noise.
- `36-aggregate / order` CPU is now `3.78B → 4.26B ns` (**+12.7%**), below the 15% limit; wall and RSS rows pass.

The rerun jobs, aggregate `method-compatibility-gate`, and final `benchmark-regression-gate` are all SUCCESS. All 40 exact-head checks are green. Under the agreed benchmark-as-safety-criterion policy, the prior blocker is cleared without requesting additional proof. Reruns: [17-string](https://github.com/johnsonlee/graphite/actions/runs/33444863652/job/99664979186), [36-aggregate](https://github.com/johnsonlee/graphite/actions/runs/33444863652/job/99664979222).

PR #114, opened 2026-09-04
Revert v2.4.8 and cut global-wide P95 beyond 10x

johnsonlee:
## Status: not ready to merge

Current head: `f32ecfd6b91b8ac999490091b8ee52f23cd3bdcf` (Attempt 169).
[Exact-head benchmark workflow](https://github.com/johnsonlee/graphite/actions/runs/33960808104)
is still running global-wide, but Method compatibility and graph routing have already failed.
Exact-head unit tests passed. Do not merge or publish without explicit authorization.

Confirmed current-head blockers (not superseded by older green results):

- Method CPU reverse-order confirmations: 4-graph late 880 -> 1,020 ms (+15.9%), order
  680 -> 880 ms (+29.4%), zero 520 -> 700 ms (+34.6%). Wall and RSS pass for these rows;
  process-CPU attribution to query/server/client/JIT/GC remains unproven.
- Routing cold K64 P50 0.298 -> 0.627 ms; startup-prepared K64 P95 2.883 -> 4.051 ms.
  Warm routing passes, but both listed states fail the existing policy.
- The global-wide result and Attempt 169's localized-early retention decision remain pending.

The latest completed benchmark run is for Attempt 167,
`2adf1e01eaead8c463e2112d106d8f2de3ca9596`, not the current head:
[CI run](https://github.com/johnsonlee/graphite/actions/runs/33959031107),
[standard benchmark report](https://github.com/johnsonlee/graphite/pull/114#issuecomment-5544626460).
That run failed. Aggregate P95 speedup above the 10x target does not override a failing
per-query, compatibility, or resource gate. Git reports no merge conflict, but required checks
remain blocking.

## Scope

- Patch-forward replacement of v2.4.8, beginning with the v2.4.7 product tree in `0396cf1f`.
  Published tags are untouched: v2.4.7 is `78ce46b57b2d88ae0f1823432ffefc5c7685bc1b`;
  v2.4.8/current base is `4e328b0109e13c896b74004823fb049fcb19251a`.
- Remove nested graph/segment parallelism from the optimized CallSite route and use bounded
  graph-level work with persisted mapped indexes. Keep authoritative fallback behavior for
  unsupported query shapes and mixed graph capabilities.
- Recent attempts introduce bounded mapped projection reuse and one-time public-row
  materialization. Attempt 167 repairs internal-provenance alias filtering in that fast path.
- Attempt 168 makes the dual-baseline combiner reject unsuccessful child statuses even with
  empty diagnostics; no measurement or threshold changes. Attempt 169 separately restores
  direct projection from an already-warm leading mapped view for bounded unscoped queries.
  The existing raw-leading method is unchanged; cold views, unsupported expressions, and
  LIMIT above 200 retain their ordinary paths.
- Each attempt, including rejected experiments, is recorded in
  [the chronological experiment log](docs/wrapped-case-insensitive-query-optimization-attempts.md).
  One experiment is recorded per commit. The log now also corrects Attempt 166's inaccurate
  claim that its shared pure-string builder was unchanged.

## CI acceptance and remaining failures

Hosted CI is the performance acceptance authority. Both references run against the candidate
on the same Linux-X64 GitHub runner, with the pinned real persisted fixtures and trusted
comparison harnesses. CPU architecture, thresholds, correctness checks, and required gates
are unchanged. Local unit tests and lint are correctness/static preflights only.

Latest completed-run conclusions, all scoped to Attempt 167:

| Evidence | CI conclusion |
| --- | --- |
| Method-level JMH, including `CypherBenchmark` | Pass under the existing comparison policy |
| End-to-end `LargeCorpusPerformanceGateTest` | Pass for the covered real corpora and lifecycle stages |
| Cold/warm/startup graph routing | Pass this run; earlier cold-K64 failures remain in the history |
| Frozen v2.4.7 global-wide goal | Fail: wrapped-case-insensitive/dense regresses in all three pairs despite aggregate P95 meeting 10x |
| Current-main global-wide non-regression | Fail: repeated provenance/zero, provenance/targeted, four-properties/targeted, aliased/targeted, and localized-early/dense rows |
| Method compatibility | All 12 shards and the aggregate gate pass this run; earlier confirmed CPU/RSS failures remain documented |

Provenance/dense no longer blocks either reference in this run, but the full suite is still red.
Attempt 169 targets the repeated localized-early failure, where main has an initialized-view
projection and the candidate instead materializes full nodes. Wrapped/dense remains a separate
unresolved regression; equal charged work does not establish equal execution cost or its root cause.
Routing also retains a first-cold K64 slowdown within its existing absolute allowance: a green
gate is not a claim of zero performance sacrifice. The exact measurements are in the linked
report and chronological log. Previously published local performance tables are not acceptance
evidence and have been removed from this PR description.

## Correctness verification for Attempts 168–169

The gate test reproduces the empty-diagnostics false pass before the repair; all 118 benchmark
logic tests then pass. Warm-leading full and partial-prefix tests fail on pre-change production
and pass after restoration. Five new tests cover exact values, duplicates, source order, graph
provenance, preferred mapped consumers, cold/large-limit refusal, and wrong-owner fallback.
Full `:cypher:test`, `:cypher:detekt`, and `git diff --check` pass locally. These are correctness
checks, not performance evidence. Current-head Method CPU and routing failures are listed above;
the global-wide comparison is still pending.

This benchmark suite covers specific workloads; it is not proof of production concurrency,
all warm/cold mixtures, or universal functionality. No readiness claim will be made until the
required exact-head checks pass and outstanding review requirements are resolved.


.github/scripts/run-real64-global-wide.sh
```diff
@@ -204,7 +204,7 @@ node "${CANDIDATE_TREE}/${COMPARATOR_PATH}" compare-global-wide-pressure \
   --run-orders candidate-base,base-candidate,candidate-base \
   --graph-manifest "${MANIFEST}" \
   --correctness-oracle "${ORACLE}" \
-  --minimum-speedup 5 \
+  --minimum-speedup 10 \
```
johnsonlee: [P1] Gate 10× against the intended release baseline. The workflow passes `github.event.pull_request.base.sha` into this script, so on #114 this compares the candidate with current main/v2.4.8—not with v2.4.7 used by the PR's 14.99×–17.30× claim. Exact-head CI consequently measured only 0.83× / 1.48× / 0.79× P95 and failed. Please either pin and authenticate v2.4.7 as the cumulative release baseline while retaining a separate current-main regression guard, or keep this PR-base comparison at an incremental threshold. As written, the required CI path cannot prove the claimed 10× result.

johnsonlee (commented):
Blocking review note: GitHub does not allow this account to submit REQUEST_CHANGES on its own PR, so this is filed as a review comment. The real64 gate must pass on the exact head before merge.

johnsonlee (commented):
Blocking review on exact head ed47b423: the dual-baseline wiring addresses the earlier baseline mismatch in design, but reverting the v2.4.8 runtime shifts substantial cost to Method discovery and graph-routed workloads. Required exact-head gates are red.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
```diff
@@ -101,139 +91,22 @@ private fun resolveNodeClass(labels: List<String>): Class<out Node>? =
...
 }
 
-private fun directStringGraphParallelism(sourceCount: Int, graphScoped: Boolean = false): Int =
-    resolveDirectStringGraphParallelism(
-        sourceCount,
-        configuredGraphWorkers = configuredDirectStringParallelism,
-        graphScoped = graphScoped
-    )
-
-private fun usesBalancedStringSplit(sourceCount: Int): Boolean =
```
johnsonlee, line 171: [P1] Preserve the current-main Method fast paths. This rollback removes the balanced split/prepared-storage scheduling stack, and the exact-head Method compatibility gate now has 13 regressions that repeat in reverse-order confirmation—not a single noisy point. Examples: 17-graph zero wall +97.7% and CPU +72.7%; 36-graph zero wall +85.9% and CPU +114.8%; 36-graph late CPU +32.5%; 4-graph suffix CPU +55.9%. A replacement optimized only for the global-wide CallSite shape cannot regress these existing Method workloads. Please preserve equivalent routing/parallelism (and its regression coverage) rather than reverting it wholesale.

graphite-webgraph/src/main/kotlin/io/johnsonlee/graphite/webgraph/MappedWebGraphBackedGraph.kt
```diff
@@ -1704,19 +1322,6 @@ internal class MappedWebGraphBackedGraph(
         return false
     }
 
-    override fun hasPreparedStringPropertyDisjunction(
```
johnsonlee, line 1707: [P1] Preserve the startup-prepared graph-routing contract or an equivalent fast path. On this exact head, the required graph-routing gate shows the candidate performing 64 parallel raw scans in the cold state while current main performs zero. The repeated impact is material across every width: cold K2 P95 0.486→2.374 ms and K64 2.752→4.107 ms; startup-prepared K2 0.480→2.355 ms and K64 2.609→3.969 ms. Removing this capability while adding a separate mapped path has shifted latency back onto graph-routed queries, so restore equivalent prepared routing before this rollback is safe.

.github/scripts/run-real64-global-wide.sh
johnsonlee: The baseline wiring itself is fixed on ed47b423: the hosted job now runs a frozen v2.4.7 10x goal comparison plus a separate exact current-main non-regression comparison. The dual-baseline gate correctly failed for real per-shape regressions, which are tracked in the new blocking threads, so this obsolete wiring thread can be resolved.

graphite-webgraph/src/main/kotlin/io/johnsonlee/graphite/webgraph/MappedWebGraphBackedGraph.kt
```diff
@@ -457,217 +422,129 @@ internal class MappedWebGraphBackedGraph(
...
-        callSitePredicatesCannotMatch(predicates, workConsumer)
+    ): List<StringPropertyDistinctRow>? {
+        if (predicates.any { predicate -> !predicate.canUseMappedCallSiteIndexView() }) return null
+        val view = mappedCallSiteStringIndexView(workConsumer) ?: return null
+        val exactMatches = if (selectedValues == null) {
+            view.exactMatchingStringIds(predicates, workConsumer)
+        } else {
+            null
+        }
+        val useMappedView = if (selectedValues != null) {
```
johnsonlee, line 450: [P1] Make the mapped-view admission preserve dense and zero-hit shapes, not just improve the aggregate tail. The completed exact-head dual-baseline gate fails six aligned dense cases versus v2.4.7 and multiple current-main cases in at least two of three forks. Examples: current-main four-properties dense 33.747→51.784 ms and zero max 421.007→552.838 ms; class-pair dense 1.812→8.889 ms. The aggregate P95 is 13.29x–16.36x versus v2.4.7, but that does not make these repeated per-shape regressions safe. Please refine this path-selection policy (or preserve the faster existing path for these shapes) until both halves of the authoritative dual gate pass.

johnsonlee (commented):
Additional exact-head blocker from the completed dual-baseline run: aggregate P95 clears 10x versus v2.4.7, but the required per-shape and current-main guards fail repeatedly.

graphite-core/src/main/kotlin/io/johnsonlee/graphite/graph/Graph.kt
```diff
@@ -75,70 +75,13 @@ interface GraphWorkBatchConsumer : GraphWorkConsumer {
...
 fun interface ParallelGraphWorkBatchConsumer : GraphWorkBatchConsumer
 
-/**
- * A thread-safe batch consumer for a cross-graph lookup that splits one process-wide CPU budget
- * between graph workers and a shared background segment pool. The calling graph worker participates
- * in the scan; [segmentWorkerCount] is the maximum number of additional workers that this lookup may
- * request from that shared pool. Concurrent graph lookups share the same pool, so this value is not
- * multiplied by the number of graph workers.
- */
-interface SplitGraphWorkBatchConsumer : ParallelGraphWorkBatchConsumer {
```
johnsonlee, line 85: [P1] Keep the v2.4.8 public API ABI-compatible. SplitGraphWorkBatchConsumer, GraphScanParallelismPlan, the Preferred* consumers, PreparedStringPropertyDisjunctionLookup, and RetainedStringPropertyDisjunctionLookup are public top-level types in the published graphite-core artifact. Deleting them in a patch release breaks downstream source references and can produce NoClassDefFoundError for already-compiled consumers. Internal code may stop using these contracts, but the exported types must remain (they can be deprecated) unless this is intentionally shipped as a major breaking release.

johnsonlee (commented):
Compatibility blocker: the rollback removes exported graphite-core types from a published patch release while the PR description claims there is no public API change.

graphite-webgraph/src/main/kotlin/io/johnsonlee/graphite/webgraph/MappedWebGraphBackedGraph.kt
johnsonlee: Addressed by 8ce816eb and fb5ac7fc. Retained graph sets are projected without node materialization, while the leading dense source can use the bounded primitive projection before any suffix work. The targeted graph-routing cases now avoid the 64 raw-scan regression; the exact final hosted gate is running on 1d477784.

graphite-core/src/main/kotlin/io/johnsonlee/graphite/graph/Graph.kt
johnsonlee: Addressed by a9080d05 with test coverage completed in 1d477784. All published v2.4.8 scheduling and lookup types remain as compatibility-only declarations while the replacement runtime stays v2.4.7-based. javap -public matches v2.4.8 for the restored types, and a core-JAR class-entry comparison found no removed v2.4.8 class.

graphite-webgraph/src/main/kotlin/io/johnsonlee/graphite/webgraph/MappedWebGraphBackedGraph.kt
johnsonlee: Addressed across fb5ac7fc, 9e4f80e3, 33121e71, and 6d5792b3. Dense leading rows stay synchronous, cold suffixes use at most four outer workers, and all-warm suffixes reuse that bounded path. The final real64 dual-baseline run passed all three forks: v2.4.7 P95 speedups 16.84x, 15.50x, and 16.64x; candidate P95 versus current main was lower by 40.29%, 39.04%, and 39.73%. The formerly failing four-property and class-pair targeted cases also passed.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
johnsonlee: Addressed by 6e409f3b, 7c1de39d, 3640e2be, 21423a1f, and 6c6ebee9. The accepted Method Attempts 091/094/097/099 are restored on top of the exact v2.4.7 rollback, with guard-path coverage. The final hosted Method matrix is being rerun on 1d477784.

graphite-core/src/main/kotlin/io/johnsonlee/graphite/graph/Graph.kt
johnsonlee: Independently verified on exact head 1d477784 against current main 4e328b0: both `:core:jar` builds succeed; the candidate core JAR removes no class entries; and `javap -public` is identical for every class shared with current main (the only class-entry delta is four additive implementation types). The ABI blocker is fixed, so I am resolving this thread.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
johnsonlee: Exact-head hosted evidence is still red on 1d477784. `method-compatibility-17-string` passed wall time and RSS, but the 17-graph string `suffix` case increased CPU from 2.86s to 3.97s (+38.8%), and reverse-order confirmation reproduced it at 3.02s to 4.18s (+38.4%) against the 15% limit. Accordingly `method-compatibility-gate` also failed. This is reverse-confirmed rather than a one-order noisy point, so the Method blocker remains unresolved.

graphite-webgraph/src/main/kotlin/io/johnsonlee/graphite/webgraph/MappedWebGraphBackedGraph.kt
johnsonlee: The exact-head hosted graph-routing gate completed and is still red. Warm and startup-prepared now pass, but cold fails: graphId P95 regresses from 0.65x speedup overall, K2 0.434→0.544 ms, K8 1.184→1.321 ms, K64 3.155→3.881 ms, and the first cold K64 request rises 582.427→703.882 ms. The candidate also still records 64 intra-graph parallel scans across all 64 graphs (peak 4 workers), versus zero on current main. This partially fixes the original issue but does not restore the cold routing contract, so the thread remains unresolved.

graphite-webgraph/src/main/kotlin/io/johnsonlee/graphite/webgraph/MappedWebGraphBackedGraph.kt
johnsonlee: The authoritative exact-head dual-baseline gate contradicts the local result and failed on 1d477784. Aggregate P95 still clears the 10x goal (worst 13.19x), but the required aligned-case gates are red: one dense case regresses versus v2.4.7 in all three pairs, and five current-main cases regress repeatedly (class-pair zero, callee-class zero/targeted, localized-early dense, and parameterized targeted). The candidate aggregate P95 is substantially lower than current main, but the per-shape non-regression contract is not met; this thread remains unresolved.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
johnsonlee: The new exact head cd7b139 fixes the previous 17-graph string failure, but the authoritative Method gate remains red after the cost moved to another covered shape. `method-compatibility-4-aggregate` now reports the 4-graph `order` CPU at 0.50s→0.93s (+86.0%), with reverse-order confirmation 0.67s→0.90s (+34.3%) against the 15% limit. Wall and RSS pass, but this is a reproduced CPU regression, so the Method thread still cannot be resolved.

johnsonlee (commented):
Blocking exact-head update for cd7b139: the Cypher capacity gate is newly red. All rejection/cancellation/recovery contract counters remain valid, but tail latency rises from 1.606s to 1.885s (+17.37%) against the hard 15% gate. The three raw forks are 1.440/1.726/1.652s on current main versus 2.088/1.970/1.597s on the candidate, so two of three forks are slower. This is separate from the Method regression and shows the optimization is shifting latency into the four-query concurrent capacity workload. GitHub cannot accept REQUEST_CHANGES from the PR author account; treat this COMMENT review as blocking until the exact-head authoritative capacity gate passes.

graphite-webgraph/src/main/kotlin/io/johnsonlee/graphite/webgraph/MappedWebGraphBackedGraph.kt
johnsonlee: The exact-head cd7b139 hosted gate removes the raw scans, but the replacement mapped-view route is still slower and now fails all lifecycle states. Cold request-selected P95 is 0.268→0.965 ms (+260%); K2 P95 0.463→1.680 ms, K8 1.236→2.468 ms, and K64 3.062→3.809 ms. Warm request-selected P95 is 0.176→0.598 ms and graph-id P95 0.305→0.939 ms; startup K8 P95 is 1.115→2.230 ms. Telemetry confirms zero raw scans but 64 mapped views and 464 prefix-cache entries, so the cost has shifted rather than the routing contract being restored. This thread remains unresolved.

graphite-webgraph/src/main/kotlin/io/johnsonlee/graphite/webgraph/MappedWebGraphBackedGraph.kt
johnsonlee: The authoritative dual-baseline gate also remains red on exact head cd7b139. Aggregate P95 still clears 10x (worst 12.81x), but `wrapped-case-insensitive/dense` regresses versus v2.4.7 in two pairs, and three aligned cases regress repeatedly versus current main: class-pair targeted, callee-class zero, and localized-early dense (the latter in all three pairs). This improves the earlier failure set but does not satisfy the per-shape gate, so the thread remains unresolved.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
johnsonlee: On exact head 7150c9b, the prior 4-graph aggregate and capacity failures are fixed, but the authoritative Method gate is still red in two other covered dimensions. The 4-graph scan `contains` CPU regression reproduces at 1.20s→1.41s (+17.5%), and the 17-graph scan `or` RSS-after regression reproduces at 2.610 GiB→3.163 GiB (+21.2%), both beyond the 15% limit. This is another cost shift rather than a fully restored Method envelope, so the thread remains unresolved.

johnsonlee: Exact-head hosted evidence is now green on cd038cf: all 12 Method shards pass, including the previously failing 4-scan, 17-scan, 4-aggregate, and 17-string cases, and the aggregate `method-compatibility-gate` succeeds. The separate capacity gate also passes. Under the repository gate contract this Method blocker is fixed, so I am resolving the thread.

graphite-webgraph/src/main/kotlin/io/johnsonlee/graphite/webgraph/MappedWebGraphBackedGraph.kt
johnsonlee: Exact-head cd038cf materially improves routing and K2/K8 now pass, but the authoritative gate is still red at the 64-graph boundary in two lifecycle states. Cold K64 is 0.358→0.838 ms P50 and 2.760→4.069 ms P95; startup-prepared K64 is 0.324→0.522 ms P50 and 2.630→3.680 ms P95. Raw scans remain zero, but the full-width mapped-view path still fails the non-regression contract, so this thread remains unresolved.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
johnsonlee: Reopening after head 19bae79: the Method gate regressed again after `Materialize cold raw projection once`. The 17-graph scan `contains` CPU increase reproduces at 3.02s→3.76s (+24.5%), and the 4-graph aggregate `count` CPU increase reproduces at 0.45s→0.65s (+44.4%), both beyond the 15% limit. The exact-head aggregate Method gate is red, so the blocker is active again.

graphite-webgraph/src/main/kotlin/io/johnsonlee/graphite/webgraph/MappedWebGraphBackedGraph.kt
johnsonlee: Head 19bae79 fixes the prior startup K64 failure and all other routing rows, but cold K64 still fails: P50 is 0.427→0.775 ms and P95 3.028→4.045 ms. This is narrower than cd038cf, yet the authoritative routing gate remains red, so the thread stays unresolved.

graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
johnsonlee: Exact-head 2adf1e0 has restored the full Method envelope: all 12 Method shards and the aggregate `method-compatibility-gate` pass, alongside the capacity gate. The prior 19bae79 confirmed CPU regressions are no longer present under the authoritative same-head run, so I am resolving this blocker again.

graphite-webgraph/src/main/kotlin/io/johnsonlee/graphite/webgraph/MappedWebGraphBackedGraph.kt
johnsonlee: The authoritative exact-head graph-routing gate now passes on 2adf1e0. The previous cold K64 and lifecycle-state regressions are absent, while unit, correctness, resource, and worker-policy checks also pass. Under the repository gate contract this routing blocker is fixed, so I am resolving the thread.

graphite-webgraph/src/main/kotlin/io/johnsonlee/graphite/webgraph/MappedWebGraphBackedGraph.kt
johnsonlee: Exact-head update for `2adf1e0`: the authoritative global-wide dual-baseline gate is still red, so this thread remains unresolved. Aggregate P95 continues to meet the >10x goal versus v2.4.7 in every fork (11.62x–15.63x), but the aligned per-shape safety gate has reproducible failures. Versus v2.4.7, `wrapped-case-insensitive/dense` is 2.589 ms → 5.449 ms in all three pairs. Versus current main, five shapes regress in at least two forks: `provenance/zero` 5.079→6.901 ms, `provenance/targeted` 6.154→7.204 ms, `localized-early/dense` 3.512→7.161 ms (all three), `four-properties/targeted` 25.371→30.484 ms, and `aliased/targeted` 9.914→12.507 ms. Method, capacity, and graph-routing gates now pass on this head; global-wide is the remaining benchmark blocker.

PR #116, opened 2026-09-05
Bound shared Cypher workers and propagate task cancellation

johnsonlee:
This PR changes Cypher runtime scheduling and cancellation. Benchmark/CI policy, harnesses and their tests have been extracted to #125; this PR is stacked on `codex/benchmark-query-gates` so those changes are excluded from its review diff. Merge #125 first, then retarget this PR to main.

## Runtime changes

- Introduce a shared `GraphTaskScheduler` for graph scans, storage segments, parallel sorting and guarded requests. Resource-owned live workers, including idle workers, are bounded to `availableProcessors()`. Unrelated JVM threads and externally owned callers are outside that resource bound.
- Update `QueryPipeline` and `MethodQueryExecutor` task submission, completion and failure handling. Preserve source order and wait for callbacks and descendant tasks to actually exit.
- Integrate request admission and task lifecycle management in `CypherQueryGuard`.
- Propagate task-group cancellation through mapped graph scanning and string-index filtering, checksum loading, posting validation and lazy posting merge. An already-running callback may finish, but cancellation stops remaining work.

The production diff comprises seven files across core, cypher, explore and webgraph, with associated runtime regression tests. Historical profiling reports and raw experiment artifacts remain in this PR; they are excluded from #125.

## Validation

The split was checked mechanically: all production code and runtime tests are byte-for-byte unchanged from `a6706055a72df48cf29d353290ba3728f30ef38e`, and all 28 extracted benchmark/CI files match #125 exactly. The only merge conflict was benchmark documentation, resolved to #125's current acceptance policy.

Runtime tests cover bounded workers, cancellation, blocked callbacks, descendant completion, partial submission, result ordering and complete retry results. Current-head CI remains required. Method-level and end-to-end performance regression conclusions, including the warmed per-query P50/P95 acceptance checks supplied by #125, remain pending; no performance gain or complete benchmark pass is claimed.

This PR remains a draft for separate runtime review. No merge is authorized by the split.


.github/scripts/benchmark-global-iteration.mjs
```diff
@@ -0,0 +1,280 @@
...
+    const progressErrors = [];
+    if (accepted?.error !== null || !accepted.status.regressionPassed || accepted.status.runs.length !== 3) {
+        progressErrors.push("Valid passing paired evidence against the last accepted iteration is required for progress");
+    } else {
+        accepted.status.runs.forEach((run, index) => {
+            const baseP95 = run?.baseP95LatencyNanos;
+            const candidateP95 = run?.p95LatencyNanos;
+            if (!Number.isFinite(baseP95) || !Number.isFinite(candidateP95) || baseP95 <= 0 || candidateP95 <= 0) {
+                progressErrors.push(`last-accepted pair-${index + 1}: finite positive P95 measurements are required`);
+            } else if (candidateP95 >= baseP95) {
```
johnsonlee, line 164: [P1] Do not treat any positive point-estimate delta as proven iteration progress. This branch's exact-head control built byte-identical base/candidate JMH JARs (`ddfed313…`) yet measured per-pair P95 ratios of 2.02x, 1.05x, and 0.34x, demonstrating that run direction alone leaves large variance. The current condition would accept the same binaries as an optimization whenever noise happens to make all three candidate point estimates lower (the added test explicitly accepts 100 → 99.999). That can advance `lastAcceptedRef` without an actual improvement and turn noise into the next baseline. Require a material/noise-floor or statistically supported improvement before setting `progressAchieved`; keep exact equality/noise as non-regression only.

johnsonlee: Current exact head `34b45bf` reproduces the problem with the same canonical JAR content hash on both sides (`ddfed3136e1c…`). The global gate reports three aggregate P95 ratios of 0.93x, 1.20x, and 0.86x for identical code, while Method compatibility independently blocks identical code on CPU (+30.8%, confirmed +32.4% for 36-graph OR; +33.9%, confirmed +41.8% for 4-graph COUNT). This is direct evidence that the current sampling/acceptance model can classify scheduler/host variance as either progress or regression. The thread remains blocking: iteration progress must require a material/noise-floor or statistically supported improvement, not merely candidate P95 < baseline P95 in each point estimate.

johnsonlee:
Reviewed exact draft head `23dafb3`. Keeping this as a draft is correct: local pair 3 regresses P95 (51.742 → 52.154 ms), hosted `run-unit-test` fails the 98% coverage gate in `core` (97.7025%) and `explore` (97.9092%), and `method-compatibility-4-scan` is already red while the pressure gates are still running.

The existing strict-progress thread remains unresolved: `.github/scripts/benchmark-global-iteration.mjs` still sets `progressAchieved` from any `candidateP95 < baseP95`, and its test still explicitly accepts 100 → 99.999. Do not advance `lastAcceptedRef` from this candidate; fix the material/noise-floor requirement and the exact-head CI failures first.

johnsonlee:
The authoritative run has now completed for exact head `23dafb3`: **FAIL (9/11)**. The global-wide candidate does not merely miss strict progress; it regresses against frozen main in two paired orders: worst P95 is 130.781 → 281.890 ms (0.46x), pair 3 is 141.640 → 150.614 ms (0.94x), and process CPU exceeds the 15% limit in pairs 1 and 2. Method compatibility is missing after the failed 4-scan shard, and the unit/coverage check is also red.

This head must remain an unaccepted draft and must not become `lastAcceptedRef`; the hosted exact-head result supersedes the local candidate table in the PR body.

johnsonlee:
[P1] Keep the raw profiling corpus out of the Git history. Relative to main this branch adds 1,413 files and 2,343,396 lines; 911 added paths are logs, JSON, HTML, text, class, JFR, or compressed captures, including 28 binary blobs. The added blobs total 135,034,951 bytes (128.8 MiB), so merging this permanently adds roughly 129 MiB to every full clone even though most of it is reproducible diagnostic output. Please retain the compact conclusions, commands, scripts, checksums, and small audited summaries in docs, and publish raw JFR, HTML, class, gzip, archive, and bulk JSON evidence as Actions artifacts or a release asset with immutable hashes and links.

PR #117, opened 2026-09-06
Serve CallSite string queries from the mapped view without worker pools

johnsonlee:
## Summary

Removes the CallSite graph/segment worker pools and answers every CallSite string projection serially on the requesting thread from the persisted `graph.callsite-string-index` sidecar. The goal is a 10x global-wide query P95 relative to `main` (`4e328b0`) without giving up existing functionality or the graph-routing performance.

- **Mapped-view engine**: a lazy planner probes trigram postings through an in-heap trigram directory, verifies only the rarest trigram span, validates only the selected posting ranges, and answers dense terms from a bounded raw storage prefix. The probe stops at the first checkpoint whose running match rate cannot fill `LIMIT` inside its budget.
- **Sidecar-less graphs** build their index once on the requesting thread, persist it, and serve the same view; no retained heap index outlives a request in the cold or warm states.
- **Repeated bounded projections** reuse per-graph row, posting-count, and absent-term caches and share their public rows through the existing `DirectProjectionResultCache`, so a routed dense hit costs the same 0.04 ms steady state as on `main`.
- **Query pipeline**: serial mapped CallSite row and `DISTINCT` paths (with complete cross-graph provenance via a shared tuple set) replace the parallel scanners, the NCPU split, and the `callSiteScanParallelism` property. The non-`DISTINCT` fast path no longer releases a zero-hit graph's rebuildable caches, matching `main`.
- **Storage helpers**: block-aware front-coded string-table search and a cached `groupedBy` on the shared tuple set.

## Gate changes

- `benchmark-gate.mjs`: candidate contract is the serial `0+0` worker plan with zero peaks and zero per-row scans; the graph-routing lifecycle now requires exactly 2,043 mapped-view lookups distributed 30..39 per graph in the cold and warm forks (base counters missing on `main` read as zero) and the unchanged retained-index lifecycle in `startup-prepared`; global-wide minimum speedup raised to `10x`, with wrapped shapes that do not carry the base P95 gated by `--minimum-wrapped-speedup`. 87 node tests pass; workflow comparator hashes re-pinned.
- Docs: `large-broad-query-pressure-benchmark.md` and Attempt 132 in the optimization attempts log.

## Local evidence (fixture64, 8 GiB heap, 4 CPUs)

| Fork | main P95 | candidate P95 |
|---|---:|---:|
| global-wide cold, three fresh JVMs each | 213.6 / 210.8 / 211.9 ms | 18.4 / 20.1 / 21.6 ms |

[...]

PR #118, opened 2026-09-07
Keep the resource benchmark's heap sampler running through the retained read

johnsonlee:
## Summary

Gate-owned repair for the `wrapped-query-resources` check, split out of #117 as requested there.

`WrappedDiscoveryResourceBenchmark` stopped its heap sampler in `finishQuery` and again at the start of `tearDownInvocation`, before the teardown collections and before `retainedHeapBytes` was read. The retained read could therefore exceed every earlier sample whenever the last sample missed the allocation region the retained read then counted, and the comparator's loaded/retained/peak relationship check failed on identical code (`invalid loaded/retained/peak heap relationship` on the 36-graph footprint, in the run and its confirmation, across the recent #117 heads).

The sampler now runs until the retained value is read:

- `finishQuery` records the peak so far through a new `peak()` (a sample taken now folded into the maximum, sampling continues) instead of stopping.
- `tearDownInvocation` runs the forced collections, reads the retained value, then folds the sampler's final value and the retained read into `peakUsedHeapBytes`, so the compared interval is consistent and the peak can only enlarge.
- `REAL_ONLY_RESOURCE_HARNESS_SHA256` in the workflow follows the new file content (the gate test pins it).

The change only enlarges the observed peak; the retained and loaded readings are unchanged.

## Validation

- `:webgraph:compileJmhKotlin` and `:webgraph:detekt` pass.
- The 113 gate tests pass with the updated pin.
- On this PR the base-owned control installed from `gate/` is still `main`'s harness, so the run here does not exercise the change; the main-vs-main validation is the first PR against `main` after this lands.

---


johnsonlee (commented):
[P1] Exercise the candidate harness before making it the trusted control

This PR’s benchmark workflow replaces both compared copies with the harness from `main`, so none of the completed checks executes `peak()` or the new teardown ordering from `305a4a5`. Compilation, detekt, the updated SHA pin, and comparator unit tests cannot catch a runtime lifecycle error here; waiting for the first PR after merge makes every subsequent resource gate depend on unvalidated control code. Add a candidate-only contract/smoke job (non-authoritative for product comparison) that runs this exact harness on the real persisted fixture workload and asserts the published `loaded ≤ retained ≤ peak` relationship, including that sampling remains active between `finishQuery` and the retained read. Until that passes on this exact head, this gate-control change is not ready to land.

.github/workflows/benchmark.yml
```diff
@@ -22,7 +22,7 @@ env:
   BENCHMARK_REPORT_TRANSITION_SHA256: 4095d5756d5985b0f6fcf4ef56caaf7a14a699bef117e985dec3a6ec78848f51
   REAL_ONLY_LATENCY_COMPARATOR_SHA256: 4095d5756d5985b0f6fcf4ef56caaf7a14a699bef117e985dec3a6ec78848f51
   LATENCY_POINT_ESTIMATE_COMPARATOR_SHA256: 4095d5756d5985b0f6fcf4ef56caaf7a14a699bef117e985dec3a6ec78848f51
-  REAL_ONLY_RESOURCE_HARNESS_SHA256: 8ff5f47c264b4502589b40630039003424b75383ab4349e7080ac5f8619d0739
+  REAL_ONLY_RESOURCE_HARNESS_SHA256: 67a9df7bc1b9615d16886e1e091b10bbf567bdbcb9cf499934f07a5ddcf3617f
```
johnsonlee, line 25: [P1] Make this pin govern a candidate-harness validation

At the current base (`4e328b0`), `gate/WrappedDiscoveryResourceBenchmark.kt` no longer contains `SingleGraphWrappedDiscoveryResourceBenchmark`, so the transition branch at lines 185–191 never selects `controls`; reference, base, and candidate JARs are all built with the base-owned harness. Consequently this updated hash is inert in this PR and the green `wrapped-query-resources` check did not execute `peak()` or the new teardown ordering. I independently built the four real fixture graphs and ran the exact-head harness for 10 invocations—all satisfied `loaded ≤ retained ≤ peak`, and a same-machine comparator control passed—but the repository safety rule requires the authoritative benchmark gate. Add a candidate-only real-fixture contract/smoke job that builds this exact harness and fails closed on the raw per-invocation relationship before promoting it to the trusted base control.

johnsonlee (commented):
Re-reviewed exact head `305a4a5544cf82080226f188d2dd7f08ffe511c0`.

Independent verification passed compilation, detekt, all 113 gate tests, real-JAR graph preparation for Android/Tika/Hive/Kotlin compiler, and a 36-mapped-graph candidate-harness run. Ten raw invocations all satisfied `loaded ≤ retained ≤ peak`, and a same-machine `main`/candidate `compare-latency-resources` control passed.

That establishes that the implementation can run, but it does not clear the inline blocker: the authoritative PR workflow still replaces the candidate harness with `main` and therefore provides no exact-head benchmark-gate evidence for this control-plane change. The aggregate benchmark gate is also currently FAIL.

.github/workflows/benchmark.yml
johnsonlee: 撤回这个 P1。这里把产品变更的 gate 要求错误地套到了 gate-control bootstrap 上，形成了循环依赖。对这个 PR，正确的验证路径是独立 meta-validation：exact-head 编译和 113 个 comparator/gate tests；从 Android、Tika、Hive、Kotlin compiler 四个真实 JAR 构建 persisted graphs；运行修改后的 36-graph harness；检查 raw invocation 契约。该验证已完成：10/10 invocation 满足 `loaded ≤ retained ≤ peak`，同机 main/candidate comparator control PASS；字节码也确认新顺序为 `peak()`（继续采样）→ GC → retained read → `stop()` → `max(..., retained)`。因此这个 comment 不再是 blocker。

PR #119, opened 2026-09-07
Measure request-serving Java-thread CPU in the method-compatibility harness

johnsonlee:
## Summary

Gate-owned repair for the `method-compatibility` shards, split out of #117 as requested there. Harness part only; the workflow metric switch follows in a separate PR once this harness is the base-installed control (see below).

The shards gate each request row on `processCpuNanos`, which folds the concurrent collector and compiler threads into the row. On the hosted runner that figure swings between runs of identical code: over 22 rows of the recent #117 heads the process-CPU deltas ranged from −32% to +57% (stdev 22%) while the wall rows stayed within −9%..+13% (stdev 6%), and every red shard was a process-CPU row with all wall rows passing.

`MethodDiscoveryCompatibilityBenchmark` now records, next to the process figure, a request-serving Java-thread CPU figure measured **by Java-thread identity**:

- `javaThreadCpuNanos`: the sum of each Java thread's own `ThreadMXBean.getThreadCpuTime` over the measured window, keyed by `Thread.getId()`. The JVM's internal native threads (collectors, compilers, the VM and service threads) have no `Thread` object and no thread id, so they are never in the sum. There is no name matching and no `/proc` internal set, so a request worker whose native name collides with a HotSpot thread (`/proc/<tid>/comm` is mutable and truncated to 15 bytes) cannot be mistaken for one and dropped.
- `gcTimeMillis` and `jitTimeMillis`: reported so a process-CPU swing can still be attributed.

A thread present at both ends of the window contributes its after-minus-before delta; one born inside the window and still alive at the end contributes its whole lifetime CPU. The two cases the two snapshots cannot see are handled by **failing closed** rather than under-reporting:

- a Java thread present at the start that is gone at the end took its in-window CPU with it;
- a Java thread created and finished entirely inside the window is in neither snapshot, but `ThreadMXBean.getTotalStartedThreadCount` still counted its start, so more starts than newly-present threads means such a worker existed.

Every request-serving pool in the engine (`callSiteScanExecutor`, `splitCallSiteExecutor`, `directStringExecutor`, the `methodGraphScanPool` `ForkJoinPool`, and `CypherQueryGuard`'s executors) is a persistent daemon pool created lazily and reused, so the real path has no born-and-died-inside-the-window worker; the fail-closed tripwire is a guard against a candidate introducing one to look cheaper. `processCpuNanos` is kept as an advisory figure (read innermost so its interval is contained in the thread snapshots); `jvmInternalCpuNanos` is the residual `process − java`.

## Landing order

The benchmark workflow installs `ExplorerMemoryBenchmark.kt` from `gate/` (the base SHA) over the candidate's copy, so this PR's own run still executes `main`'s harness on both sides and gates on the process figure as before. After this lands, a second PR switches the workflow to gate on `javaThreadCpuNanos` and keeps `processCpuNanos` as an advisory row; that PR's run is the main-vs-main validation of this harness (the base-installed control on both sides).

## Contract

`MethodCompatibilityCpuAccountingContract` runs in a JVM of its own before the shards, never inside a benchmark trial:

- **negative contract for the old accounting**: a persistent worker named `Service Thread` (colliding with a HotSpot native name, truncated to 15 bytes) must be charged in full, not subtracted. Locally: `worker 200,267,104 ns, accounted 201,195,741 ns` — the name-based code hid ~96% of it.
- a worker created and joined entirely inside the window must make `measure()` fail closed.
- the request-serving Java sum may exceed the process figure only by snapshot overhead (two process-clock ticks), idle and under a collector load.

## Validation

- `:explore:compileJmhKotlin` and `:explore:detekt` pass; the contract main passes deterministically.
- The tripwire does not fire on the real request path: 40 served requests through Javalin/Jetty + `HttpURLConnection`, and 40 `ForkJoinPool` + `ExecutorCompletionService` fan-outs (the engine's patterns), each showed zero transient Java threads.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -757,6 +757,107 @@ jobs:
         if-no-files-found: error
         retention-days: 14
 
+  validate-cpu-accounting:
+    name: validate-cpu-accounting
+    needs: [prepare-latency-fixtures]
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+
+    steps:
+    - name: Checkout PR
+      uses: actions/checkout@v7
+      with:
+        ref: ${{ github.event.pull_request.head.sha }}
+        path: candidate
+        persist-credentials: false
+
+    - name: Setup Java
+      uses: actions/setup-java@v5
+      with:
+        distribution: temurin
+        java-version: '17'
+
+    - name: Setup Gradle
+      uses: gradle/actions/setup-gradle@v6
+      with:
+        cache-read-only: true
+
+    # The candidate's own Explorer harness is built WITHOUT the base-owned gate install, so this jar
+    # runs the new RequestCpuAccounting integration -- the real forks in the paired gate do not, since
+    # they install the base harness over both trees. This proves the accounting in a real fork; it is
+    # candidate-only and never feeds the trusted paired gate.
+    - name: Build candidate-owned Explorer JMH JAR
+      run: |
+        candidate/gradlew -p candidate :explore:testClasses :explore:jmhJar --no-daemon
+        mapfile -t JARS < <(find candidate/graphite-explore/build/libs -name '*-jmh.jar' -type f)
+        test "${#JARS[@]}" -eq 1
+        jar tf "${JARS[0]}" >/dev/null
+        mkdir -p jmh-candidate-native
+        install -m 0644 "${JARS[0]}" jmh-candidate-native/candidate-native-explore-jmh.jar
+
+    - name: Restore persisted fixture graphs
+      uses: actions/cache/restore@v5
+      with:
+        path: fixture-graphs
+        key: ${{ runner.os }}-${{ runner.arch }}-wrapped-query-fixtures-v2-${{ hashFiles('candidate/gradle/libs.versions.toml', 'candidate/gradle/wrapper/gradle-wrapper.properties', 'candidate/settings.gradle.kts', 'candidate/**/build.gradle.kts', 'candidate/graphite-core/src/main/**', 'candidate/graphite-sootup/src/main/**', 'candidate/graphite-webgraph/src/main/**', 'candidate/graphite-webgraph/src/jmh/kotlin/io/johnsonlee/graphite/webgraph/BenchmarkCorpus.kt', 'candidate/graphite-webgraph/src/jmh/kotlin/io/johnsonlee/graphite/webgraph/AllFixtureBenchmarkGraphPreparation.kt') }}
+
+    - name: Verify restored real graphs
+      run: |
+        test -f fixture-graphs/.complete
+        for CORPUS in android tika hive kotlin-compiler; do
+          test -f "fixture-graphs/${CORPUS}/graph.metadata"
+          test -f "fixture-graphs/${CORPUS}/graph.nodedata"
+        done
+
+    - name: Exercise the candidate CPU accounting in a real fork
+      run: |
+        mkdir -p benchmark-results
+        JAR=jmh-candidate-native/candidate-native-explore-jmh.jar
+        GRAPH_ARGS="-Dandroid.graph.path=${PWD}/fixture-graphs/android \
+          -Dtika.graph.path=${PWD}/fixture-graphs/tika \
+          -Dhive.graph.path=${PWD}/fixture-graphs/hive \
+          -Dkotlin.compiler.graph.path=${PWD}/fixture-graphs/kotlin-compiler"
+        # The accounting contract must still hold on this runner/JVM before the fork.
+        java -cp "${JAR}" io.johnsonlee.graphite.cli.MethodCompatibilityCpuAccountingContract
+        java -jar "${JAR}" \
+          'io.johnsonlee.graphite.cli.MethodDiscoveryCompatibilityBenchmark.methodScenarioGate' \
+          -wi 0 -i 1 -f 1 -t 1 -foe true -rf json \
+          -p graphCount=4 -p scenario=count \
+          -jvmArgsAppend "${GRAPH_ARGS} -Dgraphite.method.compatibility.output=${PWD}/benchmark-results/candidate-cpu-accounting-smoke.txt" \
+          -rff "${PWD}/benchmark-results/candidate-cpu-accounting-smoke.json"
+
+    - name: Assert the real fork published a valid javaThreadCpuNanos row
+      run: |
+        FILE=benchmark-results/candidate-cpu-accounting-smoke.json
+        EXPECTED='io.johnsonlee.graphite.cli.MethodDiscoveryCompatibilityBenchmark.methodScenarioGate'
+        jq -e --arg expected "${EXPECTED}" '
+          length == 1 and
+          all(.[];
+            .benchmark == $expected and
+            .params.graphCount == "4" and
+            .params.scenario == "count" and
+            .mode == "ss" and
+            .primaryMetric.score > 0 and
+            .secondaryMetrics.requestsSucceeded.score == 1 and
+            .secondaryMetrics.javaThreadCpuNanos.score > 0 and
+            .secondaryMetrics.processCpuNanos.score > 0 and
+            .secondaryMetrics.jvmInternalCpuNanos.score >= 0 and
+            .secondaryMetrics.cpuAccountingUnderflowNanos.score >= 0 and
+            .secondaryMetrics.jvmInternalThreadsEnded.score >= 0 and
+            .secondaryMetrics.gcTimeMillis.score >= 0 and
+            .secondaryMetrics.jitTimeMillis.score >= 0
+          )' "${FILE}" >/dev/null
+        echo "cpu-accounting smoke: candidate fork published a valid javaThreadCpuNanos row"
+
+    - name: Upload CPU accounting smoke result
+      if: always()
+      uses: actions/upload-artifact@v7
+      with:
+        name: benchmark-cpu-accounting-smoke-${{ github.event.pull_request.number }}-${{ github.run_attempt }}
+        path: benchmark-results/
+        if-no-files-found: error
+        retention-days: 14
+
   method-compatibility:
     name: method-compatibility-gate
     if: always()
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -608,6 +608,9 @@ jobs:
           -Dkotlin.compiler.graph.path=${PWD}/fixture-graphs/kotlin-compiler"
         METHOD_ARGS=(-wi 0 -i 1 -f 1 -t 1 -foe true -rf json \
           -p "graphCount=${GRAPH_COUNT}" -p "scenario=${SCENARIOS}")
+        # The request-serving CPU accounting contract runs in a JVM of its own, on this runner and
+        # JVM, before any measured fork; it must never run inside a benchmark trial.
+        java -cp "${CANDIDATE_JAR}" io.johnsonlee.graphite.cli.MethodCompatibilityCpuAccountingContract
         java -jar "${BASE_JAR}" \
           'io.johnsonlee.graphite.cli.MethodDiscoveryCompatibilityBenchmark.methodScenarioGate' "${METHOD_ARGS[@]}" \
           -jvmArgsAppend "${GRAPH_ARGS} -Dgraphite.method.compatibility.output=${PWD}/benchmark-results/base-method-compatibility-${GRAPH_COUNT}-${GROUP}.txt" \
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -2808,7 +2808,7 @@ jobs:
   benchmark-regression-gate:
     name: benchmark-regression-gate
     if: always()
-    needs: [candidate-gate-tests, method-level, explorer, method-compatibility, cypher-capacity, budgeted-collection, budgeted-mapped-string, large-corpus, wrapped-query-latency, wrapped-query-resources, graph-routing-pressure-evidence, global-wide-pressure-evidence]
+    needs: [candidate-gate-tests, method-level, explorer, method-compatibility, cypher-capacity, budgeted-collection, budgeted-mapped-string, large-corpus, wrapped-query-latency, wrapped-query-resources, graph-routing-pressure-evidence, global-wide-pressure-evidence, validate-cpu-accounting]
     runs-on: ubuntu-latest
 
     steps:
@@ -2920,6 +2920,7 @@ jobs:
         LATENCY_RESOURCES_JOB: ${{ needs.wrapped-query-resources.result }}
         GRAPH_ROUTING_JOB: ${{ needs.graph-routing-pressure-evidence.result }}
         GLOBAL_WIDE_JOB: ${{ needs.global-wide-pressure-evidence.result }}
+        CPU_ACCOUNTING_SMOKE_JOB: ${{ needs.validate-cpu-accounting.result }}
       run: |
         if [ "${CANDIDATE_GATE_TEST_JOB}" != success ] || \
           [ "${METHOD_JOB}" != success ] || [ "${EXPLORER_JOB}" != success ] || \
@@ -2929,7 +2930,8 @@ jobs:
           [ "${BUDGETED_STRING_JOB}" != success ] || \
           [ "${LARGE_CORPUS_JOB}" != success ] || \
           [ "${LATENCY_JOB}" != success ] || [ "${LATENCY_RESOURCES_JOB}" != success ] || \
-          [ "${GRAPH_ROUTING_JOB}" != success ] || [ "${GLOBAL_WIDE_JOB}" != success ]; then
+          [ "${GRAPH_ROUTING_JOB}" != success ] || [ "${GLOBAL_WIDE_JOB}" != success ] || \
+          [ "${CPU_ACCOUNTING_SMOKE_JOB}" != success ]; then
           echo "::error::Benchmark execution or comparison failed"
           exit 1
         fi
--- a/graphite-explore/src/jmh/kotlin/io/johnsonlee/graphite/cli/ExplorerMemoryBenchmark.kt
+++ b/graphite-explore/src/jmh/kotlin/io/johnsonlee/graphite/cli/ExplorerMemoryBenchmark.kt
@@ -786,15 +786,18 @@ open class MethodDiscoveryCompatibilityBenchmark {
     }
 
     private fun measure(counters: MethodCompatibilityCounters, action: () -> Long): Long {
-        val beforeInternal = internalThreadCpuNanos()
-        val beforeCpu = processCpuTimeNanos()
+        // The subtracted internal interval is nested inside the process interval: the process
+        // figure is read first before the action and last after it, so any internal CPU spent in
+        // the snapshot gaps is charged to the request rather than subtracted from it.
+        val beforeRss = residentSetBytes()
         val beforeGcTime = gcTimeMillis()
         val beforeJit = jitTimeMillis()
-        val beforeRss = residentSetBytes()
+        val beforeCpu = processCpuTimeNanos()
+        val beforeInternal = internalThreadCpuNanos()
         val bytes = action()
-        val afterRss = residentSetBytes()
-        val afterCpu = processCpuTimeNanos()
         val afterInternal = internalThreadCpuNanos()
+        val afterCpu = processCpuTimeNanos()
+        val afterRss = residentSetBytes()
         counters.requestsSucceeded++
         counters.responseBytes += bytes
         counters.processCpuNanos = (afterCpu - beforeCpu).coerceAtLeast(0L)
@@ -808,7 +811,11 @@ open class MethodDiscoveryCompatibilityBenchmark {
         val internalCpu = afterInternal.entries.sumOf { (tid, cpu) -> cpu - (beforeInternal[tid] ?: 0L) }
         counters.jvmInternalCpuNanos = internalCpu.coerceAtLeast(0L)
         counters.jvmInternalThreadsEnded = beforeInternal.keys.count { it !in afterInternal }.toLong()
-        counters.javaThreadCpuNanos = (counters.processCpuNanos - counters.jvmInternalCpuNanos).coerceAtLeast(0L)
+        // With the nesting above the remainder can only fall below zero by the process clock's
+        // tick resolution; the shortfall is published instead of being hidden by the clamp.
+        val remainder = counters.processCpuNanos - counters.jvmInternalCpuNanos
+        counters.cpuAccountingUnderflowNanos = (-remainder).coerceAtLeast(0L)
+        counters.javaThreadCpuNanos = remainder.coerceAtLeast(0L)
         counters.gcTimeMillis = (gcTimeMillis() - beforeGcTime).coerceAtLeast(0L)
         counters.jitTimeMillis = (jitTimeMillis() - beforeJit).coerceAtLeast(0L)
         counters.residentSetBeforeBytes = beforeRss
@@ -873,6 +880,35 @@ open class MethodDiscoveryCompatibilityBenchmark {
             "Request-serving CPU accounting missed a worker that lived inside the window: " +
                 "worker ${expected} ns, accounted ${counters.javaThreadCpuNanos} ns, " +
                 "process ${counters.processCpuNanos} ns, internal ${counters.jvmInternalCpuNanos} ns"
+        }
+        verifySnapshotNesting("idle action") { 0L }
+        verifySnapshotNesting("collector-loaded action") {
+            // Enough short-lived garbage to run the collector's threads inside the window, so the
+            // internal share is non-trivial while the interval containment is checked.
+            val retained = ArrayList<ByteArray>(CPU_ACCOUNTING_GARBAGE_CHUNKS)
+            repeat(CPU_ACCOUNTING_GARBAGE_CHUNKS) { index ->
+                val chunk = ByteArray(CPU_ACCOUNTING_GARBAGE_CHUNK_BYTES)
+                chunk[index % chunk.size] = index.toByte()
+                if (retained.size >= CPU_ACCOUNTING_GARBAGE_RETAINED) retained.removeAt(0)
+                retained.add(chunk)
+            }
+            retained.size.toLong()
+        }
+    }
+
+    /**
+     * Contract check for the snapshot gaps: because the internal interval is read inside the
+     * process interval, the subtracted internal CPU can exceed the process figure only by the
+     * process clock's tick, whatever the JVM's threads did around the action. A larger shortfall
+     * means the intervals are no longer nested and the row would under-report, so it fails.
+     */
+    private fun verifySnapshotNesting(label: String, action: () -> Long) {
+        val counters = MethodCompatibilityCounters()
+        measure(counters, action)
+        check(counters.cpuAccountingUnderflowNanos <= CPU_ACCOUNTING_MAX_UNDERFLOW_NANOS) {
+            "Internal CPU exceeded the process interval on the $label: " +
+                "process ${counters.processCpuNanos} ns, internal ${counters.jvmInternalCpuNanos} ns, " +
+                "shortfall ${counters.cpuAccountingUnderflowNanos} ns"
         }
     }
 
@@ -1032,6 +1068,9 @@ open class MethodBenchmarkCompatibilityCounters {
 
     @JvmField
     var jvmInternalThreadsEnded: Long = 0
+
+    @JvmField
+    var cpuAccountingUnderflowNanos: Long = 0
 
     @JvmField
     var gcTimeMillis: Long = 0
@@ -1983,6 +2022,11 @@ private const val EXPLICIT_GC_CAUSE = "System.gc()"
 private const val END_OF_MAJOR_GC = "end of major GC"
 private const val END_OF_MINOR_GC = "end of minor GC"
 private const val CPU_ACCOUNTING_WORKER_NANOS = 200_000_000L
+private const val CPU_ACCOUNTING_GARBAGE_CHUNKS = 512
+private const val CPU_ACCOUNTING_GARBAGE_CHUNK_BYTES = 1 shl 20
+private const val CPU_ACCOUNTING_GARBAGE_RETAINED = 64
+/** Two ticks of the process CPU clock (10 ms on Linux), the only slack the nesting leaves. */
+private const val CPU_ACCOUNTING_MAX_UNDERFLOW_NANOS = 20_000_000L
 private const val CPU_ACCOUNTING_MIN_SHARE_PERCENT = 90L
 private const val PERCENT = 100L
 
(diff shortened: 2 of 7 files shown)


johnsonlee (commented):
[P1] Run the new measurement path on this exact head before landing it

The PR workflow installs `ExplorerMemoryBenchmark.kt` from `main` over both compared trees, so the completed checks never invoke `quiesceBeforeMeasurement()` and never publish `javaThreadCpuNanos`, `gcTimeMillis`, or `jitTimeMillis` from `466cf09`. Compile/detekt only prove syntax; they do not prove that the HotSpot option and GC-notification contract works on the hosted JVM, that the real HTTP scenario has stable worker lifetimes, or that the new CPU counter is present and nonzero. Add a candidate-only smoke/contract job that runs the candidate harness against representative real 4/17/36-graph scenarios and fails closed on a missing notification/metric or ended worker. The later workflow-switch PR can provide main-vs-main performance evidence, but runtime validation of the trusted harness itself is required before this PR lands.

graphite-explore/src/jmh/kotlin/io/johnsonlee/graphite/cli/ExplorerMemoryBenchmark.kt
```diff
@@ -784,13 +785,33 @@ open class MethodDiscoveryCompatibilityBenchmark {
...
+        val afterIds = threads.allThreadIds.toSortedSet()
         counters.requestsSucceeded++
         counters.responseBytes += bytes
         counters.processCpuNanos = (processCpuTimeNanos() - beforeCpu).coerceAtLeast(0L)
+        // CPU of the Java threads that serve the request (client, Jetty workers, scan pools),
+        // which excludes the concurrent collector and compiler threads that make the process
+        // figure swing between runs of identical code; the two remainders are reported next to it.
+        // A thread that was alive at the start and is gone at the end took its CPU with it, so
+        // the row fails closed instead of under-reporting.
+        val ended = beforeIds - afterIds
```
johnsonlee: [P1] Account for workers born and terminated inside the request

`beforeIds - afterIds` only detects a thread that existed before the action and disappeared afterward. A worker created after `beforeIds` and joined before `afterThreadCpu` is absent from both snapshots, so its lifetime CPU is omitted and `ended` remains empty. I reproduced this against the exact-head `measure()` via reflection: an action that started and joined a CPU-bound worker for ~500 ms reported `processCpuNanos=516,197,000` but only `javaThreadCpuNanos=10,215,000`, with no exception—about 98% of the request CPU vanished. A candidate that introduces per-request or structured short-lived workers can therefore look cheaper and pass the future blocking gate. Accumulate CPU inside the request-serving task wrappers before they exit (or keep process CPU blocking with a robust repeated-pair protocol), and add a contract test whose action creates and joins a short-lived CPU worker.

johnsonlee (commented):
Re-reviewed exact head `466cf094ded1e6fca8f2ce369d765006359e0701` as a gate-control change.

The earlier runtime-validation concern is cleared by independent meta-validation: compile/detekt and all 113 gate tests pass; the exact-head harness successfully loaded the four real persisted corpora and published positive `javaThreadCpuNanos`, `gcTimeMillis`, and `jitTimeMillis` for representative 4/17/36-graph cases. Five-fork controls also show the intended stability improvement for 17/count (Java-thread CPU CV 1.30% versus process CPU CV 23.80%).

One separate correctness blocker remains: the two-snapshot algorithm cannot observe a worker whose entire lifetime is inside the request window.

graphite-explore/src/jmh/kotlin/io/johnsonlee/graphite/cli/ExplorerMemoryBenchmark.kt
johnsonlee: Agreed: a worker that lives entirely inside the window is in neither snapshot, so the two-snapshot sum could not see it. Rewritten in b50067a so lifetime no longer matters:

- The row starts from the process figure, which counts every thread that ran in the window, and subtracts the CPU of the JVM's own native threads (collectors, compilers, the VM and service threads), read per thread from the kernel's scheduler accounting (`/proc/self/task/<tid>/schedstat`, nanoseconds) and recognised by their HotSpot native names. The remainder is `javaThreadCpuNanos`: the request-serving Java threads, however short-lived. The internal share is published as `jvmInternalCpuNanos`; an internal thread that appears inside the window counts in full, and one the JVM retires inside it (it retires only idle compiler threads) is counted in `jvmInternalThreadsEnded` rather than dropped silently.
- The accounting fails closed when the per-thread accounting is unavailable or unreadable, and when no internal thread is found at all.
- Contract check on the live JVM: trial setup runs `measure()` once over an action that starts, runs for 200 ms and joins a CPU-bound worker, and requires at least 90% of the worker's own thread CPU to be charged to the row, otherwise the harness stops before the measured sequence. Your reproduction is the shape of that action. Locally the same accounting charges 351 ms for a 300 ms worker plus the calling thread's own work, against a process figure of 360–420 ms with 9–69 ms of internal CPU.

Compile and detekt pass; the workflow is unchanged in this PR, so the process figure still gates until the metric switch.

---

johnsonlee:
**`run-unit-test` red on b50067a (and on 466cf09): not this PR's**

The Gradle `check` completes (`BUILD SUCCESSFUL`, all module tests pass), and the job then fails in the "Enforce coverage threshold" step:

```
Module cypher coverage is 97.9975%, below 98% threshold
```

This PR changes only the explore module's JMH harness (`ExplorerMemoryBenchmark.kt`), which is neither production code nor a test; the cypher module's sources and tests are those of `main`. The same cypher code measured at or above 98% on #118's run and on the `main` push runs, so the module sits on the threshold and its figure moves by hundredths of a percent between runs of identical code. No fix exists on `main` yet for that margin; a test that covers a currently uncovered cypher branch would belong in its own change rather than in this harness PR. I am re-running the failed job once; if it lands below 98% again I will say so here and propose the cypher coverage patch separately.

The benchmark workflow on b50067a is green so far: all twelve method-compatibility shards and the gate, `wrapped-query-resources` (with `main` now carrying #118, this run is also that harness's first base-installed validation), capacity and wrapped-latency; large-corpus, routing and global-wide are still running.

---

johnsonlee:
Correction to the note above: the re-run request was refused (`403 Resource not accessible by integration`), so I cannot re-run the failed job from here; a re-run from the Actions page would settle whether this head clears 98% on the next draw. Meanwhile I am measuring the cypher module's coverage locally across repeated runs to identify the lines whose coverage moves between runs, so the fix can be a targeted test rather than a threshold change; I will propose it as a separate change.

---

johnsonlee:
**Where the cypher coverage moves** (three local Kover runs of `main`'s cypher module, `:cypher:koverXmlReport --rerun-tasks`):

| run | covered / missed lines | line coverage |
|---|---|---|
| 1 | 6215 / 127 | 97.9975% |
| 2 | 6215 / 127 | 97.9975% |
| 3 | 6216 / 126 | 98.0132% |

Exactly two lines differ between runs, both in `QueryPipeline.kt`'s direct-string worker cancellation: line 2526 (`cancelAndJoin`: `if (future.cancel(true) && taskStarted.compareAndSet(false, true)) completion.countDown()`, partially covered in every run) and line 2623 (`stopWorkers(cancel = true)`: the same count-down for a worker cancelled before it started), covered only in run 3. Both are reached only when a cancellation wins the race against a queued worker's start on the shared fixed pool, which the existing cancellation tests hit by timing. The module's total sits at 97.9975% when that race is lost, so the 98% threshold turns on one line; this head's two runs and #118's run drew opposite outcomes on identical cypher code.

A deterministic test would saturate the direct-string pool (more graph tasks than `graphite.cypher.directStringParallelism`, the first ones blocked on a latch inside the storage call) and cancel while the rest are still queued, so `stopWorkers`' before-start branch is taken every run. That is a cypher test change on `main`, separate from this harness PR; I can open it if you want it, otherwise the next run of this head decides by the same coin toss.

---

johnsonlee:
**`global-wide-pressure-evidence` red on b50067a: not this PR's, and no fix belongs here**

The gate compares base and candidate production code and requires a 5.00x P95 speedup in every paired fork (`main`'s comparator); this PR changes only the explore JMH harness, so both sides run identical production code and the pairs read 0.99x / 1.00x / 0.63x (base P95 88.7 / 109.5 / 112.2 ms against candidate 90.0 / 109.9 / 178.2 ms, single-shot first executions on the shared runner). The same check was red on #118 for the same reason and did not block it. `graph-routing-pressure-evidence` will end the same way when it completes (10x on the query-level graphId P95 against identical code), and the aggregate `benchmark-regression-gate` with it. The checks that this PR can affect are green: all twelve method-compatibility shards and the gate, `wrapped-query-resources`, capacity, wrapped latency, large-corpus.

---

graphite-explore/src/jmh/kotlin/io/johnsonlee/graphite/cli/ExplorerMemoryBenchmark.kt
```diff
@@ -784,20 +786,96 @@ open class MethodDiscoveryCompatibilityBenchmark {
     }
 
     private fun measure(counters: MethodCompatibilityCounters, action: () -> Long): Long {
+        val beforeInternal = internalThreadCpuNanos()
```
johnsonlee: [P1] Keep the subtracted interval inside the process interval

The two deltas currently cover different windows: `beforeInternal` is sampled before `beforeCpu`, while `afterInternal` is sampled after `afterCpu`. Therefore `afterInternal - beforeInternal` includes JVM-internal CPU consumed during both snapshot gaps that is absent from `afterCpu - beforeCpu`; subtracting it can under-report request CPU (and the final clamp can hide the error). The 200 ms worker check does not prove this boundary because the worker dominates a small edge error. Sample `beforeCpu` before `beforeInternal`, and `afterInternal` before `afterCpu`, so the subtracted interval is contained in the process interval and any edge CPU is conservatively charged rather than silently removed. Add a focused contract test for the snapshot gaps.

johnsonlee: Agreed, the internal interval reached outside the process interval on both ends. Fixed in c1a1a2e:

- `measure()` reads the process figure first before the action and last after it (`beforeCpu` → `beforeInternal` → action → `afterInternal` → `afterCpu`), so the subtracted interval is contained in the process interval and internal CPU in either gap is charged to the request rather than removed.
- The clamp no longer hides anything: the remainder's shortfall is published as `cpuAccountingUnderflowNanos`. With the nesting it can only come from the process clock's tick (10 ms on Linux) against the nanosecond per-thread reads.
- Contract check for the gaps, in trial setup next to the worker check: `measure()` runs over an idle action and over a collector-loaded action (512 MiB of short-lived garbage with 64 MiB retained, which puts the collector's threads to work inside the window), and each must keep the shortfall within two ticks (20 ms) or the harness stops before the measured sequence. Locally, 40 alternating windows with the same primitives: worst shortfall 1.05 ms idle and 0 ms loaded, with up to 59.8 ms of internal CPU inside the loaded windows.

Compile and detekt pass.

---

johnsonlee:
**`method-compatibility-36-aggregate` red on c1a1a2e: the defect this PR repairs, measured by the control it replaces**

The shard's process-CPU row for 36/count reads +43.8% (confirmation +23.4%) with every wall row passing, on identical production code and, since the workflow installs `main`'s harness on both sides, identical harness code as well. It is the process-CPU swing this PR exists to take out of the gate; this head's harness change cannot affect that shard until it is the base-installed control and the metric switch follows. The remaining reds on this head are the ones already noted above (global-wide's speedup requirement on identical code, and the cypher coverage margin in `run-unit-test`).

---

graphite-explore/src/jmh/kotlin/io/johnsonlee/graphite/cli/ExplorerMemoryBenchmark.kt
```diff
@@ -784,20 +786,132 @@ open class MethodDiscoveryCompatibilityBenchmark {
...
+        }
+        val expected = workerCpu.get()
+        check(expected > 0L) { "The accounting check worker recorded no CPU time" }
+        check(counters.javaThreadCpuNanos >= expected * CPU_ACCOUNTING_MIN_SHARE_PERCENT / PERCENT) {
+            "Request-serving CPU accounting missed a worker that lived inside the window: " +
+                "worker ${expected} ns, accounted ${counters.javaThreadCpuNanos} ns, " +
+                "process ${counters.processCpuNanos} ns, internal ${counters.jvmInternalCpuNanos} ns"
+        }
+        verifySnapshotNesting("idle action") { 0L }
+        verifySnapshotNesting("collector-loaded action") {
```
johnsonlee: [P1] Do not perturb every benchmark fork with a 512 MiB allocation contract

This runs after `quiesceBeforeMeasurement()` and before the measured invocation, and it allocates/touches 512 × 1 MiB while retaining 64 MiB. That changes committed RSS, heap occupancy, collector state, and possibly leaves a concurrent cycle started after the quiescence proof; it therefore changes the latency/CPU/RSS workload the gate is supposed to observe, and the added RSS denominator can reduce sensitivity to candidate memory regressions. A measurement-contract stress test must run in a separate smoke/test JVM (or otherwise outside the real benchmark trial), not mutate every real 4/17/36-graph fork immediately before measurement.

johnsonlee: Agreed: the contract allocated 512 MiB and left collector state between the quiescence proof and the measured invocation, changing exactly what the gate observes. Fixed in 84ac083.

- The accounting moved into `RequestCpuAccounting`, and its contract is now a standalone `MethodCompatibilityCpuAccountingContract` main. The benchmark workflow runs it in a JVM of its own, before any measured fork, so the short-lived-worker check and both snapshot-nesting scenarios (idle and the collector-loaded 512 MiB / 64 MiB one) never touch a benchmark trial's heap or collector.
- Trial setup keeps only `RequestCpuAccounting.requireAvailable()`, which reads `/proc/self/task` without allocating or starting a thread and fails closed if the per-thread accounting is missing.
- `measure()` delegates to `RequestCpuAccounting.measure`, so the real forks do only the nested process/internal reads around the request.
- A gate test asserts the workflow runs the contract in its own JVM before the first `methodScenarioGate` fork. Compile, detekt and the 115 gate tests pass.

---

graphite-explore/src/jmh/kotlin/io/johnsonlee/graphite/cli/RequestCpuAccounting.kt
```diff
@@ -0,0 +1,196 @@
...
+     */
+    private fun internalThreadCpuNanos(): Map<Long, Long> {
+        val tasks = Path.of("/proc/self/task")
+        check(Files.isDirectory(tasks)) { "Per-thread CPU accounting is unavailable: $tasks is missing" }
+        val result = HashMap<Long, Long>()
+        Files.newDirectoryStream(tasks).use { entries ->
+            for (task in entries) {
+                val tid = task.fileName.toString().toLongOrNull() ?: continue
+                val name = runCatching { Files.readString(task.resolve("comm")).trim() }.getOrNull() ?: continue
+                if (!isJvmInternalThread(name)) continue
```
johnsonlee: [P1] Do not infer JVM-internal ownership from a mutable native thread name

`/proc/.../comm` is not an ownership signal: an application Java worker can use the same name prefix, and Linux truncates it to 15 bytes. I reproduced this on exact head `6355f50a` with Temurin 17/Linux: a persistent request worker named `Service Thread fake` appeared as native name `Service Thread`; for a 499,200,749 ns worker the sampler reported process=520,000,000 ns, internal=501,780,584 ns, and request-serving Java CPU=18,219,416 ns. About 96% of real request CPU was subtracted, while the current contract still passes because its short-lived worker uses a default name and is absent from both snapshots. A candidate that introduces an innocently colliding worker name can therefore hide a CPU regression and pass the future blocking gate. Use an ownership source that cannot collide with Java/application threads; if the JVM does not expose one, this derived metric cannot safely become blocking. Add the persistent colliding-name worker as a negative contract.

johnsonlee: Agreed, and the reproduction is decisive: `/proc/<tid>/comm` is a mutable, 15-byte-truncated label, so identifying the *internal* set by native name is unsound — an application worker named `Service Thread fake` collides with `Service Thread` and 96% of real request CPU is subtracted. Name-based ownership cannot back a blocking gate.

Root cause: the metric identifies the JVM-internal set (to subtract it) by name. The safe fix is to stop inferring ownership at all and measure request-serving CPU **from inside the worker threads**, where a thread reads *its own* `ThreadMXBean.getThreadCpuTime()` — no name, no `/proc` internal set, no subtraction, so a colliding name is irrelevant. That also finally closes the born-and-died-inside-the-window gap from r3946647810 (the wrapper runs on the transient worker itself), which is why the two-snapshot sum couldn't and process−internal had to.

Two directions; both need your call because A reverses b50067a and B changes the metric-switch plan:

- **A — robust, blocking-capable.** Route every request-serving Java thread (client, server workers, scan pools) through a task wrapper that accumulates its own `getThreadCpuTime()` delta into the per-request total in `try/finally`. Ownership is self-measurement, collision-proof by construction; transient workers are captured. Cost: the explore server + scan-pool submission paths must funnel through the instrumented wrapper — reachable since the harness constructs the server, but it is a real change to the request path, not just `RequestCpuAccounting`. The persistent colliding-name worker becomes a negative contract: it must not move the number.
- **B — concede the derived metric.** Keep `processCpuNanos` as the blocking figure under a robust repeated-pair protocol (your option (b) on r3946647810), and keep `javaThreadCpuNanos` advisory only. The metric-switch PR then would not promote it to the gate.

My recommendation is **A**: it is the only version that makes the per-request Java-CPU figure trustworthy enough to block on, and it subsumes both prior P1s. I'll add the persistent colliding-name worker as a negative contract under either choice. Tell me which you want and I'll implement it on this head; I won't push the redesign until you pick, since it's not a local change.

---

johnsonlee: Implemented **A** by Java-thread identity in `a3568f4`. `javaThreadCpuNanos` is now the sum of each Java thread's own `ThreadMXBean.getThreadCpuTime` over the window, keyed by thread id. JVM-internal native threads (collectors, compilers, VM/service) have no `Thread` object and no id, so they are never in the sum — no name matching, no `/proc`, and a worker named `Service Thread` is charged, not subtracted.

One refinement from what I described, forced by the request path: I traced every request-serving pool — `callSiteScanExecutor` and `splitCallSiteExecutor` (`MappedWebGraphBackedGraph`/`MappedCallSiteStringIndex`), `directStringExecutor` (`QueryPipeline`), the `methodGraphScanPool` `ForkJoinPool` (`MethodQueryExecutor`), and `CypherQueryGuard`'s executors. **None accepts an injected `ThreadFactory`/`Executor`**, and all are persistent daemon pools created lazily then reused. So there is no seam to wrap without production hooks across webgraph/cypher/explore, and the real path has no born-and-died-inside-the-window worker to capture.

So instead of wrapping to *capture* a transient worker, the accounting *fails closed* on one: a pre-existing thread gone at the end, or (via `getTotalStartedThreadCount`) more thread starts than newly-present threads, throws rather than under-reports. A candidate that introduces a per-request short-lived worker can't look cheaper — the gate errors. That keeps your "can't hide a regression" guarantee with zero production change; true capture stays a follow-up if you'd rather have the number than the tripwire.

Negative contract added and passing: a persistent worker named `Service Thread` is charged in full — the contract prints `worker 200,267,104 ns, accounted 201,195,741 ns` where the old name-based code hid ~96% of it — and a transient worker makes `measure()` fail closed. `processCpuNanos` stays advisory.

Validated the tripwire does not fire on the real path: 40 served requests through Javalin/Jetty + `HttpURLConnection`, and 40 `ForkJoinPool` + `ExecutorCompletionService` fan-outs (the engine's patterns), each showed zero transient Java threads. Contract passes deterministically across runs.

If you want true capture instead of fail-closed for transient workers, say so and I'll add the production accounting hook; otherwise this is the collision-proof, no-production-change form.

---

graphite-explore/src/jmh/kotlin/io/johnsonlee/graphite/cli/RequestCpuAccounting.kt
```diff
@@ -0,0 +1,233 @@
...
+        // The process figure is read innermost so its interval is contained in the thread
+        // snapshots; the request-serving Java sum can then exceed it only by snapshot overhead.
+        val beforeProcess = processCpuTimeNanos()
+        val result = action()
+        val afterProcess = processCpuTimeNanos()
+        val afterIds = threads.allThreadIds
+        val afterCpu = threadCpuSnapshot(threads, afterIds)
+        val afterStarted = threads.totalStartedThreadCount
+
+        val beforeIdSet = beforeIds.toHashSet()
```
johnsonlee: [P1] Base liveness checks on successful CPU reads, not the earlier ID list

There is still a fail-open race between `allThreadIds` and `threadCpuSnapshot`. A worker can be present in `afterIds`, then terminate before its id reaches `getThreadCpuTime`; that read returns `-1` and line 102 deliberately omits it from `afterCpu`. Lines 64–72 nevertheless use the earlier raw ID lists: the worker counts as present/appeared, the started-count delta is balanced, and neither check throws. Its entire in-window CPU is then absent from the sum at lines 78–82. The symmetric race exists in the before snapshot as well. Thus the stated contract (“a thread gone at the end fails closed”) is not guaranteed, and a candidate can still hide CPU in a worker that exits during the end snapshot. Derive liveness/appeared sets from the ids whose CPU read actually succeeded (or use an atomic JVM snapshot facility), and add a contract that forces a worker to exit between end-ID enumeration and its CPU read.

johnsonlee: Correct — that was a real fail-open race. Fixed in `3af8b17`.

The liveness, appeared, and started-count checks now derive from the ids whose `getThreadCpuTime` **succeeded** — the snapshot map keys (`beforeCpu.keys` / `afterCpu.keys`) — not the raw `allThreadIds`. So a worker enumerated in the end id list that terminates before its read returns -1 is absent from `afterCpu`, and is therefore counted as vanished (or, if it was born inside the window, is missing from `appeared`, so the `getTotalStartedThreadCount` delta exceeds `appeared`). Either way the window fails closed instead of summing it as present-with-no-CPU. The symmetric before-snapshot case is handled the same way: a thread whose start read fails is not in `beforeCpu.keys`, so it is simply not tracked — it was already terminating at the window start with ~0 in-window CPU, and thread ids are not reused, so it cannot reappear at the end.

Deterministic contract added, as you asked: a seam forces one id's end read to be treated as failed (standing in for "exited between end enumeration and its read"), and a persistent worker under it must make `measure()` fail closed. The contract now prints `end-snapshot read race failed closed = true`, alongside the colliding-name worker charged in full (`201,931,586 ns` for a `200,307,968 ns` worker) and the transient-worker fail-closed. Compile, detekt, and the contract pass deterministically across runs.

---

.github/workflows/benchmark.yml
```diff
@@ -608,6 +608,9 @@ jobs:
           -Dkotlin.compiler.graph.path=${PWD}/fixture-graphs/kotlin-compiler"
         METHOD_ARGS=(-wi 0 -i 1 -f 1 -t 1 -foe true -rf json \
           -p "graphCount=${GRAPH_COUNT}" -p "scenario=${SCENARIOS}")
+        # The request-serving CPU accounting contract runs in a JVM of its own, on this runner and
+        # JVM, before any measured fork; it must never run inside a benchmark trial.
+        java -cp "${CANDIDATE_JAR}" io.johnsonlee.graphite.cli.MethodCompatibilityCpuAccountingContract
```
johnsonlee, line 613: [P1] Exercise the new accounting in the benchmark forks, not only in a standalone contract

This contract validates the new class, but the actual CI benchmark never uses it. `build-explore-jmh` still installs `ExplorerMemoryBenchmark.kt` from the base checkout (`gate`) into **both** source trees at lines 90–98. On this PR the base-owned file is the old process-only harness, so both generated JMH jars execute the old `measure()` and publish no `javaThreadCpuNanos`; the validation at lines 631+ likewise requires and gates only `processCpuNanos`. Thus a green contract only proves a dormant helper in the candidate jar, while none of the real 4/17/36-graph requests exercise its integration, quiescence, counters, or fail-closed behavior. Add an independent candidate-harness smoke/meta-validation path (or install the reviewed new harness identically into base and candidate for a separate paired run) and assert the resulting JMH JSON contains valid `javaThreadCpuNanos` rows before this measurement can be considered established.

johnsonlee: Correct — on this PR the new accounting is dormant: `build-explore-jmh` installs the base-owned `gate/ExplorerMemoryBenchmark.kt` (main's process-only harness) over both trees, so both jars run the old `measure()`, and only `RequestCpuAccounting.kt` (not in the install list) is new but uncalled. The standalone contract proves the class, not its integration in a real fork.

Two ways to close it; the second bears on the landing order you designed, so I want your call before I touch the workflow:

- **A — candidate-harness smoke (self-contained in #119).** In `build-explore-jmh`, for the candidate revision, build a second jar from the candidate's own (un-gate-installed) `ExplorerMemoryBenchmark.kt` and upload it separately. A small job (reusing the method shard's corpora setup) runs one `methodScenarioGate` fork (`graphCount=4`, one scenario, `-wi 0 -i 1 -f 1`) with that jar and asserts the JMH JSON row has `secondaryMetrics.javaThreadCpuNanos.score > 0` and `requestsSucceeded == 1` — proving integration, quiescence, counters, and a real row. It's a candidate-only smoke: it does not gate on the number and does not touch the trusted `processCpuNanos` gate on the base-installed harness.
- **B — paired run with the new harness installed on both sides.** This is essentially what the metric-switch PR already is ("the main-vs-main validation of this harness, base-installed control on both sides"). Folding it here would pull that validation forward and change the landing order.

My recommendation is **A**: it establishes the harness within this PR without disturbing the trusted control or the landing order, and leaves B as the metric-switch PR. Caveat: the smoke is corpora-dependent, so I'll validate the workflow wiring and add a gate test for it locally, but the fork's `javaThreadCpuNanos` row is proven in CI (the four method corpora aren't available locally). Confirm A (or say you'd rather do B) and I'll implement it on this head.

---

johnsonlee: Choose **A**. A candidate-owned smoke is the right proof boundary for #119: it exercises the new harness integration without letting the PR replace the trusted paired gate that evaluates it. Use `graphCount=4` so the single fork actually includes all four real corpora; assert the exact benchmark/scenario, `requestsSucceeded == 1`, `javaThreadCpuNanos.score > 0`, and that the accounting diagnostics are present/nonnegative. Keep the smoke artifact/job distinct from the base-owned paired artifacts so it cannot be consumed accidentally by the existing gate. The later metric-switch PR can perform B and compare the same reviewed harness on both production revisions.

johnsonlee: Implemented A in `59b3983`.

New `validate-cpu-accounting` job (needs `prepare-latency-fixtures`): it builds the candidate's **own** Explorer JMH jar with no base-owned gate install — so this jar actually runs the new `RequestCpuAccounting` integration, unlike the paired gate — restores the four-corpus fixture cache, runs the contract on that jar, then runs one real `methodScenarioGate` `graphCount=4` fork (`-wi 0 -i 1 -f 1 -t 1`, scenario `count`) over all four corpora and asserts the JMH JSON row:

- exactly one row, `benchmark == …methodScenarioGate`, `graphCount == "4"`, `scenario == "count"`, `mode == "ss"`, `primaryMetric.score > 0`;
- `requestsSucceeded.score == 1`, `javaThreadCpuNanos.score > 0`;
- accounting diagnostics present and nonnegative: `processCpuNanos > 0`, and `jvmInternalCpuNanos`, `cpuAccountingUnderflowNanos`, `jvmInternalThreadsEnded`, `gcTimeMillis`, `jitTimeMillis` all `>= 0`.

It is candidate-only and uploads a distinct `benchmark-cpu-accounting-smoke-…` artifact — it does not feed or replace the trusted paired gate on `processCpuNanos`, and it has no base counterpart. The metric-switch PR still performs B (the reviewed harness installed on both revisions).

A gate test asserts the smoke builds the candidate harness (and does **not** install the base one), runs the `graphCount=4` fork, asserts the `javaThreadCpuNanos` row, and keeps its artifact distinct from the paired jars.

Validated locally: the workflow parses, the `jq` assertion accepts a valid row and rejects `javaThreadCpuNanos == 0`, `@AuxCounters` emits the row (as the existing `processCpuNanos` assertion already relies on), and all 89 node gate tests pass. The four method corpora aren't available locally, so the fork's real end-to-end run (quiescence + `measure()` integration producing the row) is proven in CI — which is the point of the smoke.

---

.github/workflows/benchmark.yml
```diff
@@ -754,6 +757,107 @@ jobs:
         if-no-files-found: error
         retention-days: 14
 
+  validate-cpu-accounting:
```
johnsonlee, line 760: [P1] Make the smoke a dependency of the authoritative aggregate gate

The new job proves the integration and its exact-head run passed, but `benchmark-regression-gate.needs` does not include `validate-cpu-accounting`, and the enforce step does not check its result. A future failure of this smoke can therefore coexist with a successful authoritative aggregate check; if branch protection requires only `benchmark-regression-gate` (the purpose of that aggregate), the broken accounting can still merge. Add this job to the aggregate `needs`, expose `${{ needs.validate-cpu-accounting.result }}`, fail unless it is `success`, and extend the workflow test to assert that dependency/enforcement—not merely that a standalone job exists.

johnsonlee: Done in `90dc4e5`.

- `validate-cpu-accounting` is now in `benchmark-regression-gate.needs`.
- The Enforce step exposes `CPU_ACCOUNTING_SMOKE_JOB: ${{ needs.validate-cpu-accounting.result }}` and fails unless it is `success` — so a smoke `failure`, `skipped`, or `cancelled` now fails the authoritative aggregate. Branch protection on `benchmark-regression-gate` therefore covers the accounting: broken accounting can no longer merge behind a green aggregate.
- The gate test is extended to assert the dependency and the enforcement (the aggregate's `needs` includes `validate-cpu-accounting`, and the Enforce step exposes and checks its result), not merely that the standalone job exists.

Locally: the workflow parses and all 90 node gate tests pass.

---

johnsonlee:
`cypher-capacity-gate` flaked on `90dc4e5`, not a regression from this PR.

The failing step is the parity gate: candidate must stay within 15% of base on `processCpuNanos`, `residentSetAfterBytes`, and `tailLatencyNanos`; the comparator exited 1 because one metric crossed the 15% line. This is a harness-only PR — it changes only the explore method-compat benchmark harness and the workflow — so base and candidate run **identical cypher production code**; the crossing is hosted-runner variance (the same 2-of-6 flake seen earlier on this gate), not a candidate change. It passed on the previous head `59b3983`, and `90dc4e5`'s diff is limited to `benchmark.yml` (wiring `validate-cpu-accounting` into the aggregate `needs`/enforce) and a node gate test, neither of which touches the cypher-capacity job.

I can't re-run the job from here (`rerun_failed_jobs` returns 403 for this integration); a re-run from the Actions page would settle it. Leaving it as a flake to re-run, with nothing to fix in this PR.

---

PR #120, opened 2026-09-08
Cover the direct-string worker cancel-before-start branch deterministically

johnsonlee:
## Summary

The cypher module's line coverage sits exactly on the 98% gate and flips between **97.9975%** (fails `run-unit-test`) and **98.0132%** (passes) run to run. The two lines that flip are the direct-string worker's cancel-before-start branch, inlined at three sites in `QueryPipeline.kt`:

```kotlin
if (future.cancel(true) && started.compareAndSet(false, true)) completion.countDown()
```

That not-started branch is reached only when a query stops early or fails while a pooled worker future is still queued, i.e. only when cancellation wins the race against the worker starting. Normal single-query runs size the pool to the worker count, so the branch is hit or missed by timing, and the module's coverage lands on the wrong side of the threshold at random. It has intermittently failed `run-unit-test` on unrelated PRs (#117, #119).

## Change

- Extract the branch into `internal fun cancelDirectStringWorkerBeforeStart(future, started, completion)`; the three sites (`runDirectStringTasksInOrderUntil`, the fixed-worker variant's `cancelAndJoin`, and its `stopWorkers`) now call it. No behavior change: a not-started future is cancelled and its latch released, a started worker keeps running to its own `finally`, a future that can no longer be cancelled is left alone.
- `DirectStringWorkerCancellationTest` drives the helper's three outcomes directly with `FutureTask` and `AtomicBoolean`, no threads or pool, so the branch is covered every run.

## Validation

- `:cypher:compileKotlin`, `:cypher:compileTestKotlin`, `:cypher:detekt` pass.
- Full `:cypher` kover after the change: **98.0142%** (covered 6219 / missed 126), with the helper's lines covered every run.

This unblocks `run-unit-test` for any PR whose cypher coverage was riding the threshold.

---

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ gradle-app.setting
 
 # # Work around https://youtrack.jetbrains.com/issue/IDEA-116898
 # gradle/wrapper/gradle-wrapper.properties
+.kotlin/
--- a/graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
+++ b/graphite-cypher/src/main/kotlin/io/johnsonlee/graphite/cypher/QueryPipeline.kt
@@ -153,6 +153,25 @@ private val configuredDirectStringParallelism by lazy { System.getProperty(DIREC
 private val directStringExecutorParallelism: Int by lazy {
     resolveDirectStringExecutorParallelism(configuredGraphWorkers = configuredDirectStringParallelism)
 }
+/**
+ * Cancels a direct-string worker that has not started yet and releases its completion latch, so a
+ * query that stops early or fails does not wait on a worker the executor never ran. A worker that
+ * has already started (its [started] flag is set) keeps running to its own `finally`, and a future
+ * that can no longer be cancelled is left alone. Extracted so the not-started branch is covered
+ * deterministically rather than only when cancellation happens to win the race against worker start.
+ */
+internal fun cancelDirectStringWorkerBeforeStart(
+    future: Future<*>,
+    started: AtomicBoolean,
+    completion: CountDownLatch
+): Boolean {
+    if (future.cancel(true) && started.compareAndSet(false, true)) {
+        completion.countDown()
+        return true
+    }
+    return false
+}
+
 private val directStringExecutor by lazy {
     Executors.newFixedThreadPool(directStringExecutorParallelism) { runnable ->
         Thread(runnable, "graphite-cypher-scan-${directStringWorkerNumber.incrementAndGet()}").apply {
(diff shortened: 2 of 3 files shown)


johnsonlee:
The `method-compatibility-4-aggregate` check (and the sibling method-compatibility shards) are failing on `processCpuNanos` variance, not on anything in this PR.

- This PR is cypher-only: it extracts `cancelDirectStringWorkerBeforeStart` in `QueryPipeline.kt` and adds a deterministic cypher test. It touches neither `ExplorerMemoryBenchmark` nor the benchmark harness.
- The `methodScenarioGate` benchmark itself ran and saved results; the **comparator** exited 1 on run-to-run process-CPU noise (this run: candidate `processCpuNanos` count 730M / order 710M vs base 960M / 820M). Wall time was steady (~347–397 ms/op), only the process-CPU metric diverged.
- This is the known by-design red on `main` (the base of this PR): the method-compat gate measures process-CPU, whose stdev on identical code is ~22% vs ~6% for wall time. It reproduces identically on every head until the metric switch lands.

The fix (switch the gate to per-request `javaThreadCpuNanos` with `processCpuNanos` kept advisory) is staged but not yet opened as a PR — it is sequenced behind #119 (which introduces the accounting) merging. No re-run needed; it reproduces deterministically on unrelated code. Leaving this red as expected and not widening this PR.

---

johnsonlee:
Consolidated note on the benchmark suite for this PR (covers `global-wide-pressure-evidence`, the `benchmark-regression-gate` rollup, and the method-compatibility shards already discussed above).

This PR is cypher-only — it extracts `cancelDirectStringWorkerBeforeStart` in `QueryPipeline.kt` and adds a deterministic cypher test. It changes nothing in `graphite-webgraph`, `graphite-explore`, or the CallSite query engine, so:

- **global-wide-pressure-evidence**: the gate requires a large P95 speedup (5x on `main`'s comparator) of candidate vs base. Base and candidate run identical global-wide production code here, so the speedup is ~1x and the gate fails by design. No fix belongs in this PR.
- **method-compatibility shards / method-compatibility-gate**: process-CPU variance on identical explore benchmark code, as detailed in my earlier comment — by-design until the metric switch lands.
- **benchmark-regression-gate**: the rollup of the above; it is red purely because those by-design components are red.

The check this PR actually targets — coverage — is green: all six modules pass, cypher back to a stable 98.0% now that the cancel-before-start branch is deterministically covered. No re-runs will change the benchmark rows (identical code), so I am leaving the benchmark suite red as expected and not widening this PR. Future benchmark-suite failures on this head are covered by this note.

---

PR #121, opened 2026-09-09
Gate method-compatibility CPU on javaThreadCpuNanos, keep process CPU advisory

johnsonlee:
## Summary

The metric switch that follows the harness PR (#119). #119 landed the request-serving **Java-thread CPU** accounting (`javaThreadCpuNanos`, measured by Java-thread identity) as main's base-installed `MethodDiscoveryCompatibilityBenchmark`. Because the benchmark workflow installs the base-owned harness over both trees, both sides of the `method-compatibility` gate now emit `javaThreadCpuNanos`. This PR flips the blocking CPU gate to it.

## What changes

In the `method-compatibility-shard` gate step:

- **`cpu` gate → `javaThreadCpuNanos` (blocking).** The request row's CPU is now the sum of each Java thread's own `ThreadMXBean.getThreadCpuTime` over the window — the JVM's internal collector/compiler/VM threads are never in it.
- **`processCpuNanos` → advisory (`process-cpu`, non-blocking).** Whole-process CPU is still reported so a swing can be attributed (against `gcTimeMillis`/`jitTimeMillis`), but it no longer fails the gate.
- The advisory `process-cpu` report and status join the shard report/status aggregation alongside the existing metrics.

No production code changes; this is a workflow gate change only.

## Why

The shards gated each request row on `processCpuNanos`, which folds the concurrent collector and compiler threads into the row. On the hosted runner that figure swings between runs of **identical code** — over 22 rows of recent heads the process-CPU deltas ranged −32%..+57% (stdev 22%) while the wall rows stayed within −9%..+13% (stdev 6%), and every red shard was a process-CPU row with all wall rows passing. Gating on the request-serving Java-thread CPU removes that collector/JIT noise from the blocking decision while keeping the process figure visible.

This run is the main-vs-main validation of the #119 harness: the reviewed harness is the base-installed control on both revisions, so any movement in `javaThreadCpuNanos` here is the metric's own run-to-run behaviour on identical code, not a candidate diff.

## Validation

- `.github/scripts/benchmark-gate.test.mjs`: 91 node tests pass, including a new test asserting the shard gates `cpu` on `javaThreadCpuNanos` (blocking), keeps `processCpuNanos` advisory, and aggregates the advisory process-CPU report and status.
- `benchmark.yml` parses as valid YAML.

---

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -670,7 +670,8 @@ jobs:
 
         NEED_CONFIRMATION=0
         gate_metric wall '' 'wall time' false || NEED_CONFIRMATION=1
-        gate_metric cpu processCpuNanos 'CPU time' false || NEED_CONFIRMATION=1
+        gate_metric cpu javaThreadCpuNanos 'CPU time (Java threads)' false || NEED_CONFIRMATION=1
+        gate_metric process-cpu processCpuNanos 'process CPU time (advisory)' true || NEED_CONFIRMATION=1
         gate_metric rss-after residentSetAfterBytes 'RSS after query' false || NEED_CONFIRMATION=1
         gate_metric rss-delta residentSetDeltaBytes 'RSS query delta (advisory)' true || NEED_CONFIRMATION=1
 
@@ -704,7 +705,8 @@ jobs:
         FINAL_FAILURE=0
         for SPEC in \
           'wall::wall time:false' \
-          'cpu:processCpuNanos:CPU time:false' \
+          'cpu:javaThreadCpuNanos:CPU time (Java threads):false' \
+          'process-cpu:processCpuNanos:process CPU time (advisory):true' \
           'rss-after:residentSetAfterBytes:RSS after query:false' \
           'rss-delta:residentSetDeltaBytes:RSS query delta (advisory):true'; do
           IFS=: read -r SLUG METRIC TITLE ADVISORY <<< "${SPEC}"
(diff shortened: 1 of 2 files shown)


.github/workflows/benchmark.yml
```diff
@@ -670,7 +670,8 @@ jobs:
 
         NEED_CONFIRMATION=0
```
johnsonlee, line 671: [P1] This switch makes a known failing measurement path authoritative. On this exact head, the base fork aborts before comparison in `method-compatibility-17-string`, `36-string`, and `36-scan`: `RequestCpuAccounting.measure` reports 1–2 Java threads present at the start but gone at the end. The failures occur on real 17/36-graph scenarios (for example `prefix` and `contains`), so the single 4-graph/count smoke from #119 did not establish that this counter is usable across the gate matrix. This is not a process-CPU fluctuation and rerunning cannot validate the proposed gate while the measurement itself aborts. Fix the accounting/lifecycle issue (or land a prerequisite harness repair), add coverage for the affected long-running scenarios, and show the full method matrix completes before promoting `javaThreadCpuNanos` to blocking.

johnsonlee: Agreed on all points — this switch cannot land while the measurement aborts, and the 4-graph/count smoke was too short a window to surface it.

Root cause: the method-compatibility benchmark issues its requests through an embedded Javalin/Jetty server, and Jetty's default `QueuedThreadPool` reaps idle worker threads on their idle timeout (and idle reserved threads via `ReservedThreadExecutor`). On the longer 17/36-graph scenarios (`prefix`, `contains`) that reaping lands inside a measured window, so a Java thread present at the start is gone at the end and `measure` fails closed — on identical code. It stayed hidden on #119 because the workflow installed the old process-only base harness on both sides; now that the identity harness is the base-installed control, the tripwire reaches the real 17/36 path (#117 hits it too).

Landed the prerequisite harness repair as #122, not weakening the tripwire:
- The benchmark server's Jetty pool is pinned — no idle timeout, no reserved threads, a fixed floor — so its threads stay present for the life of the trial. A genuine request worker that vanishes still fails closed.
- Two deterministic contract checks: `checkPresentThenGoneFailsClosed` (a thread present at the window start that exits inside it must fail closed — the exact branch that fired, previously uncovered) and `checkPinnedServerPoolHoldsThreadsAcrossWindow`. Both pass locally.
- `validate-cpu-accounting` now also runs `graphCount=36` over `scenario=prefix,contains` on the candidate (pinned) harness with `-foe true`, so an abort fails the job — the full-matrix proof on the candidate harness (the shards install the base harness, so they can't show the fix until #122 lands).

Since the method-compatibility shards install the base-owned harness, #122 must land on `main` before the shards stop aborting; once it does I'll rebase this PR and re-run so the full matrix completes before `javaThreadCpuNanos` is promoted to blocking. Holding this PR until then.

---

PR #122, opened 2026-09-09
Disable client keep-alive so the method-compat accounting window has no vanishing Keep-Alive-Timer

johnsonlee:
## Summary

Prerequisite harness repair for the `javaThreadCpuNanos` metric switch (#121), addressing the P1 there: on the longer 17/36-graph method scenarios the base fork aborts before comparison because `RequestCpuAccounting.measure` reports 1–2 Java threads present at the window start but gone at the end.

### Root cause (corrected)

The vanished thread is **not** a Jetty worker. Javalin 6.4.0's default `QueuedThreadPool` uses a 60s idle timeout and the server starts immediately before measurement, so its workers cannot reach that timeout within the 11–19s window where #121 failed. The actual thread is the JDK `HttpURLConnection` client's **`Keep-Alive-Timer`**, which the benchmark's `rawRequest()` spawns during setup; it is present at the window start and exits when the keep-alive cache idles, landing inside the longer windows.

Reproduced locally on JDK 17 (isolated, no project deps): after one `HttpURLConnection` request + `disconnect()`, `Keep-Alive-Timer` is in the start snapshot and vanishes within ~15s; with `-Dhttp.keepAlive=false` it is never created and nothing vanishes — exactly the liveness pattern and timing of #121.

This stayed hidden on #119 because the workflow installed the old process-only base harness on both sides; now that the identity harness is the base-installed control on `main`, the tripwire reaches the real 17/36 path (both #121 and #117 hit it).

### Fix

- **Fix the actual lifecycle**: the method-compatibility fork runs with `-Dhttp.keepAlive=false`, so the client never spawns `Keep-Alive-Timer`. What's measured is the server's request-serving CPU; disabling client connection reuse is symmetric across base and candidate. The tripwire is unchanged — a genuine request worker that vanishes still fails closed.
- **Name the offender**: `measure()` now reports each vanished thread by `name#id`, captured at the start snapshot (names are only reported, never used for an accounting decision), so a future non-request lifecycle is identified rather than guessed.

### Coverage

- `MethodCompatibilityCpuAccountingContract.checkPresentThenGoneFailsClosed` (deterministic) asserts a thread present at the window start that exits inside it fails closed **and** that the failure names the thread. Passes locally.
- `validate-cpu-accounting` runs `graphCount=36` over `scenario=prefix,contains` on the candidate harness (which runs with `http.keepAlive=false`) with `-foe true`, so an abort fails the job. It passes because the cause is addressed, and would still abort without the flag.

### Landing order

The method-compatibility shards install the base-owned harness, so this repair must land on `main` before the shards stop aborting and before #121 promotes `javaThreadCpuNanos` to a blocking gate. Expect this PR's own `method-compatibility-*` shards to still abort (they run the pre-merge base harness); the candidate-harness `validate-cpu-accounting` job is the proof for this PR.

### Validation

- `:explore:compileJmhKotlin` passes for the candidate, and with `origin/main`'s harness installed over the candidate `RequestCpuAccounting` (the `build-explore-jmh` scenario).
- The contract passes locally (`present-then-gone failed closed = true, named = true`).
- 90 node gate tests pass; `benchmark.yml` parses as valid YAML.

(The branch name still says "pin-jetty-pool" from the first attempt; left as-is to avoid churning the PR.)

### DIFF
--- a/graphite-explore/src/jmh/kotlin/io/johnsonlee/graphite/cli/ExplorerMemoryBenchmark.kt
+++ b/graphite-explore/src/jmh/kotlin/io/johnsonlee/graphite/cli/ExplorerMemoryBenchmark.kt
@@ -12,6 +12,7 @@ import io.johnsonlee.graphite.graph.MethodPattern
 import io.johnsonlee.graphite.input.LoaderConfig
 import io.johnsonlee.graphite.sootup.JavaProjectLoader
 import io.johnsonlee.graphite.webgraph.GraphStore
+import org.eclipse.jetty.util.thread.QueuedThreadPool
 import org.openjdk.jmh.annotations.AuxCounters
 import org.openjdk.jmh.annotations.Benchmark
 import org.openjdk.jmh.annotations.BenchmarkMode
@@ -366,6 +367,21 @@ open class ExplorerMemoryBenchmark {
 }
 
 /**
+ * A Jetty pool that keeps every server thread for the life of the trial. Jetty's default
+ * [QueuedThreadPool] reaps idle worker threads on their idle timeout and idle reserved threads via
+ * its [org.eclipse.jetty.util.thread.ReservedThreadExecutor]; on the longer 17/36-graph method
+ * scenarios that reaping lands inside a measured window, and [RequestCpuAccounting] then fails
+ * closed because a Java thread present at the window start is gone at the end. Disabling the idle
+ * timeout (and reserved threads) removes that benign churn without weakening the tripwire: a genuine
+ * request worker that vanishes still trips it.
+ */
+internal fun pinnedServerThreadPool(): QueuedThreadPool =
+    QueuedThreadPool(SERVER_MAX_THREADS, SERVER_MIN_THREADS, NO_IDLE_TIMEOUT_MILLIS).apply {
+        name = "graphite-method-bench-jetty"
+        reservedThreads = 0
+    }
+
+/**
  * Paired migration gate for the method-discovery HTTP surface.
  *
  * The candidate copy of this source is compiled into both revisions. A base server is therefore
@@ -424,6 +440,13 @@ open class MethodDiscoveryCompatibilityBenchmark {
         topology = TopologyService(registry, emptyList(), root).also { it.rebuild() }
         app = Javalin.create { config ->
             config.jsonMapper(JavalinGson(GsonBuilder().create()))
+            // Pin the embedded Jetty pool so no idle server thread is reaped inside a measured
+            // window. RequestCpuAccounting fails closed when a Java thread present at the window
+            // start is gone at the end; Jetty's default QueuedThreadPool reaps idle worker and
+            // reserved threads on their idle timeout, which on the longer 17/36-graph scenarios
+            // lands inside the window and trips the tripwire on identical code. See
+            // pinnedServerThreadPool.
+            config.jetty.threadPool = pinnedServerThreadPool()
         }.start(0)
         cypherGuard = CypherQueryGuard(TARGET_CYPHER_CONCURRENCY, TARGET_CYPHER_WORK_BUDGET)
         ExploreRoutes(cypherGuard).register(app, registry, topology)
@@ -1681,6 +1704,12 @@ private fun forceTopologyHeapGc() {
         Thread.sleep(TOPOLOGY_HEAP_GC_PAUSE_MS)
     }
 }
+
+// A very large idle timeout keeps the benchmark server's Jetty threads from being reaped inside a
+// measured window; Int.MAX_VALUE ms is ~24 days, far beyond any trial.
+private const val NO_IDLE_TIMEOUT_MILLIS = Int.MAX_VALUE
+private const val SERVER_MIN_THREADS = 8
+private const val SERVER_MAX_THREADS = 64
 
 private const val TOPOLOGY_HEAP_GRAPH_COUNT = 3
 private const val TOPOLOGY_HEAP_GC_ATTEMPTS = 3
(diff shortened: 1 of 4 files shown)


graphite-explore/src/jmh/kotlin/io/johnsonlee/graphite/cli/ExplorerMemoryBenchmark.kt
```diff
@@ -424,6 +424,13 @@ open class MethodDiscoveryCompatibilityBenchmark {
         topology = TopologyService(registry, emptyList(), root).also { it.rebuild() }
         app = Javalin.create { config ->
             config.jsonMapper(JavalinGson(GsonBuilder().create()))
+            // Pin the embedded Jetty pool so no idle server thread is reaped inside a measured
+            // window. RequestCpuAccounting fails closed when a Java thread present at the window
+            // start is gone at the end; Jetty's default QueuedThreadPool reaps idle worker and
+            // reserved threads on their idle timeout, which on the longer 17/36-graph scenarios
+            // lands inside the window and trips the tripwire on identical code. See
+            // pinnedServerThreadPool.
+            config.jetty.threadPool = pinnedServerThreadPool()
```
johnsonlee: [P1] The evidence does not support Jetty reaping as the vanished thread, and this change likely pins the wrong pool. Javalin 6.4.0 creates its default `QueuedThreadPool` with a 60,000 ms idle timeout; the exact #121 failures happen after roughly 11–19 seconds, and this server is started only immediately before the measurement setup, so its workers have not reached that timeout. In contrast, this same benchmark calls `rawRequest()` through `HttpURLConnection` during setup and for every query. I reproduced the JDK 17 behavior independently: after one local `HttpURLConnection` request plus `disconnect()`, `Keep-Alive-Timer` is present in the start snapshot and vanishes within a 12-second window—exactly the liveness pattern and timing seen in #121. Pinning Jetty cannot keep that client-side Java thread alive. Please identify the vanished thread by id/name from the before snapshot, then fix or exclude the actual non-request lifecycle; the 36-graph smoke must pass because the cause is addressed, not coincidentally.

johnsonlee: You're right, and I reproduced it. The Jetty pin was the wrong pool: Javalin 6.4.0's default `QueuedThreadPool` has a 60s idle timeout and the server starts immediately before measurement, so its workers cannot reach that timeout in the 11-19s window. The vanished thread is the JDK `HttpURLConnection` client's `Keep-Alive-Timer`, spawned by the harness's `rawRequest()`.

Local JDK 17 repro (isolated, no project deps — one `HttpURLConnection` request + `disconnect()`, then watch the start-snapshot threads over 15s):
```
BEFORE has Keep-Alive-Timer? true
t=15.0s VANISHED: Keep-Alive-Timer
```
With `-Dhttp.keepAlive=false`: `BEFORE has Keep-Alive-Timer? false`, nothing vanishes. Exactly your pattern and timing.

Pushed the corrected fix (drops the Jetty pin entirely):
- **Fix the actual lifecycle**: the method-compatibility fork now runs with `-Dhttp.keepAlive=false`, so the client never spawns `Keep-Alive-Timer`. What's measured is the server's request-serving CPU; disabling client connection reuse is symmetric across base and candidate.
- **Name the offender**: `measure()` now reports each vanished thread by `name#id`, captured at the start snapshot (names are only reported, never used for an accounting decision). The present-then-gone contract asserts the failure names the thread — so a future non-request lifecycle is identified, not guessed.
- The `validate-cpu-accounting` `graphCount=36` `prefix,contains` smoke proves it end to end, because that fork runs with `http.keepAlive=false` — it passes because the cause is addressed, and would still abort without the flag (the timer exits mid-window on those scenarios).

Verified locally: `:explore:compileJmhKotlin` (candidate, and with origin/main's harness installed over the candidate `RequestCpuAccounting` — the `build-explore-jmh` scenario), the contract (`present-then-gone failed closed = true, named = true`), 90 node gate tests, and YAML all pass. Branch name is now a misnomer; I'll leave it to avoid churning the PR, and have retitled the PR to match the fix.

---

johnsonlee: CI confirms the fix on `6fa46fa`: **`validate-cpu-accounting` passed** — the candidate harness (with `http.keepAlive=false`) ran the contract plus `graphCount=4/count` and `graphCount=36` over `prefix,contains` with `-foe true`, and completed. The 36-graph smoke passes because the cause is addressed.

A second, independent confirmation of the root cause landed in the same run's paired shards (which install the base harness, keep-alive still on): only `method-compatibility-36-scan` and `36-string` aborted, while `4-*`, `17-*`, `36-position`, and `36-aggregate` all passed. The Keep-Alive-Timer only exits inside the window on the longest scenarios — exactly your 11-19s timing — so only those two trip on the unfixed base harness, and every shorter scenario finishes before it.

Remaining reds on this PR are all expected and not this PR's own:
- `method-compatibility-{36-scan,36-string,gate}`: the paired shards run the pre-merge base harness, which lacks this fix; they clear once #122 lands on `main`.
- `global-wide-pressure-evidence` / `graph-routing-pressure-evidence`: harness-only PR, so both sides run identical production code and cannot meet the speedup gate (same as #119).
- `benchmark-regression-gate`: the rollup of the above.

Holding for your merge decision; once #122 lands I'll rebase #121 (the metric switch) onto the fixed base and #117 picks up the fixed harness on its next main merge.

---

johnsonlee: Verified on exact head `6fa46fa`: the Jetty pin is removed, the fork disables `HttpURLConnection` keep-alive, vanished threads are now named diagnostically, and the candidate-owned 36-graph `prefix,contains` smoke completed with `-foe true`. This directly addresses and reproduces the reported lifecycle rather than weakening the fail-closed accounting. Resolving.

PR #123, opened 2026-09-10
Account for mid-window thread exits with an in-window CPU sampler

johnsonlee:
## Why

The method-compatibility CPU accounting (`RequestCpuAccounting`) keys request-serving work by **Java thread identity** and read each thread's CPU only at the window's **two ends**. A thread that exited mid-window failed the end read and **aborted the whole measurement** (fail-closed).

Disabling the JDK client's `Keep-Alive-Timer` (#122) removed one such thread, but a **second, default-named incidental thread** still vanished intermittently on the longest 36-graph scan windows and re-aborted the harness. Whack-a-mole on named threads cannot close this: the failure mode is *any* non-request thread exiting inside the window.

## What

Replace the two-ended read with a **background daemon sampler** that refreshes each Java thread's most recent on-CPU time on a fixed **10 ms** interval across the window:

- A thread that exits mid-window keeps the CPU it accrued **up to its last sample**, instead of being lost. No incidental thread can abort the accounting.
- Bound: at most **one sample interval** of any single thread's CPU can go unattributed, and that remainder still surfaces in the process residual (`jvmInternalCpuNanos`).
- The sampler thread **excludes its own id** from the sum.
- `javaThreadCpuNanos` is now the sum over sampled threads of each thread's `latest − before` delta; a thread born inside the window contributes its whole in-window CPU.

`RequestCpuSample`'s shape (and the published secondary metrics) are unchanged, so the paired gate and the metric-switch work downstream are unaffected.

## Contract & smoke

- Contract checks flip from **asserting fail-closed** to **asserting the work is accounted (≥90%)**: a transient worker born and joined inside the window, and a worker present at the start that exits mid-window, must both be counted.
- The `validate-cpu-accounting` smoke broadens from `prefix,contains` to the full **scan group** (`contains, regex, or`) plus `prefix` at 36 graphs — the longest windows, where the incidental-thread churn appeared — and asserts a valid `javaThreadCpuNanos` row on each.

## Validation (local)

- `:explore:compileJmhKotlin` — candidate **and** base-harness-installed tree (build-explore-jmh scenario): compiles.
- `:explore:jmhJar` builds; `MethodCompatibilityCpuAccountingContract` **PASS** (colliding-name 202M/200M, transient 196M/200M, present-then-gone 195M/200M, idle & collector-loaded within slack).
- `benchmark-gate.test.mjs`: 90/90.
- `benchmark.yml`: YAML valid.

## Landing order

This is a prerequisite harness repair (same role as #122): it must merge to `main` before the method-compatibility shards — which install the base harness over both trees — benefit. The candidate-owned `validate-cpu-accounting` job proves the new harness in a real fork pre-merge.

---

### DIFF
--- a/graphite-explore/src/jmh/kotlin/io/johnsonlee/graphite/cli/RequestCpuAccounting.kt
+++ b/graphite-explore/src/jmh/kotlin/io/johnsonlee/graphite/cli/RequestCpuAccounting.kt
@@ -2,6 +2,8 @@ package io.johnsonlee.graphite.cli
 
 import java.lang.management.ManagementFactory
 import java.lang.management.ThreadMXBean
+import java.util.concurrent.atomic.AtomicBoolean
+import java.util.concurrent.atomic.AtomicLong
 import kotlin.system.exitProcess
 
 /** One measured window of the request-serving CPU accounting. */
@@ -24,18 +26,17 @@ internal data class RequestCpuSample(
  * that happens to share a HotSpot thread's name (`/proc/<tid>/comm` is mutable and truncated to
  * 15 bytes) cannot be mistaken for one and dropped.
  *
- * A thread present at both ends of the window contributes its after-minus-before delta; one born
- * inside the window and still alive at the end contributes its whole lifetime CPU (its start value
- * is zero). The two cases the two snapshots cannot see are handled by failing closed rather than
- * under-reporting:
- * - a Java thread present at the start that is gone at the end took its in-window CPU with it;
- * - a Java thread created and finished entirely inside the window is in neither snapshot, but the
- *   JVM's [ThreadMXBean.getTotalStartedThreadCount] still counted its start, so more starts than
- *   newly-present threads means such a worker existed.
+ * A background sampler refreshes each Java thread's most recent on-CPU time on a fixed interval
+ * over the window, so a thread that exits mid-window still contributes its CPU up to its last
+ * sample rather than being lost (or forcing the window to fail closed on a benign, incidental
+ * thread such as the JDK client's `Keep-Alive-Timer`). At most one sample interval of any single
+ * thread's CPU can go unattributed, and that remainder still shows in the process figure. The
+ * sampler thread excludes itself from the sum. A thread present at both ends contributes its
+ * after-minus-before delta; one born inside the window contributes its whole in-window CPU.
  *
  * `processCpuNanos` is kept as an advisory figure (the whole process, read innermost so its
  * interval is contained in the thread snapshots); `jvmInternalCpuNanos` is the residual
- * `process - java`, the share spent on the JVM's native threads.
+ * `process - java`, the share spent on the JVM's native threads (and on the sampler itself).
  */
 internal object RequestCpuAccounting {
     /** Fails closed, without allocating or starting threads, when the accounting is unavailable. */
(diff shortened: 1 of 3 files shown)


johnsonlee:
## `method-compatibility-gate` red is the landing-order trap, not a defect in this PR

On `5d57474`, every gate is green **except** `method-compatibility-gate` (7/12 shards: `4-scan`, `4-aggregate`, `17-scan`, `17-position`, `36-scan`, `36-string`, `36-aggregate`). All 7 abort identically:

```
IllegalStateException: 1 Java thread(s) present at the start of the window were gone at the end (Thread-2#19); their in-window CPU cannot be accounted
  at RequestCpuAccounting.measure$explore_jmh(RequestCpuAccounting.kt:79)
```

That message and `RequestCpuAccounting.kt:79` are the **old** two-ended, fail-closed harness — the exact throw this PR removes (line 79 in this PR's file is blank; the "gone at the end" string no longer exists). So the abort comes from the **base** jar, not the candidate.

**Why the base jar runs the old harness:** the paired gate builds two jars. `Install base-owned Explorer harnesses` overlays only `ExplorerMemoryBenchmark.kt` and `CypherCapacityBenchmark.kt` from the base; `RequestCpuAccounting.kt` stays per-revision. So the base jar carries `main`'s old fail-closed accounting, which intermittently aborts when an incidental non-request thread (`Thread-2#19`) exits mid-window. This PR's file only becomes the base side once it is **on `main`** — the same prerequisite/landing-order pattern as #122.

**The candidate side is proven:** `validate-cpu-accounting` (candidate-owned harness = this PR's sampler) is **green**, having run the contract plus the 36-graph `prefix,contains,regex,or` scan group — the longest windows and the exact churn case — to completion with valid `javaThreadCpuNanos`.

**Standing down on this gate:** it cannot go green from within the PR, because its base side is `main`'s harness, which no diff here can change. Re-running is not a fix — 7 independent shards at this abort rate have ~0.2% odds of all clearing, and the cause is deterministic, not a flake. The base-side abort rate (7/12, including short 4-graph windows) shows `main`'s method-compat gate is currently flaky for **every** PR; merging this fix is what repairs the base side for this PR and all downstream ones.

---

johnsonlee:
## `global-wide-pressure-evidence` red is also not this PR's code

`global-wide-status.json` reports `passed:false` (candidate global-wide P95 over the 10x line on at least one row). This PR's diff is confined to the jmh CPU-accounting harness (`RequestCpuAccounting.kt`) and CI (`benchmark.yml`, `benchmark-gate.test.mjs`) — it changes **no** webgraph/query production code. So the candidate's global-wide latency is byte-for-byte `main`'s latency; this gate is measuring `main`'s pressure behavior, not anything introduced here.

Getting global-wide under the 10x line is the separate, ongoing optimization tracked on #117 — it cannot be moved from this harness-only PR. Standing down on it here for the same reason as the method-compat gate: no diff in this PR can change it.

Net state on `5d57474`: every gate green except the two whose reds originate outside this PR's diff — `method-compatibility-gate` (base-owned old harness, landing-order) and this one (`main`'s global-wide pressure). The harness fix itself is proven by `validate-cpu-accounting` (green, including the 36-graph scan group).

---

graphite-explore/src/jmh/kotlin/io/johnsonlee/graphite/cli/RequestCpuAccounting.kt
```diff
@@ -313,12 +267,22 @@ object MethodCompatibilityCpuAccountingContract {
...
+    }
 }
 
 private const val CPU_ACCOUNTING_WORKER_NANOS = 200_000_000L
 
-/** A distinctive name the present-then-gone contract asserts appears in the vanished-thread failure. */
-private const val PRESENT_THEN_GONE_WORKER_NAME = "cpu-accounting-present-then-gone"
+/** How often the background sampler refreshes per-thread CPU; the most any one vanishing thread's
+ * CPU can go unattributed. Small enough that a 200 ms worker is captured to well over 90%. */
+private const val CPU_ACCOUNTING_SAMPLE_INTERVAL_MILLIS = 10L
```
johnsonlee, line 330: [P1] The 10 ms sampler makes the new blocking CPU metric evadable by short-lived workers. The stated bound is per thread, not per request: each thread whose entire lifetime falls between samples contributes zero, so splitting work across N such threads can lose N intervals. I reproduced this algorithm on the local JDK with 100 sequential workers doing 2 ms CPU each: the workers recorded 200.0 ms, latest-minus-before accounted only 30.0 ms (15.0%), while process CPU recorded 216.2 ms. Because PR #121 makes javaThreadCpuNanos blocking and leaves process CPU advisory, a candidate can shift substantial CPU into sub-interval workers and still pass. The current contract only burns 200 ms in one worker, so it cannot catch this. Please either retain process CPU as a blocking backstop or use accounting that preserves completed-thread CPU, and add a many-short-lived-workers contract.

johnsonlee: Confirmed and valid — a worker that both starts and ends inside one 10 ms interval is never sampled, and a dead thread's CPU can't be read back, so `javaThreadCpuNanos` drops it entirely. Reproduced your attack in the new contract: 100 sequential 2 ms workers burn ~206 ms, the sampler accounts only ~52 ms (25%), while process CPU records ~280 ms.

Fix pushed in `ecfe079`, taking the **process-CPU-as-blocking-backstop** option:

- **New contract** `checkManyShortLivedWorkersAccountedByProcessBackstop` (100 sub-interval workers) asserts `processCpuNanos` accounts the burned CPU to ≥90%, and deliberately does **not** require the Java-thread sum to — that's the hole the backstop closes. It runs inside `MethodCompatibilityCpuAccountingContract`, so `validate-cpu-accounting` exercises it every run.
- **Docs corrected**: the old "most any one vanishing thread can go unattributed" bound was wrong (it's a whole-lifetime loss, unbounded across N sub-interval workers). `RequestCpuAccounting` and the sample-interval constant now name `processCpuNanos` the blocking backstop, not advisory.

The gate half — keeping `processCpuNanos` **blocking** (with a wider allowance that absorbs the collector/JIT swing) instead of flipping it fully advisory — is the metric-switch's job, so I'll carry it in #121 when I rebase it onto this once #123 lands, rather than muddy the metric selection here. Leaving this thread open until that gate change is in.

---

johnsonlee:
## CI state on `ecfe079` — all red gates are outside this PR's diff

This PR changes **only** the jmh CPU-accounting harness (`RequestCpuAccounting.kt`) and CI (`benchmark.yml`, `benchmark-gate.test.mjs`) — **zero production code**. Every benchmark/pressure/latency gate therefore measures code byte-identical to `main`, so their reds are base-level noise or the landing-order trap, none introduced here:

- **`validate-cpu-accounting` (the one gate that actually exercises this PR's harness): green** — contract + 36-graph scan group + the new many-short-lived-workers backstop check all pass.
- **`method-compatibility-*`** — two orthogonal not-this-PR causes: (a) the paired gate's **base** jar runs `main`'s old fail-closed harness, which intermittently aborts on an incidental vanished thread (the landing-order trap this PR fixes once merged); (b) the shards that ran to completion failed the **process-CPU** comparison on identical code — e.g. 4-aggregate showed `jvmInternalCpuNanos` 302 ms of a 700 ms process figure (43% collector/JIT), exactly the hosted-runner swing the metric switch (#121) exists to remove.
- **`global-wide-pressure-evidence`** — `main`'s global-wide P95 vs the fixed 10x line; #117's optimization domain.
- **`wrapped-query-latency` (real-a)** — `main`'s wrapped-query P95 vs the fixed pre-PR-95 reference (ran to completion, p95 392 ms, failed the confirmation); run-to-run variance on identical webgraph code, passed on the prior head `5d57474`; #117's wrapped-query-margin domain.

So from this PR's own diff there is nothing to fix — re-running shards can't turn the aggregate green while `main` carries the old harness and sits near the pressure lines. Landing `#123` repairs the base-side harness for every downstream PR; `#121` then carries the metric switch **with `processCpuNanos` kept as a blocking backstop** (the P1 fix's gate half). I'm keeping the PR watched.

---

graphite-explore/src/jmh/kotlin/io/johnsonlee/graphite/cli/RequestCpuAccounting.kt
johnsonlee: Confirmed on exact head `ecfe0794`: the new 100×2 ms contract and the corrected documentation are present, but the gate is intentionally not fixed in this PR. PR #121 is still at `37aaf0f` and still makes `processCpuNanos` advisory, so the evasion remains possible in the composed path today. Keeping this thread open is correct; resolve it only after #121 is rebased and the authoritative workflow proves that process CPU remains blocking alongside `javaThreadCpuNanos`.

PR #125, opened 2026-09-12
Gate per-query Cypher correctness and warmed P50/P95 under 8 GiB

johnsonlee:
Give each Cypher query an independent correctness and warmed P50/P95 gate under an actual 8 GiB maximum heap. Pooling percentiles across different queries can hide individual regressions; this change retains every call and evaluates each query separately.

## Scope

Benchmark harnesses, CI tooling/tests, reporting and documentation only; no production `src/main` changes. #126 and #127 have merged, and main `3af3284` (#128) is in this PR's ancestry after rebasing. Runtime scheduler/cancellation work remains in #116.

## Acceptance contract

- **72 distinct Cypher queries:** original 34 plus 38 supplemental queries, including AND/OR/mixed predicates, four-term OR, zero/single/multiple graph hits, and ordinary/DISTINCT projections. Performance inputs are 64 real persisted graphs from four pinned corpora.
- **Actual maximum heap 8 GiB.** Every query warms up for at least 10 seconds AND 5 calls, then measures for at least 10 seconds AND 40 calls. Each phase stops at the first eligible boundary. Every warmup and measured result is checked against the complete sealed oracle; every measured call contributes to the quantiles.
- **Three fresh JVM pairs**, serially ordered candidate/base, base/candidate, candidate/base. Each query's base/candidate measurements run on the same machine without overlap. The two measurement jobs cover 71 standard queries and one full-scan query; aggregation verifies the complete, disjoint 72-query set.
- **5% is the paired regression threshold.** Every paired candidate P50 and P95 must be strictly below 105% of the corresponding base value; exactly 5% fails. Each revision's cross-fork `(max-min)/min` is reported as a diagnostic, with no pass/fail threshold. Its recorded values must still match the measurements. A paired exceedance is an observed regression risk, not proof that candidate code caused it. No uncalibrated confidence-interval gate is added.
- Partial results never pass. No filtering, averaging away failures, or retry-to-pass policy is introduced. Process CPU, peak heap/RSS growth, startup/cold latency and historical speedup targets are diagnostic. Correctness, actual heap, accounting/evidence integrity, existing allocation/GC guardrails and Explorer memory stability remain blocking.

## Implementation

Main #128 adds the independent `slow-query-shapes` component. CI now uses its new steady-state entry point for the original 12 Android hit/miss queries: each fresh JVM warms for at least 10 seconds AND 5 calls, then measures for at least 10 seconds AND 40 calls. Three candidate/base, base/candidate, candidate/base pairs run serially per query on one host. Every call checks the complete ordered result oracle; every measured sample contributes to per-query P50/P95, with the same strict <5% paired threshold and diagnostic-only fork spread. The original single-shot entry point remains available for explicit standalone use. The 72-query/64-graph catalog is unchanged.

The slow-query driver checks actual heap and runtime, executable JAR hashes, every private fixture's identity and cleanup, complete raw evidence, and shared fixture bytes even after failure. Pair-one numerical suspects retain evidence through the reversed pair; integrity errors stop immediately. Partial checkpoints cannot pass, and aggregate reporting rejects stale policy/revision or incomplete passing summaries. Its job budget is 180 minutes, allowing for 72 JVMs, private copies and builds. Exact JARs and the original persisted Android graph are archived separately so the report aggregator need not download them. Preserving the graph enables follow-up profiling to verify every file against the failed run: rebuilding otherwise changes the properties-file timestamp. SHA-pinned controls cover the new harness, driver and comparator.

The harness caches reflective metric metadata while reading and resetting actual values on every call. Complete raw rows are flushed outside the query timer before correctness verification or the next invocation, preserving evidence after cancellation. Canonical signatures cover values, row order, query identity and graph provenance; receipts bind revisions, artifacts, protocol and fixture bytes. Comparison results and shard receipts identify the acceptance policy, and downstream consumers reject older policy verdicts without re-evaluation.

On cache misses, two independent jobs generate the primary and repeat 64-graph corpora concurrently from identical sealed inputs/runtime. Preparation compares both corpora, executes tamper checks and publishes a shared artifact. Cache hits authenticate restored bytes and their reproducibility receipt. Explicit downstream dependency conditions allow intentional producer skips without skipping valid consumers. Fixture generation may run on separate machines; paired latency measurement stays together.

Runtime errors stop immediately, and completed-pair integrity failures stop before the next pair. First-pair paired latency exceedances are retained while the reversed second pair runs. After pair two, cumulative paired exceedances stop remaining work; a passing second pair cannot erase a first-pair failure. Cross-fork spread never stops a shard. Successful acceptance still requires three complete pairs. Reports display paired exceedances and integrity failures separately from diagnostic variation.

Watchdogs cancel independent work after failed jobs preserve evidence. Cancellation publication branches on the upstream aggregate result, including when the publisher's own `cancelled()` is false. PR summaries are capped at 16,000 UTF-8 bytes, link complete artifacts, reject stale heads/runs and expose publication errors.

## Validation and outstanding work

Current head: **`383b1a20c1b0bccc7cc029030cb3b6cb7b603e57`**. **Not ready to merge:** exact-head CI and complete warmed 72-query/three-pair acceptance remain outstanding; the open review has not been resolved.

| Evidence | Result |
|---|---|
| Local checks | 355 Node tests pass; workflow/shell lint and checksum checks pass. `:webgraph:compileJmhKotlin` passes with Java 17 for the final modified harness in a normal clone (the local publication plugin does not support linked-worktree metadata). Independent review verified slow-query aggregation, watchdog coverage and both authenticated controls selectors. |
| Previous head fb45896 CI | [Unit CI](https://github.com/johnsonlee/graphite/actions/runs/34765457721) PASS, including coverage. Method and large-corpus E2E gates PASS. [Benchmark CI](https://github.com/johnsonlee/graphite/actions/runs/34765457707) stopped after `qualifiedIdHit` pair 3 exceeded the slow-query threshold: P50 +7.096%, P95 +8.146%. Its first two pairs remained below 5%; completed checkpoints had zero integrity errors. `valueHit` and `valueMiss` completed three passing pairs. The other nine slow queries and full 72-query acceptance remain incomplete because of fail-fast. This observation is being investigated; it is not yet evidence of a candidate-code cause. |
| Exact-head CI | [Unit run 34767352638](https://github.com/johnsonlee/graphite/actions/runs/34767352638), Method and large-corpus E2E PASS. [Benchmark run 34767352655](https://github.com/johnsonlee/graphite/actions/runs/34767352655) failed slow-query-shapes again on qualifiedIdHit pair 3: P50 +6.220%, P95 +6.774%; pairs 1/2 improved on both metrics. The nine completed checkpoints had zero integrity errors. Fail-fast cancelled the remaining wrapped shards and both fixture64 generators. The other nine slow queries and all 72 wide-query acceptance remain incomplete. The only change since fb45896 preserves the original persisted Android fixture as a separate diagnostic artifact. Complete formal latency acceptance remains unproven. |
| Coverage scope validation | Exact class-level XML comparison: 5902/6106 lines (96.659%) becomes 5902/6016 (98.105%). Only 90 uncovered lines in `SlowQueryShapesSteadyState` and its nested sample class are removed; all 121 retained classes have identical counters. Actual application JAR contains neither class; the JMH JAR contains both. Ordinary webgraph tests and `:webgraph:koverVerify` pass. The workflow threshold remains 98%. |
| Superseded 7335110 CI | [Benchmark CI](https://github.com/johnsonlee/graphite/actions/runs/34764639498) failed before compilation: an unnecessary helper copy used `src/jmh` instead of its actual `src/main` location. Removed that copy; an executable setup regression test fails on the old script and reaches Gradle on the fix, preserving revision-owned production bytes. No steady-state samples were produced. [Unit CI](https://github.com/johnsonlee/graphite/actions/runs/34764639527) failed coverage because the two new JMH-only helper classes were counted as application code. The narrow exclusion below fixes that scope error. |
| Superseded d547cc8 CI | [Unit CI](https://github.com/johnsonlee/graphite/actions/runs/34762696639) PASS. [Benchmark CI](https://github.com/johnsonlee/graphite/actions/runs/34762696632) was cancelled after `slow-query-shapes` failed: WARM/wrappedCallerHit mean latency increased 17.16% initially and 24.8% in reverse-order confirmation. Ordered result parity passed. This inherited gate uses one priming call and one timed call per JVM, so its result is not warmed P50/P95 evidence or proof of a candidate-code regression. The two fixture generators stopped 6–8 seconds after failure; the 72-query measurements had not started. Method, all Method compatibility, Explorer, CPU accounting, capacity, resources and large-corpus E2E passed. Method and E2E separately report no blocking regression under their existing thresholds. |
| Superseded fec2397 CI | [Unit CI](https://github.com/johnsonlee/graphite/actions/runs/34761319367), Method, Method compatibility, Explorer, CPU accounting and large-corpus E2E passed against old base 144d98e. [Benchmark run](https://github.com/johnsonlee/graphite/actions/runs/34761319370) had begun wide measurements when main advanced and required this rebase. These checks do not prove acceptance against main 3af3284. |
| [Previous unit CI, head 08aadea6](https://github.com/johnsonlee/graphite/actions/runs/34755125366) | PASS; not current-head proof. |
| [Previous benchmark CI, head 08aadea6](https://github.com/johnsonlee/graphite/actions/runs/34755125410) | Method-level, compatibility and large-corpus E2E checks passed. Standard wide-query measurement stopped after pair two; 87 paired quantile exceedances and 106 same-revision spread exceedances were recorded, with no checkpoint integrity errors. The watchdog cancelled the remaining work. Complete warmed acceptance was not obtained; removing spread failures alone does not establish a pass. |
| Actual-data re-evaluation | Old and new comparators replayed the same 71-query/two-pair raw evidence: quantiles, counts, durations, diagnostic spreads and integrity results are identical. Exactly 106 spread failures removed; all 87 paired exceedances retained. New-policy checkpoint still fails. |
| Reverse-pair and fail-fast behavior | Verified in hosted execution: first-pair exceedances allowed the reverse-order pair, the second checkpoint retained original failures, and the other measurement job was cancelled 18 seconds after the failed standard job finished. |
| Fixture scheduling | Parallel generation/comparison/publication passed on the earlier cache-miss path. Cache-hit execution verified intentionally skipped producers followed by successful bundle construction and both measurement jobs starting. |

The [reverse-order review](https://github.com/johnsonlee/graphite/pull/125#discussion_r3999637204) was resolved after reviewer verification of head `08aadea6`. The [stability review](https://github.com/johnsonlee/graphite/pull/125#discussion_r3999092753) remains open. The acceptance-policy correction removes same-revision spread as a blocking criterion; it does not prove that all paired increases are caused by candidate code or that the remaining paired regression requirement passes.

Isolated profiling remains outside this PR. [Attempt 096](https://github.com/johnsonlee/graphite/blob/25d0a6a6e851524f75cb0e1c05032a6446f05e13/.github/diagnostics/linux-aa/attempt096-results/REPORT.md) retains its original failed experiment and is not adopted. [Attempt 097 report](https://github.com/johnsonlee/graphite/blob/960bb08d7247643c9c184869730bc6c18471c9e2/.github/diagnostics/linux-aa/attempt097-results/REPORT.md) records [run 34758640598](https://github.com/johnsonlee/graphite/actions/runs/34758640598), which completed three serial same-baseline JVMs over the original 34 queries: 1,697,959 calls and 204 phases passed correctness/protocol analysis, and archive/source/runtime provenance was independently verified. Its 29 of 68 cross-fork metrics exceeding 5% are diagnostic variation, not candidate regressions; it contains no candidate pair or formal acceptance result.

[Qualified-ID profiling run 34767807984](https://github.com/johnsonlee/graphite/actions/runs/34767807984) completed three fixed serial same-baseline JVMs using the archived JAR and original Android graph. All 16 exported graph files match the failed run’s hashes. Independent validation checks 386 raw/CPU/JFR calls, including 197 measured calls, full result correctness and runtime/fixture receipts. P50 was 152.858/152.175/151.639 ms; P95 was 164.646/153.536/156.710 ms (7.236% same-version spread). The sustained 160 ms level was not reproduced. Integer ID formatting and mapped ID iteration dominate execution samples; background compilation persists across the measurement phase. This does not establish the cause of the old paired exceedance or substitute for a formal gate. The preceding setup-only run failed before measurements because setup-java rejected a four-component runtime version; the fix uses the formal CI Java 17 selector and checks exact runtime 17.0.20.1+1 before execution. No extra measurement forks were added.

A fixed follow-up with 200 warmup calls ([run 34768410933](https://github.com/johnsonlee/graphite/actions/runs/34768410933)) reduced measurement-phase compilation log counts but did not stabilize latency: P50 152.521/162.272/152.147 ms and P95 153.727/178.564/154.557 ms. All 793 calls, including 193 measurements, passed independent protocol/correctness audits. Warmup extension alone is not adopted. [Both experiments’ allowlisted raw/CPU data and results](https://github.com/johnsonlee/graphite/blob/f1a8cb89e3dc67dfc34e7d972217d935ac4a7fd4/.github/diagnostics/qualified-id-profile/results/README.md) are preserved. The original public diagnostic archives were removed because default JFR environment recording captured the expired workflow credential; future recordings disable that event and query JVMs no longer inherit CI tokens. No credential values are included in the published results.

[Fixed 20-batch A/A calibration](https://github.com/johnsonlee/graphite/blob/a1975279cd40ee06697452d597d9c87bca1f5c1a/.github/diagnostics/qualified-id-profile/results/calibration20/REPORT.md) completed with the actual formal harness, original phase rules, same sealed JAR and graph. Independent verification covers 120 JVMs, 60 pairs and 15,586 calls (7,898 measured). The current any-pair 5% rule flags 8/20 unchanged-code batches: 4/10 training and 4/10 held-out. This establishes observed false alarms in these controls, not a universal 40% rate. A model-conditional three-pair t interval gives P95 5/20 below-margin and 15/20 inconclusive outcomes; it is not adopted as a calibrated formal policy. Uncertainty handling and adequate measurement design remain unresolved; all original PR failures remain recorded.

See the [gate contract](https://github.com/johnsonlee/graphite/blob/fb458965da9d7363144534efec0ac71c8e84de7b/docs/benchmark-regression-gate.md) and [broad-query protocol](https://github.com/johnsonlee/graphite/blob/fb458965da9d7363144534efec0ac71c8e84de7b/docs/large-broad-query-pressure-benchmark.md).

### Measurement-noise design and execution

[Design and execution record](https://github.com/johnsonlee/graphite/blob/18ca83bdb09b40785b5e66c214238cbdc404edbd/docs/benchmark-noise-reduction-design.md) preserves the per-query5% latency threshold, full correctness oracle, real fixtures and8GiB heap. Formal gate changes remain unshipped pending noise validation.

The first preregistered same-host ASLR experiment completed20/20 jobs (240JVMs;34,355calls) with full independent raw integrity verification. It **did not meet validation criteria**: disabled/control paired-log-ratio MSE was0.938 forP50 and1.051 forP95 in the held-out10jobs. These are noise metrics, not latency regressions. ASLR disabling is not adopted. [Results](https://github.com/johnsonlee/graphite/blob/18ca83bdb09b40785b5e66c214238cbdc404edbd/.github/diagnostics/qualified-id-profile/results/aslr20/REPORT.md).

Runner inventory found heterogeneous CPU models. Hardware counters were unavailable even to root on the testedEPYC9V74; some Intel jobs expose frequency interfaces whose control effectiveness is unverified. [Scoped capability evidence](https://github.com/johnsonlee/graphite/blob/18ca83bdb09b40785b5e66c214238cbdc404edbd/.github/diagnostics/qualified-id-profile/results/perf-capability/REPORT.md). Further environmental validation and full formalCI acceptance remain outstanding; this PR is not yet merge-ready.


.github/scripts/benchmark-wide-latency.mjs
```diff
@@ -0,0 +1,146 @@
...
+            }
+            runs.push(run);
+        }
+        const stability = {};
+        for (const revision of ["base", "candidate"]) for (const quantile of ["P50", "P95"]) {
+            const values = runs.map(run => run[`${revision}${quantile}Nanos`]);
+            if (values.length !== WIDE_FORK_COUNT) continue;
+            const min = Math.min(...values), max = Math.max(...values);
+            stability[`${revision}${quantile}SpreadPercent`] = (max - min) / min * 100;
+            if (BigInt(max) * 100n >= BigInt(min) * 105n) {
```
johnsonlee: [P1] Calibrate the hard 5% stability rule before making it authoritative

The exact-head run `34702768735` is a direct falsification of this threshold: 60 of the 72 independent query checks failed even though this PR changes only benchmark/workflow code, not the query engine. Most failures are this base-side spread check, so they cannot be candidate regressions. For example, `or-four-single-late-distinct` has base P95 values 8.401/8.475/9.104 ms (8.37% spread) while the candidate spread is only 1.99%; `or-four-all-distinct` also fails on base P95 spread. Conversely, `or-four-zero-rows` fails because one paired P95 delta is only about +5.8% (1.933 -> 2.045 ms) despite both sides having <4% cross-fork spread. A rule that hard-fails a benchmark-only PR on normal base variance makes the authoritative gate permanently red and does not distinguish regressions from runner noise. Derive the threshold from repeated same-revision controls and/or use a paired confirmation/statistical interval; base instability should invalidate/retry that measurement rather than mark the candidate as regressed.

.github/scripts/benchmark-gate.mjs
```diff
@@ -2118,6 +2164,7 @@ export function renderGlobalWidePressureReport(comparison) {
         comparison.passed ? "**Result: PASS**" : `**Result: FAIL**\n\n${comparison.errors.join("\n")}`,
         ""
     ].join("\n");
+    return comparison.repeatedLatency ? `${renderWideLatency(comparison.repeatedLatency)}\n${legacy}` : legacy;
```
johnsonlee, line 2186: [P1] Keep the PR comment below GitHub’s size limit and fail if current-head evidence cannot be published

This concatenates the full 72-query table and every repeated-latency error into the aggregate report. On exact-head run `34702768735` the body grew to roughly 298 KB; GitHub rejected the issue-comment update with HTTP 422 `Validation Failed`. Because the existing attach step is `continue-on-error`, `benchmark-comment` remained green and PR #125 still displays the stale `afd07db6` report (`global-wide: MISSING`) instead of the `701b641b` result (global-wide FAIL, 60/72 query checks red). Render a bounded PR summary with a link to the full artifact, assert the head SHA/size, and make publication failure visible; add a 72-query worst-case size test.

.github/scripts/run-wide-latency-shard.sh
```diff
@@ -0,0 +1,56 @@
...
+      -jvmArgs "-Xmx8g -Dgraphite.broad.pressure.graphs=$SHARED/graphs/graphs.tsv \
+        -Dgraphite.broad.pressure.correctness.mode=verify \
+        -Dgraphite.broad.pressure.latency.only=true \
+        -Dgraphite.broad.pressure.latency.shard=$SHARD \
+        -Dgraphite.broad.pressure.latency.oracle=$BUNDLE/oracle.correctness \
+        -Dgraphite.broad.pressure.latency.output=$OUTPUT/$REVISION-$FORK.tsv"
+  done
+  # A completed pair's regression, or an observed cross-pair spread, cannot be repaired
+  # by later forks. Preserve raw evidence and an explicit failed checkpoint before stopping.
+  node "$SCRIPT_DIR/benchmark-wide-shards.mjs" check-progress \
```
johnsonlee, line 54: [P1] Do not stop before the reverse-order control runs

`check-progress` treats any single-pair latency miss as terminal, so pair 1 (`candidate-base`) can abort before pair 2 reverses the order. Exact-head run 34734536122 did exactly that: pair 1 completed with zero integrity errors, then 18 of 71 standard queries crossed the 5% threshold and this call exited 1; the `base-candidate` and third pairs never ran. Therefore the result cannot distinguish a candidate regression from order/thermal/runner drift, and it never reaches the three-pair/cross-fork contract the PR claims as its gate. Early termination is safe for integrity failures, but latency must at least run the reverse-order confirmation (and the full three pairs if cross-fork stability is part of acceptance).

.github/scripts/benchmark-gate.mjs
johnsonlee: Verified on exact head `e9752057`: PR publication now renders a bounded summary capped at 16 KB, keeps the complete 72-query report in an attempt-scoped artifact, binds candidate SHA/run/attempt and required gate result, rejects stale writers, and no longer swallows HTTP 422. The tests cover a greater-than-298-KB worst case, stale head/attempt, missing or ambiguous artifacts, and failed/skipped/cancelled required verdicts. This thread is genuinely fixed.

.github/workflows/benchmark.yml
```diff
@@ -2979,57 +3356,120 @@ jobs:
...
+    - name: Checkout exact-head publication code
+      if: ${{ !cancelled() }}
+      uses: actions/checkout@v7
+      with:
+        ref: ${{ github.event.pull_request.head.sha }}
+        persist-credentials: false
+
     - name: Download aggregate benchmark report
-      continue-on-error: true
+      if: ${{ !cancelled() }}
```
johnsonlee: [P1] Select the cancellation publisher from the upstream result, not `cancelled()`

Exact-head run 34739834718 proves this branch selection does not work. The workflow was cancelled by the fail-fast monitor and `benchmark-regression-gate` was cancelled, but the new `benchmark-comment` job started with `cancelled()` false. It therefore ran this normal download, failed because no aggregate report artifact existed, attempted the normal attach, and skipped `Publish bounded cancellation evidence`. The PR comment is still the stale `6d4c333f` report instead of current head `e9752057`. Branch on `needs.benchmark-regression-gate.result` (or another explicit upstream verdict)—for example normal only when it is not `cancelled`, cancellation publisher when it is `cancelled`—and add a workflow-level test that models the actual downstream job context. The current extracted-script unit test cannot catch GitHub expression semantics.

johnsonlee: Verified on exact head `e9d4ebe2`: all normal publication steps now additionally require `needs.benchmark-regression-gate.result != cancelled`, while the API-only path selects on the complementary upstream result. The test explicitly reproduces the prior state (`cancelled() == false`, upstream result `cancelled`) and selects only the cancellation publisher. This fixes the expression-routing defect; the current hosted run is still in progress, so actual cancellation publication remains useful end-to-end confirmation rather than a condition for resolving this code bug.

.github/scripts/benchmark-wide-latency.mjs
johnsonlee: The new Linux same-JAR control strengthens this P1 rather than resolving it. I independently recomputed the measurement quantiles from artifact `linux-same-jar-aa-34741006640-1` (243,247 retained calls): two fresh JVMs running the identical sealed base JAR show class-pair-targeted P95 spread 7.9345%, provenance-targeted P50/P95 5.5227%/7.1112%, and aliased-targeted P95 7.2491%, with all correctness checks passing. Therefore the current hard 5% rule rejects same-revision controls on the intended Linux environment. Presentation-only relabeling in `26257284` does not calibrate or change the authoritative verdict; this remains unresolved.

.github/scripts/run-wide-latency-shard.sh
johnsonlee: Reopening: relabeling the result as an observed threshold exceedance does not make the authoritative acceptance decision valid. The Linux same-JAR controls now prove that a single fresh-JVM pair can exceed the 5% limit without any candidate change (for example, class-pair-targeted P95 spread 7.93% and provenance-targeted P50/P95 5.52%/7.11%). Pair 1 is therefore not irreversible evidence of a candidate regression. Because this runner intentionally alternates candidate-base and base-candidate to control order/JIT effects, terminating before pair 2 defeats that control; a separate diagnostic run cannot repair an authoritative red gate after the fact. Keep early termination for integrity/correctness failures, but require at least the reverse-order pair before making the terminal latency decision.

johnsonlee: Verified on exact head `08aadea6`. The partial comparator now allows continuation after pair 1 when the only findings are numeric latency exceedances (`expectedForks === 1`), while integrity errors still stop immediately. After pair 2, any retained numeric failure makes `canContinue` false, so the reverse-order pair cannot erase the first observation. The shell-level tests exercise the actual checkpoint path and confirm pair 2 runs and the original failure remains in the terminal verdict. This fixes the specific premature-stop defect; resolving this thread. The separate hard-5% stability-policy thread remains unresolved.

.github/scripts/benchmark-wide-latency.mjs
johnsonlee: Head `d547cc8` fixes the base-instability half of this P1 by making cross-fork spread diagnostic, but the acceptance rule still hard-fails any single paired P50/P95 increase >=5%, even when the reversed pair passes. The original evidence includes exactly that case (`or-four-zero-rows`, ~+5.8% with both sides individually under 4% spread), and the Linux same-JAR control shows two identical fresh JVMs can differ by 5.5–7.9%. Running pair 2 is useful, but retaining pair 1 as an irreversible failure is not paired confirmation; it still rejects identical code on ordinary JVM/order variance. Require the candidate slowdown to confirm in the reverse order (or use a calibrated statistical rule) before making a numeric miss terminal. This thread remains unresolved.

johnsonlee: Exact-head `fb458965` now reproduces this false-positive in the new steady-state gate. Artifact `benchmark-slow-query-shapes-125-1` has complete correctness/integrity for `qualifiedIdHit`: pair 1 is 159.281/162.226 ms base vs 158.472/161.060 ms candidate (candidate faster), reversed pair 2 is 149.556/154.437 vs 151.481/156.000 ms (+1.29%/+1.01%), while pair 3 alone is 149.589/150.489 vs 160.203/162.748 ms (+7.10%/+8.15%). #125 has no production `src/main` diff, so both sides execute the same #128 qualifiedId implementation. The job nevertheless fails because one of three pairs crosses 5%. Base cross-fork variation is 6.50%/7.80%, consistent with ordinary JVM drift, and neither of the first two orders confirms a regression. This is concrete same-exact-head proof that diagnostic spread plus an irreversible any-pair 5% rule still rejects unchanged production code. Require consistent reverse-order confirmation or another calibrated decision rule; the current red gate is not evidence of a candidate regression.

johnsonlee: Exact-head `383b1a20` reproduces the same false-positive after the fixture-preservation-only update. In run 34767352655, `qualifiedIdHit` pair 1 is candidate faster by 0.96%/2.66% (P50/P95), reversed pair 2 is candidate faster by 0.41%/0.77%, and pair 3 alone is +6.22%/+6.77%, which trips the irreversible 5% rule. Cross-fork spreads are 6.06%/5.83% for base and 6.00%/6.73% for candidate. The prior two orders again do not confirm a regression, while the third pair lies inside the observed same-run drift. The new commit preserves the original fixture for diagnosis but does not change the decision rule, so this thread remains unresolved.

PR #127, opened 2026-09-13
Fix benchmark startup; make cold latency and resource growth diagnostic

johnsonlee:
CI currently has three benchmark blockers: Javalin’s transient startup watcher can invalidate CPU snapshots; the 34-query cold suite demands 5× acceleration in every pair; and resource-growth limits reject otherwise passing latency measurements. This benchmark-only prerequisite repairs sampling and makes cold latency and resource growth diagnostic, before the independent runtime fix in #126 and the full warmed per-query gate in #125.

Retire the startup watcher before sampling and run the actual server lifecycle contract. Both revisions use the same hash-pinned harness repair; CPU accounting stays strict and base-owned.

The CI driver explicitly selects cold diagnostic evaluation. Cold latency and historical speedup misses remain visible, while correctness, 8 GiB, graph coverage, worker limits and measurement integrity still block. Direct global-wide comparator calls retain their strict default; diagnostic evaluation cannot publish the old strict-target success context. Graph-routing warm checks retain their existing latency rules. Startup-prepared only prebuilds indexes and does not warm queries, so its first-query latency is diagnostic with correctness and index integrity still required.

Method CPU/RSS, capacity CPU/RSS and wrapped retained/peak heap growth are also diagnostic in both initial and final comparisons. Invalid or missing measurements, heap limits, result correctness, cancellation/recovery, wall/tail latency, allocation/GC and Explorer memory-stability assertions remain blocking. Comparison failures retain fixture and build provenance instead of exiting before metadata is written.

Acceptance-policy decision: removing process-CPU and peak-heap/RSS growth blockers is an explicit requirement of this change, to focus acceptance on diverse queries, correctness and warmed latency at an actual 8 GiB heap. There is intentionally no replacement CPU-growth blocker in this prerequisite, and landing #121 is not a prerequisite for that policy. CPU usage remains visible; missing/invalid measurements and the CPU-accounting contracts still fail. The [CPU-policy review](https://github.com/johnsonlee/graphite/pull/127#discussion_r3998742891) proposes a different acceptance policy; restoring that blocker would contradict this PR's requested scope. Green checks demonstrate the stated policy, not the absence of CPU growth.

Build and benchmark checks also accept target branch names containing `/`, so stacked PRs receive both checks. Push triggers retain their existing scope.

No production code or 72-query harness is included. #125 separately supplies all 72 warmed P50/P95 gates with strictly <5% paired regression and cross-fork fluctuation under an 8 GiB heap.

Validation:

- 121 Node tests pass at `4cb50e3`, including execution from outside the checkout; actionlint and shell syntax checks pass. Earlier `:explore:check :explore:jmhJar` and fresh-JVM lifecycle/accounting contracts passed; the Kotlin harness is unchanged by the policy follow-ups.
- Replaying the original global-wide CI samples preserves nine historical-target misses and two cold-latency regressions. Strict mode still fails. Explicit cold diagnostic mode has zero integrity errors; a changed result or 4 GiB heap still fails.
- Replaying the two failed method shards preserves their CPU growth (17.1% and 25.3% in reverse-order confirmation) as diagnostic. Their latency checks pass. Deliberately doubling wall latency still fails; allocation/GC and missing/invalid metric tests remain blocking.

The follow-up at `699752f` changes only these two PR branch filters; actionlint passes. GitHub has started both workflows for #127 and its stacked #126.

Hosted [run 34736562534](https://github.com/johnsonlee/graphite/actions/runs/34736562534) passes all 14 method-level checks and the real Tika, Hive and Kotlin compiler end-to-end checks at the earlier head `699752f`. Neither comparison detects a regression under its existing gate. These existing gates do not establish #125’s separate warmed per-query <5% requirement. The earlier `699752f` passed all hosted checks, including `benchmark-regression-gate`. The follow-up `e4e23e5` corrects the startup-prepared classification after #126 exposed that these timings have no query warmup. All 123 local tests pass, including result/index tampering and unchanged warm blocking checks. Replaying #126’s original failure keeps the same K64 latency observation as diagnostic; invalid results or missing indexes still fail. Follow-up `8bfc404` also fixes the zero-valid-run failure path: missing speedups no longer abort description formatting before the complete evidence manifest is written. An executed comparator/driver test verifies all 19 sealed files and the original failure exit; 124 local Node tests pass. This addresses [the manifest review](https://github.com/johnsonlee/graphite/pull/127#discussion_r3998742936). At exact head `8bfc404`, [benchmark run 34738802377](https://github.com/johnsonlee/graphite/actions/runs/34738802377) completed successfully with 39 successful jobs, and [unit run 34738802401](https://github.com/johnsonlee/graphite/actions/runs/34738802401) also passed. This verifies the existing prerequisite gates, not #125’s separate 72-query contract; the CPU-growth policy is explicitly documented above, and no merge has been performed.


.github/workflows/benchmark.yml
```diff
@@ -670,8 +689,8 @@ jobs:
 
         NEED_CONFIRMATION=0
         gate_metric wall '' 'wall time' false || NEED_CONFIRMATION=1
-        gate_metric cpu processCpuNanos 'CPU time' false || NEED_CONFIRMATION=1
-        gate_metric rss-after residentSetAfterBytes 'RSS after query' false || NEED_CONFIRMATION=1
+        gate_metric cpu processCpuNanos 'CPU time (advisory)' true || NEED_CONFIRMATION=1
```
johnsonlee, line 692: [P1] Do not remove the CPU regression gate before its replacement lands

This changes method CPU from blocking to advisory, but the same job does not require or gate `javaThreadCpuNanos`; lines 658-663 only require `processCpuNanos` and RSS. The capacity job likewise drops process CPU without adding a replacement. Since #121 is still open, merging this prerequisite first creates a main-branch window where a candidate can consume arbitrarily more CPU (for example on an extra worker while wall time stays flat) and still pass. The exact-head green method shards only show that the newly advisory comparisons no longer block; they do not close this coverage gap. Either land the trustworthy Java-thread CPU gate in this PR, retain the current blocker until #121 lands, or make the landing atomic so main never has no blocking CPU comparison.

.github/scripts/run-real64-global-wide.sh
```diff
@@ -239,8 +246,9 @@ jq -n \
 
 WORST_P50=$(jq '[.runs[].p50Speedup] | min' "${OUTPUT_DIR}/global-wide-status.json")
```
johnsonlee: [P2] Finish the evidence manifest when the comparator has zero valid runs

The promised failure provenance is still incomplete for an integrity failure that yields `.runs == []`. Both `min` expressions then produce JSON `null`, and `printf %.2f` exits nonzero under `set -e` before `evidence-manifest.json` is written and before the explicit `COMPARISON_EXIT` propagation. This is reproducible with a zero-valid-run comparator result: the status contains the integrity errors, but the hash-bound manifest is missing. Make the description null-safe (or build the manifest without numeric formatting) and add a zero-run failure test so every comparator failure reaches line 270 before exiting.

johnsonlee: Verified on exact head `8bfc4044`: comparator failure now selects the non-numeric description path before formatting speedups, writes the hash-bound `evidence-manifest.json`, then propagates the original comparator exit. The new test executes the real zero-valid-run path, verifies all 19 hashes and confirms no publication occurs. This thread is genuinely fixed.

.github/workflows/benchmark.yml
johnsonlee: Documenting this as an intentional policy does not fix the technical coverage gap: exact-head success now proves wall/correctness/integrity under a policy that permits unbounded relative CPU growth. That is a concrete shifted-cost case, not a request for redundant evidence. Since `javaThreadCpuNanos` already exists after #119, either gate it here, keep process CPU blocking until a replacement lands, or explicitly obtain acceptance of having no CPU-regression criterion. Reopening because no code or landing-order change addressed the finding.

PR #129, opened 2026-09-14
Cross-compile the aarch64 musl release binary with cross-rs

johnsonlee:
The `v3.0.0-alpha1` publish run ([34804206444](https://github.com/johnsonlee/graphite/actions/runs/34804206444)) failed in the `aarch64-unknown-linux-musl` build: jemalloc's C build under Ubuntu's arm64 `musl-tools` compiler stops with `initializer element is not constant` on `ATOMIC_INIT` and `control reaches end of non-void function` throughout, the shape of a configure step that found no usable atomics. The `x86_64-unknown-linux-musl` build of the same package on the same run passed, and the macOS builds passed, so it is the arm64 toolchain, not the code.

## Change

- The aarch64 target is cross-compiled on the x86_64 runner with [cross-rs](https://github.com/cross-rs/cross), whose image carries a musl cross toolchain that builds jemalloc; tikv-jemallocator's own CI builds this target the same way. `Cross.toml` passes `GRAPHITE_VERSION` into the container so the version stamp survives.
- The `--version` self-check for that target runs the static binary under `qemu-user-static`; `require-static.sh`, packaging and the checksums are unchanged. `x86_64` stays a native `musl-tools` build, which passed.

## Verification

- `publish.yml` parses; the matrix, `Install cross and qemu-user`, the branched build step and the `exec` prefix are the only edits.
- The development sandbox has no Docker daemon, so `cross` cannot run here. The first real build is the next publish run; the recommended order is the dry run on `main` after merge (`gh workflow run publish.yml --ref main -f tag=v3.0.0-alpha2`), then the `v3.0.0-alpha2` tag.

---


.github/workflows/publish.yml
```diff
@@ -157,20 +163,30 @@ jobs:
...
+      if: matrix.cross
+      run: |
+        cargo install cross --locked
+        sudo apt-get install -y -q qemu-user-static
     - name: Build release binaries
       env:
         GRAPHITE_VERSION: ${{ env.RELEASE_VERSION }}
-      run: cargo build --release --locked --target ${{ matrix.target }} -p graphite-cli
+      run: |
+        if [ "${{ matrix.cross }}" = "true" ]; then
```
johnsonlee, line 179: [P1] The only code path this PR changes was never run on exact head 4339aa5: the successful Rust, unit, and benchmark workflows do not execute publish.yml or build aarch64-unknown-linux-musl, and there is no Publish run at this SHA. The older failure proves the native ARM path was broken, not that this cross path produces a runnable static release asset. Before the next release, run the dispatchable publish dry-run on the merged content and require this matrix row's cross build, QEMU --version check, require-static.sh, artifact upload, and aggregate verification to pass.

.github/workflows/publish.yml
```diff
@@ -157,20 +163,30 @@ jobs:
...
     - name: Install the musl C toolchain
-      if: contains(matrix.target, 'musl')
+      if: contains(matrix.target, 'musl') && !matrix.cross
       run: |
         sudo apt-get install -y -q musl-tools
         echo "${{ matrix.cc }}=musl-gcc" >> "${GITHUB_ENV}"
+    - name: Install cross and qemu-user
+      if: matrix.cross
+      run: |
+        cargo install cross --locked
```
johnsonlee, line 173: [P1] cargo install cross --locked does not pin cross itself; it selects the latest published cross and only honors that release's dependency lockfile. cross then selects its target image through a mutable version tag, so the compiler/toolchain used for a shipped ARM binary can change without any repository diff. Pin cross to an exact version or revision and configure the aarch64 image by immutable digest in Cross.toml.

johnsonlee (commented):
Post-merge exact-head review found two release-pipeline risks. The cross-rs wiring is internally consistent and an isolated reproduction passes the original jemalloc atomics failure point, but the changed target was not exercised by any PR check.

PR #124, opened 2026-09-10
Graphite 3.0: one native graphite CLI (build, query, serve, mcp) on a Rust backend

johnsonlee:
Graphite 3.0. One native `graphite` binary replaces `graphite.jar` as the installed tool: `build` runs the JVM frontend, `query` and `serve` run the Rust backend, and the Model Context Protocol server is built in (`graphite mcp` over stdio, `POST /mcp` on `serve`), retiring the `graphite-mcp` npm package. The backend reads the same on-disk graph format and serves the same HTTP API as the Kotlin server, checked route by route, so it is a drop-in replacement against an unmodified graph directory. The repository is laid out as `frontend/`, `backend/`, `cli/`, per `docs/architecture-frontend-backend.md`.

The Kotlin modules move to `frontend/jvm/` unchanged but for one fix: `ExploreRoutes.buildSubgraph` walks with an explicit stack instead of recursing once per node, after CI hit a StackOverflowError in `buildSubgraph caps excessive traversals` (a 2,000-hop chain, which the node cap allows). Visiting order and both caps are unchanged; a new test pins the edge cap. `origin/main` (#127, #128) is merged in.

## Layout

```
graphite/
├── frontend/jvm/         Kotlin: core, cypher, sootup, webgraph, query, explore
│                         (Gradle project names :core, :sootup, ... are unchanged, mapped in settings.gradle.kts)
├── backend/              Rust backend: three library crates
│   ├── storage/          persisted-graph reader (mmap), indexes, columns
│   ├── cypher/           Cypher parser, planner, executor
│   ├── explore/          HTTP server, MCP server, UI, C4, topology; Dockerfile for the image
│   └── bench/            differential harness, MCP smoke, backtests, fixtures
├── cli/                  `graphite`, the only binary
└── Cargo.toml            workspace root (version 3.0.0): backend/* and cli
```

Directory names carry no `graphite-` prefix; Cargo package names (`graphite-storage`, ...) and Maven artifactIds are unchanged. The benchmark gate compares a base tree with a candidate tree by path, so every step and script that touches a base, gate, or tagged tree resolves a module through a `jvm <tree> <module>` helper (`frontend/jvm/<module>` when present, else the pre-move `graphite-<module>`); this PR's own gate run, runs after merge, and historical tags all keep working. `docs/architecture-frontend-backend.md` records the graph IR, the core model with backward-compatible property aliases, the migration phases, the Web/iOS frontends, and how frontends are distributed and found by the CLI (§6.6).

[...]

PR #130, opened 2026-09-14
Publish the GitHub Release re-runnably on an immutable-release repository

johnsonlee:
The `v3.0.0-alpha2` publish run ([34810436905](https://github.com/johnsonlee/graphite/actions/runs/34810436905)) got past the fix from #129: all four binaries (the aarch64 musl one via cross-rs included), `graphite.jar` and the Maven artifacts were built and verified. The `release` job then failed while uploading the assets:

```
Cannot upload asset graphite-3.0.0-alpha2-aarch64-unknown-linux-musl.tar.gz.sha256 to an immutable release.
GitHub only allows asset uploads before a release is published ...
```

This repository has immutable releases enabled (`GET /repos/johnsonlee/graphite/immutable-releases` answers, and the alpha2 release reports `immutable: true`). `softprops/action-gh-release` creates the release published first and uploads the assets afterwards, which an immutable release refuses. The `v3.0.0-alpha2` release therefore exists, published and immutable, with zero assets, and the `docker` job (which needs `release`) never ran.

## Change

- **Preparation before finalization.** The `release` job builds the jar, collects the binaries, verifies the checksums, renders and syntax-checks the formula and keeps all of it as the `release-<tag>` artifact (on every run, so the dry run's output keeps the same shape) before the single step that changes remote state.
- **`.github/scripts/publish-release.sh`** replaces `softprops/action-gh-release`: it finds or creates a draft, makes the draft's assets equal to `release-assets/` by name and sha256 (API `digest`, or download-and-hash when absent), verifies, then publishes with a PATCH whose result is not trusted: the release state is read back and decides. A re-run that finds the release already published verifies names and digests against the local assets and uploads nothing, or fails saying the tag cannot be repaired when they differ or the prerelease flag disagrees. Its `--self-test` runs against a stub `gh` on every publish run: fresh tag, re-run after publication with no writes, unrepairable mismatch, draft left by a failed upload with a stale asset, ambiguous publish response.
- **Independently re-runnable consumers.** The tap push is its own `tap` job (needs `release`, stable tags only, formula from the artifact); `docker` already takes its binaries from the published release. Either can be re-run from a failure without the release job being asked to change an immutable release.
- CLAUDE.md names the renamed dry-run artifact.

## After merge

`v3.0.0-alpha2` cannot be repaired (immutable releases cannot be edited), so the next rehearsal is a new tag: dry run with `gh workflow run publish.yml --ref main -f tag=v3.0.0-alpha3`, then push `v3.0.0-alpha3`.


johnsonlee:
`slow-query-shapes` failed on 6e0dc07 on one row that this PR cannot have moved: the diff touches only `.github/workflows/publish.yml`, so the base and candidate JMH jars are built from identical Kotlin sources (both builds in the artifact succeed, `errors: []`, 24 rows).

| Row | Initial | Confirmation | Gate |
|---|---:|---:|:---:|
| WARM/valueHit | 25.7% | 15.2% | FAIL (threshold 15%) |

WARM/valueHit is a ~65 ms query, and the same run shows the runner's variance elsewhere in the table (WARM/wrappedCallerMiss went 75.1% initial, then −27.8% on confirmation). The identical gate passed on #129 an hour earlier with the same benchmark code.

No fix to port: there is no code change behind the number. This session's token gets 403 on re-running jobs, so one re-run of `slow-query-shapes` is needed from the owner; a second failure on identical sources would point at the runner, not this PR.

---

.github/workflows/publish.yml
```diff
@@ -246,17 +246,34 @@ jobs:
...
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    - name: Publish the GitHub Release
+      if: env.DRY_RUN != 'true'
+      env:
+        GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      run: |
+        EXPECTED=$(ls release-assets | wc -l)
+        UPLOADED=$(gh release view "$RELEASE_TAG" --json assets --jq '.assets | length')
+        test "$UPLOADED" -eq "$EXPECTED"
+        gh release edit "$RELEASE_TAG" --draft=false
```
johnsonlee: [P1] Do not make the release immutable before the workflow has a retry-safe boundary. After this command succeeds, the same release job still renders and syntax-checks the formula and may clone/commit/push the Homebrew tap; Docker also waits for the whole job. If any of those steps fails—or GitHub publishes but this client times out—the job is red after the tag and assets are immutable. Rerunning starts at action-gh-release, whose default overwrite path finds the published release and tries to delete/re-upload same-name assets, which immutable releases reject, so the retry can never reach the failed step and Docker remains skipped. Move all preparation before finalization, split post-publication consumers into independently retryable jobs, and add an already-published path that verifies the exact remote asset names/digests and skips upload/finalize. Merely moving this line to the end does not cover an ambiguous publish response.

johnsonlee (commented):
The draft → upload → publish sequence matches GitHub's immutable-release guidance, but the workflow is not retry-safe after publication. The recovery blocker is inline below.

johnsonlee:
[P1] The documented pre-merge/after-merge rehearsal is still not executable. The workflow-dispatch dry-run validates Maven files under `~/.m2/repository/io/github/johnsonlee/graphite-${MODULE}/${VERSION}` and expects `graphite-${MODULE}-*.{pom,jar}`, but `settings.gradle.kts` assigns the published projects the prefix-free artifactIds `core`, `cypher`, `sootup`, and `webgraph`. The successful alpha2 Maven job does not exercise this check because tag-push runs skip the dry-run branch. Therefore the proposed `workflow_dispatch ... alpha3` will fail before it can be used as rehearsal evidence. Fix the lookup to `${MODULE}/${VERSION}/${MODULE}-...` (the existing root cause is also recorded at https://github.com/johnsonlee/graphite/pull/124#discussion_r4002090137), then run the dry-run on this exact head. Note that dry-run still skips the newly changed draft/publish steps, so it proves artifact preparation but not the immutable transition.

.github/workflows/publish.yml
johnsonlee: Done in the second commit, along the three lines you named:

- **Preparation before finalization.** The release job now builds the jar, collects the binaries, verifies the checksums, renders and syntax-checks the formula and uploads all of it as the `release-<tag>` artifact before the one step that changes remote state. That step is the last one in the job.
- **Post-publication consumers are their own jobs.** The tap push moved to a `tap` job (needs `release`, stable tags only, takes the formula from the artifact); `docker` already took its inputs from the published release. Both can be re-run alone once the release exists.
- **Already-published path, no trust in the publish response.** `softprops/action-gh-release` is gone; `.github/scripts/publish-release.sh` finds or creates the draft, makes the draft's assets equal to `release-assets/` by name and sha256 (API `digest`, or download-and-hash when absent), verifies, then publishes with a PATCH whose result is discarded: the state is read back and decides. On a re-run that finds the release published it verifies names and digests against the local assets and uploads nothing (`already-published`), or fails saying the tag cannot be repaired if they differ or the prerelease flag disagrees.

The script has a `--self-test` against a stub `gh` that runs on every publish run (dry runs included): fresh tag, re-run after publication with no writes, unrepairable mismatch, draft left by a failed upload with a stale asset to remove, and a publish request that reports failure after GitHub applied it.

---

johnsonlee:
Fixed in ce9d2b2. The dry-run check now looks at `io/johnsonlee/graphite/${MODULE}/${RELEASE_VERSION}/${MODULE}-${RELEASE_VERSION}.*`, which is what `:core:publishToMavenLocal` produces here and what Maven Central holds for alpha1/alpha2 (`io/johnsonlee/graphite/{core,cypher,sootup,webgraph}/`). It also asserts the POM's `groupId`/`artifactId` and the sources and javadoc jars, so a change in the publication would fail the rehearsal rather than pass by accident.

I cannot start the rehearsal myself: this session's token gets `403` on `workflow_dispatch`. On this head it is

```
gh workflow run publish.yml --ref claude/graphite-explorer-rust-vl2pm8 -f tag=v3.0.0-alpha3
```

Agreed on its scope: it proves artifact preparation (Maven local, four binaries, jar, formula rendered and installed, both image architectures) and the publish script's self-test, not the draft→publish transition, which the dry run skips by design. The transition is exercised for real by the `v3.0.0-alpha3` tag; a re-run of the `release` job after that tag is the check that the already-published path holds.

---

johnsonlee (commented):
Post-merge exact-head review found one deterministic blocker in the new self-test; the first real alpha3 release run has now failed on it before publication.

.github/scripts/publish-release.sh
```diff
@@ -0,0 +1,273 @@
...
+  check_published
+  echo "published"
+}
+
+# --- self-test against a stub gh ------------------------------------------------------
+# The stub keeps releases and assets in a directory and answers only the calls above.
+
+self_test() {
+  local root; root=$(mktemp -d)
+  trap 'rm -rf -- "${root}"' EXIT
```
johnsonlee, line 171: [P1] This EXIT trap dereferences a local after self_test has returned. With global set -u, root is then out of scope, so the script prints that all five scenarios passed but exits 1 from the trap (root: unbound variable). I reproduced the exact script under Ubuntu 24.04 with exit 1, and the real v3.0.0-alpha3 release job 103899295103 failed the same way; the subsequent publish step never ran and tap/docker were skipped. Perform cleanup while root is still in scope, use a RETURN cleanup, or capture a safely quoted concrete path in the trap, and assert the whole self-test process exits 0.

.github/scripts/publish-release.sh
```diff
@@ -0,0 +1,273 @@
...
+# Sets RELEASE_ID / RELEASE_DRAFT / RELEASE_PRERELEASE from the one release that owns the
+# tag, or leaves RELEASE_ID empty. A published release wins over a stale draft; two drafts
+# for one tag is a state this script never creates, so it stops rather than guess.
+read_release() {
+  RELEASE_ID= RELEASE_DRAFT= RELEASE_PRERELEASE=
+  local published=() drafts=() line
+  while read -r line; do
+    [[ -n "${line}" ]] || continue
+    if [[ "${line}" == *" false "* ]]; then published+=("${line}"); else drafts+=("${line}"); fi
+  done < <(matching_releases)
```
johnsonlee, line 43: [P1] Process-substitution failure is not the while loop's exit status: if matching_releases fails because gh api or jq fails, this loop still returns success with both arrays empty. read_release then reports no release and main attempts create_draft, so a transient state-read failure can create a second draft (or at least turn a safely retryable published/draft state into a misleading create failure). This is directly reproducible under set -euo pipefail with a while/read loop fed by process substitution running false; the function exits 0. Capture and validate the complete matching_releases output first, then parse it, and add a self-test where the release-list API fails.

johnsonlee (commented):
A second recovery-path issue remains in the merged script: state-read failures are currently converted into an empty release list.

.github/scripts/publish-release.sh
```diff
@@ -0,0 +1,273 @@
...
+    return
+  fi
+  if [[ -z "${RELEASE_ID}" ]]; then
+    create_draft
+    read_release
+    [[ -n "${RELEASE_ID}" ]] || die "draft release ${RELEASE_TAG} was not created"
+  else
+    log "draft release ${RELEASE_TAG} exists; reconciling its assets"
+  fi
+  reconcile_draft
```
johnsonlee, line 159: [P1] Serialize publishers for the same RELEASE_TAG before reconciling this draft. publish.yml has no per-tag concurrency group, and repository_dispatch can start the same tag more than once. Two runs may rebuild different bytes, both read the same draft, then interleave clobber/verify/publish; one can make the release immutable after the other has changed an asset, leaving a mixed set that post-publish verification can only report as irreparable. Add workflow concurrency keyed by repository and release tag, with cancel-in-progress false, and cover the single-writer invariant.

.github/scripts/publish-release.sh
```diff
@@ -0,0 +1,273 @@
...
+    log "the publish request did not return success; reading the release state back"
+  fi
+  read_release
+  [[ -n "${RELEASE_ID}" && "${RELEASE_DRAFT}" == false ]] || die "release ${RELEASE_TAG} is still a draft"
+}
+
+check_published() {
+  [[ "${RELEASE_PRERELEASE}" == "${PRERELEASE}" ]] \
+    || die "release ${RELEASE_TAG} is published with prerelease=${RELEASE_PRERELEASE}, tag parses as prerelease=${PRERELEASE}"
+  verify_assets || die "release ${RELEASE_TAG} is published and immutable with different assets; it cannot be repaired, release a new tag"
```
johnsonlee, line 137: [P1] This exact-digest retry contract assumes the assets rebuilt by a later release-job attempt are byte-identical, but the workflow does not establish that. Linux packages use ordinary tar -czf, which embeds current mtimes, and Gradle 8.13 shadowJar has no preserveFileTimestamps=false/reproducibleFileOrder=true configuration here. Even at the same commit, rerunning the job can therefore produce different tar.gz or graphite.jar bytes and this path will reject the already-correct immutable release as unrepairable. Either reuse the original immutable run artifacts for recovery or make every archive reproducible and add a two-independent-build digest check.

johnsonlee (commented):
Two additional retry/integrity gaps remain in the merged implementation after the immediate self-test failure is fixed.

PR #132, opened 2026-09-14
Plan a string scan past a conjunct the planner cannot push

johnsonlee:
## Problem

This shape timed out at the 60 s server maximum against a production corpus:

```cypher
MATCH (n)
WHERE n.graphId IS NOT NULL
  AND (coalesce(toString(n.value), "") CONTAINS "<term>"
    OR coalesce(toString(n.name), "") CONTAINS "<term>"
    OR coalesce(toString(n.id), "") CONTAINS "<term>")
RETURN n.graphId, n.type, n.id, n.name, n.value, n.path, n.class_name, n.method_name
LIMIT 80
```

`ScanPlan::build` hands the WHERE clause to `collect_tree`, which required **both** sides of an `AND` to be pushable. `n.graphId IS NOT NULL` has no leaf (`collect_leaf` knows string ops, `=`, `IN`), so the whole tree came back `None`, no plan was built, and the query ran through the row-by-row matcher: every node of every graph decoded and the full expression evaluated on each. The string side alone (`coalesce(toString(n.prop), '') CONTAINS`) was already recognised; the one extra conjunct disabled it.

## Change

`collect_tree` plans a conjunction on whichever side is pushable. The candidates of one side are a superset of `a AND b`, and the WHERE clause is already evaluated on every survivor, so the dropped conjunct is still applied. A plan built this way is marked `relaxed`, and `ScanPlan::verified` then refuses to let any stream's survivors skip that check: the `All`, `Column` and exact CallSite streams normally report their survivors as verified, which would silently drop the conjunct. A disjunction is unchanged: `a OR b` needs both sides' candidates.

The relaxation also covers `n.callee_class CONTAINS 'x' AND n.line > 20`, `... AND NOT n.callee_name = 'y'`, `... AND n.callee_name IS NULL`, and such conjunctions nested under an `OR`.

## Evidence

64-graph corpus (`fixture64`, MAPPED), `/api/cypher`, two runs each; rows compared by SHA-256 over sorted JSON:

| Query | Before | After | After, `GRAPHITE_NO_FASTPATH=1` | Rows |
|---|---:|---:|---:|---|
| shape above, hit term | 29 789 / 18 915 ms | 486 / 15 ms | 18 341 / 17 697 ms | identical (80) |
| shape above, miss term | 28 867 / 26 599 ms | 210 / 17 ms | 25 439 / 24 639 ms | identical (0) |
| shape above, `count(*)` | 27 222 / 25 763 ms | 25 / 25 ms | 26 085 / 25 399 ms | identical (216) |
| same without `graphId IS NOT NULL` | 511 / 18 ms | 15 / 14 ms | 14 623 / 17 049 ms | identical (80) |

Single graphs (the CI core fixture and the acme graph): 12 shapes × 2 graphs, fast path vs `GRAPHITE_NO_FASTPATH=1`, all 24 outputs byte-identical.

## Tests

- **Planner units** (scan.rs): the shape plans with `relaxed` set and the three string leaves; the conjunct on either side and under `OR` still plans; a non-relaxed plan keeps trusting its streams; `IS NOT NULL` alone, two unpushable conjuncts, and an `OR` with an unpushable side build no plan.
- **Runtime against the real fixture** (scan.rs, gated on `GRAPHITE_INDEX_FIXTURE`, run by Rust CI next to the storage crate's fixture test): a cross-graph executor checks that the dropped conjunct decides the rows: `A AND B` + `NOT A AND B` = `B` for an ordering on the node id over the call-site and annotation sweeps; `graphId IS NOT NULL` keeps the plain count and `graphId IS NULL` empties it; `callee_name IS NULL` empties a call-site search whose stream would otherwise be taken as verified.
- **Parity** (`backend/bench/parity.py`): the shape and its relatives, in pairs that differ only by the dropped conjunct, run against the Kotlin server on both routes in CI.
- Cypher crate: 73 tests; workspace clippy `-D warnings` and fmt clean.

## Coverage

- JVM modules are untouched, so their coverage stays at the reported ≥ 98%.
- Rust (`cargo llvm-cov -p graphite-cypher`, with the fixture): 19 of the 20 instrumented production lines this PR adds are executed. The one that is not is the third of three identical `self.verified(...)` pass-throughs, inside the annotation sweep of the CallSite-only path; reaching it needs a graph whose string dictionary holds a CallSite property name as an annotation attribute key, which neither fixture does. For context, the crate's unit-test line coverage is 48.2% with the fixture (35.0% without); most of the engine is exercised by the differential harnesses in CI rather than by unit tests.


johnsonlee:
`slow-query-shapes` failed on 0ce3945 on one row of the Kotlin JMH gate, which this PR cannot have moved: the diff touches `backend/cypher` (Rust), `backend/bench/parity.py` and `rust.yml`; no JVM source changes, so the base (5313f7b) and candidate JMH jars are built from identical Kotlin (both builds succeed, `errors: []`, 24 rows).

| Row | Base ms | Candidate ms | Initial | Confirmation | Gate |
|---|---:|---:|---:|---:|:---:|
| WARM/wrappedCallerHit | 276.3 | 318.9 | 15.4% | 37.0% | FAIL (threshold 15%) |
| WARM/wrappedCallerMiss | 292.7 | 140.7 | −51.9% | +140.5% | PASS |

The neighbouring row swinging from −52% to +140% between the two runs on the same jars shows the runner's variance on this family (the same `wrappedCaller*` rows have been the noisy ones on #130 and earlier). Every other check on this head is green: Rust (incl. the new fixture-gated test and the Kotlin-vs-Rust differential with the added parity shapes), unit tests with coverage ≥ 98% in every module, and the remaining gate components.

No fix to port: there is no code change behind the number. This session's token gets 403 on re-running jobs, so one re-run of `slow-query-shapes` is needed from the owner.

---

PR #139, opened 2026-09-14
Plan a disjunction inside any(k IN keys(n) ...) and expose a local variable's method to keys(n)

johnsonlee:
## Problem

The search-every-property shape the UI emits with several terms,

```cypher
MATCH (n) WHERE any(k IN keys(n) WHERE toString(n[k]) CONTAINS "xxx" OR toString(n[k]) CONTAINS "yyy" OR toString(n[k]) CONTAINS "xxyy") RETURN n LIMIT 80
```

had no plan: `expand_keys_predicate` accepted exactly one test under the quantifier, so an `OR` inside it decoded every node of every graph. On the 64-graph corpus a miss (or a `count(*)`) took 166–174 s where the single-term form takes 60 ms.

## Change

**Planner (`scan.rs`).** `any(k IN keys(n) WHERE A OR B)` is `any(k IN keys(n) WHERE A) OR any(k IN keys(n) WHERE B)`, so every disjunct's per-key leaves now join one disjunction (`keys_test_leaves`). An unpushable disjunct (regex, a test on `k` itself) still declines the whole quantifier, and a conjunction is never pushed: `any(A AND B)` needs one key satisfying both, which the union of per-key leaves does not express.

| Shape (64 graphs) | Before | After |
|---|---:|---:|
| 3-term OR, hits, `LIMIT 80` | 1,484 / 1,336 ms | 102 / 102 ms |
| 2-term OR, both miss, `LIMIT 80` | 166,222 / 174,012 ms | 5,781 / 2,628 ms |
| 2-term OR, `count(*)` | 163,560 / 161,667 ms | 1,911 / 1,988 ms |
| `graphId = x AND` 2-term OR | 957 / 987 ms | 81 / 73 ms |
| single term (unchanged path) | 167 / 60 ms | 173 / 53 ms |

**Correctness (`props.rs`).** Checking the rows turned up a pre-existing disagreement between the streamed and the decoded path for this shape, single-term included: `any(k IN keys(n) WHERE toString(n[k]) CONTAINS 'getInstance') RETURN count(*)` gave 34,048 planned and 20,607 with `GRAPHITE_NO_FASTPATH=1`, all of the difference in `LocalVariable`. The baseline's `keys()` and `properties()` come from `NodePropertyAccessor.getAllProperties`, which lists a local variable's `id`, `name`, `type` and `method`; only its `RETURN n` map (`nodeToMap`) leaves the method out. The Rust `node_properties` mirrored the `RETURN n` map, so the decoded evaluation of `keys(n)` never saw `method` while the planner (correctly) had a `method` leaf for local variables. `node_properties` now carries `method` and `node_display_properties` drops it for local variables, so `RETURN n` is unchanged. With the fix, the decoded path counts 34,048 / 8,862 / 42,909 for the single-term, single-term and two-term shapes, identical to the streamed path.

## Tests

- `a_disjunction_inside_the_keys_quantifier_plans_as_one_disjunction`: the three-term shape yields exactly three times the single-term leaves, one set per literal, all `Contains` and `from_keys`; mixed operators and operands; the shape under `AND` with a pushable and with an unpushable conjunct; a regex disjunct, a test on `k`, and a conjunction decline.
- `a_local_variable_exposes_its_method_to_the_keys_search` (fixture-gated, runs in CI's `GRAPHITE_INDEX_FIXTURE` step): `keys(n)` of a local variable contains `method`; the streamed plan, the same query relaxed by an unpushable conjunct (decoded rows judged by the WHERE clause) and the spelled-out `n.method OR n.name OR n.type` form count the same rows. Fails without the `props.rs` change.
- `backend/bench/parity.py`: four OR-inside-`any` queries (string, mixed operators, an `AND` inside the quantifier) and three local-variable rows (`keys(n)`, `properties(n)`, the keys search restricted to `LocalVariable`) against the Kotlin engine.
- `cargo test --release --locked --workspace` with and without the fixture, `cargo clippy --all-targets -D warnings`, `cargo fmt --check`: clean.

---


johnsonlee:
Benchmark run [547](https://github.com/johnsonlee/graphite/actions/runs/34855971297) on `6028b59` is red for a reason outside this PR: `graph-routing-pressure-evidence` failed in its first build step, before checkout of the corpus or any benchmark, because the Gradle wrapper could not fetch its distribution:

```
java.io.IOException: Server returned HTTP response code: 504 for URL: https://github.com/gradle/gradle-distributions/releases/download/v8.13.0/gradle-8.13-bin.zip
```

at 14:49:46 UTC. Every other job in the run that builds with Gradle restored the same distribution from the `setup-gradle` cache; the two `*-pressure-evidence` jobs are the only ones without that action, so they download from GitHub Releases each time. GitHub's artifact and release services have been failing intermittently this hour (#141's run 548 hit a 403 finalizing artifact uploads at 14:33). This PR touches `backend/cypher` and the parity harness only.

This session's token cannot re-run failed jobs (403 on `rerun-failed-jobs`), so a re-run of run 547 from the Actions page is needed. Adding `gradle/actions/setup-gradle@v6` (`cache-read-only: true`) to the two evidence jobs would make them immune to this, but that is a workflow change and belongs with #141 rather than this planner PR.

---

PR #141, opened 2026-09-14
Warm the fixture64 cache from main so pull requests start from a hit

johnsonlee:
## Problem

#137's own `prepare-fixture64` still took 20 minutes (run 34847519168): a 1m18s build with the warm Gradle cache, then a 16-minute miss. The miss was structural, not eviction. `benchmark.yml` runs only on `pull_request`, and a cache entry saved by a pull request is visible only to that pull request's branch, so every new branch generates the corpus twice on its first run, and the entry it saves helps nobody else. GitHub makes entries saved from the default branch visible to every pull request; nothing on main saved one.

This commit was pushed to #137 at 13:41 and the PR merged at 13:43 without it, so it comes on its own branch.

## Change

`.github/workflows/fixture64-cache.yml`: on push to `main` (paths limited to the key's inputs: `frontend/jvm/**`, the Gradle files, the three fixture64 scripts, the workflow itself), weekly on Monday 03:00 UTC (inside the seven-day eviction window for unused entries), and on dispatch. It checks out main as `candidate` (the key's `hashFiles` globs are under that path), builds the generator, restores by the same key, and only on a miss generates the corpus, runs the reproducibility test with the repeat copy under `$RUNNER_TEMP`, and saves the corpus plus receipt.

`benchmark-gate.test.mjs` now extracts the `key: >-` block from both workflows and asserts they are identical character for character, so the two cannot drift apart silently; it also checks the warm-up triggers on `main` and saves with `actions/cache/save@v5`. 133 gate tests pass.

**Second commit:** `graph-routing-pressure-evidence` and `global-wide-pressure-evidence` gain `gradle/actions/setup-gradle@v6` (`cache-read-only: true`). They were the only Gradle jobs in the workflow without it, so each downloaded `gradle-8.13-bin.zip` from GitHub Releases on every run; runs 547 (#139) and 548 (this PR) lost both jobs to an HTTP 504 on that download before any benchmark step ran. With the cache the distribution is restored the way every sibling job restores it.

The first run on main after merge misses once (about 18 minutes); every pull request after that restores main's entry and runs the 6-minute hit path from #137.


johnsonlee:
Benchmark run [548](https://github.com/johnsonlee/graphite/actions/runs/34855984857) on `f380592` is red for a reason outside this PR: `build-wrapped-query-jmh-base`, `build-explore-jmh-base` and `build-explore-jmh-candidate` each built their jar (`BUILD SUCCESSFUL`) and then failed in `actions/upload-artifact` with

```
Failed to FinalizeArtifact: Received non-retryable error: Failed request: (403) Forbidden: Error from intermediary with HTTP status code 403 "Forbidden"
```

at 14:33:23–14:33:36 UTC, while `build-wrapped-query-jmh-candidate` and `build-wrapped-query-jmh-reference` uploaded the same way seconds earlier and succeeded. Every later failure (`wrapped-query-latency`, `method-compatibility-gate`, the skipped shards and the gate) is downstream of those missing jars. This PR changes only `.github/workflows/fixture64-cache.yml` and a gate-test assertion, neither of which is on that path, and the gate tests passed in `candidate-gate-tests`.

This session's token cannot re-run failed jobs (403 on `rerun-failed-jobs`), so a re-run of run 548 from the Actions page is needed; a second identical 403 would mean the artifact service problem is persisting rather than transient. Runs 545 (on `64266fe`) was cancelled by the branch update, not failed.

---

PR #146, opened 2026-09-15
Serve every .graphite file under --data by its file name

johnsonlee:
## Why

With a graph as one file (#145), a data directory is a catalog. `graphite serve --data /data/graphs` (and `graphite mcp --data`) now serves every `*.graphite` file directly under it without `--graph id:path` for each.

## Rules

- Every `*.graphite` file directly under `--data` is loaded, sorted by name, under its file name without the extension: `orders.graphite` is `orders`. Startup logs `Discovered N graph(s) in <dir>`.
- A file name that is not a valid graph id fails startup naming the file; an id both discovered and given with `--graph` is an error. The directory's contents are the catalog, with no ambiguity.
- Directories, other files (`.sha256` sidecars included) and nested files are not graphs here. A directory graph is still named with `--graph`; the Kotlin server's `--data` keeps its meaning, and the differential (explicit `--graph`, no `--data`) is unaffected.
- Discovery happens at startup only; a watch mode is the next step on top of this.

## Verification

- Unit tests in `backend/explore/src/serve.rs`: discovery lists only `.graphite` files at the top level, in name order, ignores sidecars, directories and nested files, returns empty for an absent directory, and refuses a file name that is not a graph id.
- Locally: a directory with `acme-shop.graphite` and `android00.graphite` serves both under those ids with `--data` alone; adding `--graph acme-shop:<dir>` fails with `Graph id 'acme-shop' is both given with --graph and discovered as .../acme-shop.graphite under --data`.
- The Rust CI job now serves a directory holding one `.graphite` file and asserts `/api/graphs` lists exactly `acme-shop` and the log reports one discovered graph.
- Workspace clippy `-D warnings`, rustfmt and the explore crate's 84 tests pass.

---


johnsonlee:
**`wrapped-query-latency-shard-real-a` failed on `decd17b9`, and it is not this PR's.**

The row is the JVM `webgraph.AllFixtureWrappedDiscoveryLatencyBenchmark.zeroHitBroadContainsCaseInsensitiveDiscovery`: base 0.227 s/op, PR 0.276 s/op (+21.5%), confirmation 0.222 vs 0.288 s/op (+30.2%), with `confidenceSeparated: false` both times (the confirmation's candidate mean is 0.508 ± 1.856 s/op at 99.9%).

This PR changes `backend/explore/src/serve.rs`, the README and `rust.yml` only; nothing under `frontend/jvm/`. I downloaded the run's `jmh-wrapped-query-base-146` and `jmh-wrapped-query-candidate-146` artifacts: both jars are 60,328,199 bytes with 29,806 entries, and every entry's CRC-32 and size are identical. The shard compared identical bytecode, so the +30% is hosted-runner noise on a 0.25 s single-shot benchmark (the same row read −22.9% on #145 an hour earlier). The point-estimate policy of this gate blocks regardless of interval overlap, so the shard fails by design.

No code fix applies. Every other job on this head is green (`rust-build-test` with the new `--data` discovery smoke, unit tests, all other benchmark shards and gates). I cannot re-run jobs from this session; a re-run of the failed job should clear it.

---

PR #151, opened 2026-09-17
Time MCP tool calls by tool and outcome under --metrics

johnsonlee:
## Why

`--metrics` saw `POST /mcp` as one line in `http_server_requests_seconds`: every tool and every JSON-RPC method (`initialize`, `tools/list`, `tools/call`) landed in the same series. The API hop a tool makes inside the process goes through the router captured *before* `instrumented()` wraps it, so it was not timed under its route template either. A slow tool could not be told from a fast one.

## What

Under `--metrics` the MCP server records two families, exported after the Cypher families:

- `graphite_mcp_requests_total{method}`: JSON-RPC requests by method. The label set is closed: `initialize`, `ping`, `tools/list`, `tools/call`, and everything else as `other`. Notifications are not requests and are not counted.
- `graphite_mcp_tool_duration_seconds{tool,outcome}`: a histogram over the Cypher SLO buckets, plus a `_max` gauge. `tool` is the server's own `&'static str` tool name (what `tools/list` returns), never the client's string; an unknown tool is refused before it is timed. `outcome` is `ok`, or `error` when the tool answered `isError`.

Recording is off without `--metrics` (`McpServer::new`) and over stdio (`graphite mcp`), so the default request path stays uninstrumented. A tool call remains one `/mcp` request in `http_server_requests_seconds`; the inner hop is not counted a second time.

## Verification

- Unit tests: closed method set and `other` mapping, bucket placement, tool/outcome series, refused calls not timed, a server without metrics records nothing, and a `/metrics` scrape through the instrumented router before and after a `tools/call`.
- `backend/bench/parity.py` lists the three families as Rust-only so the Kotlin comparison keeps passing.
- The Rust CI smoke scrapes `/metrics` after `mcp-smoke.py` and asserts the `tools/call` count and the `graphs` tool series. Run locally against a fixture graph: 33 smoke checks pass and the scrape shows the expected lines.
- `cargo fmt`, `cargo clippy --workspace --all-targets -D warnings`, `cargo test -p graphite-explore` (88 passed).

stdio metrics are out of scope by request.

---


johnsonlee:
`validate-cpu-accounting` failed on a6ed5453 before its first step ran: the runner could not download `gradle/actions@v6` (`codeload.github.com` answered 429 three times, see the [job log](https://github.com/johnsonlee/graphite/actions/runs/35242169408/job/105274035502)). Nothing in this PR touches that job or Gradle; the report job depends on it, so the benchmark gate will show red for this run as well.

I do not have permission to re-run the job from this session. A re-run of the failed jobs on run 35242169408 should clear it; I will keep watching the PR and act on the rest of CI as it lands.

---

PR #153, opened 2026-09-21
Stream MATCH [WHERE] WITH through the scan pushdown and group-by

johnsonlee:
## Problem

A grouped count over the fleet timed out at 60 s in cross-graph mode (`all_graphs: true`, `mode: cross-graph`):

```cypher
MATCH (n {type: "CallSiteNode"})
WHERE n.callee_class STARTS WITH "com.example.apigateway.services."
  AND NOT n.caller_class STARTS WITH "com.example.apigateway.services."
WITH n.graphId AS graphId, split(replace(n.callee_class, "com.example.apigateway.services.", ""), ".")[0] AS provider, count(*) AS calls
RETURN graphId, provider, calls
ORDER BY graphId ASC, calls DESC
LIMIT 160
```

The fused shape covered `MATCH [WHERE] RETURN …` only. With a `WITH` in its place the match ran through the generic matcher: every call site of every graph decoded, the `WHERE` interpreted per node, every survivor kept as a row until the `WITH` grouped them. The cost was the corpus size whatever the predicate matched. The same query written with `RETURN` answered from the string index in a fraction of the time.

## Change

- `FusedShape::detect_with` recognises the `MATCH [WHERE] WITH <items> [WHERE]` prefix.
- `run_fused` is split so the `RETURN` shape and the new `WITH` prefix share `fused_rows`: the same streaming scan pushdown, projection and group-by.
- The `WITH`'s own `WHERE` and everything after it stay with the clause loop, so the rows it sees are the ones `project()` would have built (order stashes, provenance and all). No clause semantics change.
- A row-preserving `WITH` under a literal `LIMIT` keeps the early bound `compute_early_limit` already gave the generic match.

## Measurements

Sixteen copies of a 63K-node graph, cross-graph, release build:

| shape | before | after |
|---|---:|---:|
| reported query, dense prefix | 620–746 ms | 100 ms (`RETURN` twin: 98 ms) |
| reported query, sparse prefix | 580 ms | 7.5 ms |

## Tests

- Two parse-level tests for `detect_with` (what is and is not the prefix shape, which `ORDER BY` belongs to the `WITH`).
- A fixture-gated test (`GRAPHITE_INDEX_FIXTURE`, as CI sets it) runs ten `WITH` shapes as two cross-graph sources and compares columns, rows and provenance against the generic pipeline (the same query with an `UNWIND [1] AS one` forcing it): grouped counts, a `LIMIT`-bounded row-preserving `WITH`, the `WITH`'s own `WHERE`, `WITH DISTINCT`, `WITH … ORDER BY … LIMIT` on pre-projection expressions, `WITH *`, group-less aggregates over `graphId`, and an unknown label.
- `GRAPHITE_NO_FASTPATH=1` and the new path produced byte-identical output on eight `WITH` shapes over the sixteen-copy corpus.
- `cargo test --workspace` with the fixture and `cargo fmt --check` pass.

https://claude.ai/code/session_018wuL99M8WLQ8ZXJRCm7C7U

---
_Generated by [Claude Code](https://claude.ai/code/session_018wuL99M8WLQ8ZXJRCm7C7U)_


johnsonlee:
CI note: the advisory JVM engine gates flap on this PR while every blocking gate passes.

- On `86c51c5`: `cypher-capacity-gate` (jq envelope over the Kotlin engine's `CypherCapacityBenchmark` single-shot run) and `method-compatibility-4-position` (`middle` scenario 443.9 ms candidate vs 358.9 ms base, past 15%; `zero` and `early` within noise) failed; `benchmark-regression-gate` reported **PASS, 7/7 blocking**.
- On `4bffbc1` (docs-only push): both of those passed. Instead `method-compatibility-4-string` failed, and `wrapped-query-latency-shard-real-a` (so `wrapped-query-latency`) failed on `AllFixtureWrappedDiscoveryLatencyBenchmark`, whose three single shots came back as `0.297 ± 0.285 s/op` for the zero-hit query: the error bar is the measurement.

All of these measure the Kotlin engine's JMH jars (`explore`, `webgraph`) with one fork, no warm-up and one or three single shots per scenario against a 15% threshold. This PR changes one Rust file (`backend/cypher/src/engine/pipeline.rs`) plus a docs entry, so base and candidate run byte-identical Kotlin, and the shard that trips is a different one each run: hosted-runner variance, not a regression. Per `benchmark-regression-gate` these jobs are advisory (a warning, never the verdict); `rust-latency` is within noise on every row and the Rust and unit-test workflows are green on both heads.

I cannot re-run jobs from this session (Actions API returns 403). If the red advisory checks should not show on PRs at all, the small workflow change is `continue-on-error: true` on `method-compatibility-shard`, `wrapped-query-latency-shard` and `cypher-capacity-gate` (their results already reach the verdict as advisory); that is a CI design choice for a separate PR, not something I will fold into this one.

---

